Issue #4, opened 2021-12-25
The game becomes unresponsive when tabbing out


platnummunkey:
The game becomes unresponsive when tabbing out on fullscreen.

MiloAkerman:
Unresponsive how?

Haplo164:
I tabbed out during a map change and got an infinite loading screen.

ASpoonPlaysGames:
Closing as stale, there are quite a few reasons why this could happen (things like the server dying)

PR #5, opened 2021-12-25
Ping code - For display in server browser

barnabwhy:
- Added a system for looking up and displaying ping, although slightly intrusive
- Can be made less taxing on game servers if PR [#4](https://github.com/R2Northstar/NorthstarMasterServer/pull/4) on NorthstarMasterServer is added
- ~~Requires PR [#9](https://github.com/R2Northstar/NorthstarMods/pull/9) on NorthstarMods~~ May not be necessary if [F1F7Y/Better.Serverbrowser](https://github.com/F1F7Y/Better.Serverbrowser) is going to be included by default


barnabwhy:
Ping display:
image

When ping cannot be fetched:
image

NorthstarDedicatedTest/masterserver.cpp
```diff
@@ -545,6 +567,134 @@ void MasterServerManager::AuthenticateWithServer(char* uid, char* playerToken, c
...
+		printf("\tSent icmp message to %s\n", ip);
+		if (dwRetVal > 1) {
+			printf("\tReceived %ld icmp message responses\n", dwRetVal);
+			printf("\tInformation from the first response:\n");
+		}
+		else {
+			printf("\tReceived %ld icmp message response\n", dwRetVal);
+			printf("\tInformation from this response:\n");
+		}
+		printf("\t  Received from %s\n", inet_ntoa(ReplyAddr));
```
PixieCore, line 680: inet_ntoa is deprecated maybe use inet_ntop() instead?

NorthstarDedicatedTest/masterserver.cpp
```diff
@@ -545,6 +567,134 @@ void MasterServerManager::AuthenticateWithServer(char* uid, char* playerToken, c
...
+	// Declare and initialize variables
+
+	HANDLE hIcmpFile;
+	unsigned long ipaddr = INADDR_NONE;
+	DWORD dwRetVal = 0;
+	char SendData[32] = "Data Buffer";
+	LPVOID ReplyBuffer = NULL;
+	DWORD ReplySize = 0;
+
+	ipaddr = inet_addr(ip);
```
PixieCore, line 648: Maybe use InetPton instead here also.

NorthstarDedicatedTest/masterserver.cpp
```diff
@@ -54,6 +61,13 @@ RemoteServerInfo::RemoteServerInfo(const char* newId, const char* newName, const
 
 	playerCount = newPlayerCount;
 	maxPlayers = newMaxPlayers;
+
+	ping = newPing;
+}
+
+void RemoteServerInfo::SetPing(std::string newPing)
```
PixieCore, line 68: Can be const std::string&

NorthstarDedicatedTest/masterserver.cpp
```diff
@@ -35,7 +42,7 @@ CHostState__State_ChangeLevelSPType CHostState__State_ChangeLevelSP;
 typedef void(*CHostState__State_GameShutdownType)(CHostState* hostState);
 CHostState__State_GameShutdownType CHostState__State_GameShutdown;
 
-RemoteServerInfo::RemoteServerInfo(const char* newId, const char* newName, const char* newDescription, const char* newMap, const char* newPlaylist, int newPlayerCount, int newMaxPlayers, bool newRequiresPassword)
+RemoteServerInfo::RemoteServerInfo(const char* newId, const char* newName, const char* newDescription, const char* newMap, const char* newPlaylist, int newPlayerCount, int newMaxPlayers, bool newRequiresPassword, std::string newPing)
```
PixieCore, line 45: Can be const std::string&

NorthstarDedicatedTest/masterserver.cpp
```diff
@@ -545,6 +567,134 @@ void MasterServerManager::AuthenticateWithServer(char* uid, char* playerToken, c
...
+
+			REQUEST_END_CLEANUP:
+			spdlog::error("Finished authenticating with server");
+			return (std::string)("NaN");
+
+		});
+
+	requestThread.detach();
+
+	return (std::string)("NaN");
```
PixieCore, line 634: std::string(); here.

PixieCore (commented):
General code changes.

NorthstarDedicatedTest/masterserver.cpp
barnabwhy: Any idea how I would go about changing it to use InetPton? This is basically my first time using C++ so I'm clueless

barnabwhy:
It now crashes and idk why

PixieCore:
> It now crashes and idk why

Test review changes 1by1, I might also be wrong, its just general advice. If something does not work the way I thought it would then it be like that.

barnabwhy:
Currently requires a refresh of the browser to show the ping of servers, unsure why but I would greatly appreciate some help solving this issue.
Apart from this, everything works great!

barnabwhy:
It sometimes works without a refresh and sometimes doesn't. I really don't know why

Alystrasz:
Could you use a translation key to display "Unknown" when ping cannot be fetched?
I'm not sure most players understand the `NaN` value ;)

barnabwhy:
> Could you use a translation key to display "Unknown" when ping cannot be fetched? I'm not sure most players understand the `NaN` value ;)

I had changed it to say "-1" however it does make more sense to say unknown, should be an easy change

nanovad:
> Could you use a translation key to display "Unknown" when ping cannot be fetched? I'm not sure most players understand the `NaN` value ;)

You could avoid having to localize it by having the UI formatting display a `?` question mark when the ping result is -1.

Alystrasz:
> > Could you use a translation key to display "Unknown" when ping cannot be fetched? I'm not sure most players understand the `NaN` value ;)
> 
> You could avoid having to localize it by having the UI formatting display a `?` question mark when the ping result is -1.

That's actually a good idea.
@barnabwhy it's up to you!

barnabwhy:
This works as intended for the most part. Might be useful to have it checked over once or twice.

barnabwhy:
Also I'll still keep this as a draft until I have confirmation on the proposed change to the master server.

barnabwhy:
Currently resolving conflicts with newer code. Hopefully the last time I have to do this cause it sometimes means I need to reorganise stuff.

barnabwhy:
Should be good for release once `/client/servers_with_addresses` is merged, honestly

barnabwhy:
Works a charm in tests on my local machine

MiloAkerman:
i'm unsure what the current implementation is but to avoid exposing server IPs to user couldn't you tell the masterserver you want the ping for a server, after which the masterserver would relay the info to the server, who would ping you, and then send the ping num back through the masterserver to the client?

NorthstarDedicatedTest/masterserver.cpp
```diff
@@ -927,6 +954,99 @@ void MasterServerManager::RemoveSelfFromServerList()
...
+					return;
+				}
+				ReplySize = sizeof(ICMP_ECHO_REPLY) + sizeof(SendData);
+				ReplyBuffer = (VOID*)malloc(ReplySize);
+				if (ReplyBuffer == NULL) {
+					spdlog::error(fmt::format("Encountered an error pinging server \"{}\". Error: Unable to allocate memory", server->name));
+					server->pingPending = false;
+					return;
+				}
+				dwRetVal = IcmpSendEcho2(hIcmpFile, NULL, NULL, NULL,
```
RaidMax, line 998: Many machines will have ICMP packets blocked via firewall eg all windows machines by default which will make this ineffective. Additionally ICMP is low priority and will not be an accurate indicator of latency. This should be done by sending a query packet to the UDP game port(s) and measuring the time until a response is received, and ideally average a few.

barnabwhy: The problem with pinging the UDP port is that it exposes which port the server is using and negates any reason for people to set up servers on non-default ports as it loses all security value.
From my testing about 40-50% of servers had ICMP allowed. Server owners could also just be asked to allow ICMP if they want their server’s ping showing in the server list.
It also actually seems like it’s a relatively accurate indicator by chance, deviating ~10ms due to the time lost by ICMP’s low priority approximately lining up with the game server’s internal latency.

barnabwhy:
> i'm unsure what the current implementation is but to avoid exposing server IPs to user couldn't you tell the masterserver you want the ping for a server, after which the masterserver would relay the info to the server, who would ping you, and then send the ping num back through the masterserver to the client?

The problem is that almost no clients will have ports open or ICMP allowed through their firewall. As a result, the game server can’t ping clients. If it worked, however, it would also require an update for servers as well meaning the rollout time would be slow.

NorthstarDedicatedTest/masterserver.cpp
RaidMax: I’m not sure I understand your first point. Unless the game traffic is proxied through the master which is highly inefficient and a bad idea, the client is still going to be transmitting and receiving packets directly to/from the game ports which is extremely trivial to figure out with any packet capturing software. To my knowledge in context where latency is extremely important there is no way to truly obfuscate remote endpoint connections so any attempt is going to be trivial to bypass.

barnabwhy: It exposes the port to people not trying to join the server, making it easy for them to be attacked

RaidMax: Once again a malicious actor can very easily determine the game port. Using ICMP is not going to do anything other slow them down for maybe 5 seconds.

barnabwhy: Yeah but I also wrote the code to do ICMP and i had looked into making it do a TCP/UDP ping which is way harder. But god was it a pain in the arse to program it even for ICMP, I'd rather just stick with what I've got.

Rackover:
I think exposing game servers IP is perfectly fine, they broadcast their IP to the master server anyway and when you connect to a game you're getting the IP of whoever you're connecting too, so it's not like this is a security issue

GeckoEidechse:
Given that we kinda decided to not do ping code due to exposing many IPs all at once, I'm closing this PR.

PR #10, opened 2021-12-26
Add version resource

### LINKED ISSUE #9: Add version info to NorthstarLauncher.exe
Adding version info via a [version resource](https://docs.microsoft.com/en-us/windows/win32/menurc/versioninfo-resource) would allow third party update checkers to easily check which version is installed

MindSwipe:
This fixes #9


abarichello:
Thanks!

PR #16, opened 2021-12-27
Add server browser filter

KanjiuAkuma:
Adds a simple search box to the server browser to search for servers.
Input is treated as regex because I'm too lazy to implement anything myself, could change that if needed.

Mods PR: https://github.com/R2Northstar/NorthstarMods/pull/16

Example:
image


KanjiuAkuma:
see https://github.com/R2Northstar/NorthstarMods/pull/16#issuecomment-1001786649

PR #18, opened 2021-12-28
Make timestamp year-month-day

GeckoEidechse:
Make timestamp `year-month-day` instead of `day-month-year`. 

The old format is bothering me more than it should :P

### DIFF
--- a/NorthstarDedicatedTest/logging.cpp
+++ b/NorthstarDedicatedTest/logging.cpp
@@ -176,12 +176,12 @@ void InitialiseLogging()
 	spdlog::flush_on(spdlog::level::info);
 
 	// log file stuff
-	// generate log file, format should be nslog%d-%m-%Y %H-%M-%S.txt in gamedir/R2Northstar/logs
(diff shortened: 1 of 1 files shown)


MiloAkerman:
`day-month-year` makes sense though?

GeckoEidechse:
The advantage of `year-month-day` is that when the sort order is _alphabetical_ (which is the default in Windows) logs are also sorted _chronologically_.
Making finding the newest log as simple as just scrolling to the bottom.

The issue isn't apparent yet due to the short time span since public release but will be more obvious once the next month rolls in where the newest logs will be squished between those at the beginning of January and the old ones from December. This problem will only exaggerate as time increases.

Further there's no confusion where a US user would interpret `nslog01-02-2022 13-18-17.txt` as 2nd of January 2022 while the (sensible) rest of the world would read it as 1st of February 2022.

As no one uses `year-day-month`, I'd argue it's harder to misinterpret `nslog2022-02-01 13-18-17.txt` as 2nd of January 2022.

PR #19, opened 2021-12-29
Big refactor of launching process

p0358:
The new launcher exe will start the game directly in its own address space. No illegal calls will be made in DllMain of Northstar.dll. Fixes Windows 7 access violation issues.

---

Includes optional launcher.dll proxy (need to rename old launcher.dll to launcher.org.dll). This allows the game to be launched through Origin's DRM (original Titanfall2.exe) and completely avoid the "file corruption" error. Good backup solution if someone keeps getting that error and cannot get rid of it.

The above is not yet recommended as a default installation option, probably, because of no super-straightforward way of toggling between vanilla game and Northstar, and more complicated installation. A file like `run_northstar.txt` with content of either `0` or `1` is one idea on how to handle it. Currently both launch methods will launch the original game if passed the `-vanilla` parameter.

---

Both loading method can be used at once and they will not interfere with each other negatively.

TODO: Currently support for `ns_startup_args.txt` and `ns_startup_args_dedi.txt` was removed.
My easiest recommendation is to restore its support in Northstar.dll itself, hook CommandLineA() inside of InitialiseNorthstar() (before official LauncherMain is called), since LauncherMain does use this function to initialize the CommandLine() wrapper (it ignores the LauncherMain function parameters completely).

---

Also included a little refactor of masterserver communication by setting client options in one place and allowing to disable HTTPS verification with `-msinsecure` option for local debugging of the communication (@BobTheBob9 fix that 500 backend error I'm getting smh my head)


pg9182:
Is there a way you can do this without replacing DLLs? Maybe by loading both from a separate process? I'm hoping to implement an option to load tf2 from a separate dir so I can have it mounted read-only in the container so it can run without side-effects or race conditions.

p0358:
> Is there a way you can do this without replacing DLLs? Maybe by loading both from a separate process?

Sadly as far as I thought, not really. Though today after sleep I think that maybe a proxy of some launcher.dll's or Titanfall2.exe's dependency could work, maybe... In theory this hook dll doesn't do anything with launcher.dll itself, so it could work.

> I'm hoping to implement an option to load tf2 from a separate dir so I can have it mounted read-only in the container so it can run without side-effects or race conditions.

Maybe CoW or symlinks could help, I'm just wondering how much such feature could screw things up without that xd

But, launcher.dll proxy is only needed for client where Origin throws up, I assume container is for dedi, so it's not needed there, you can just use the new launcher exe without replacing that DLL

p0358:
FYI I fucked something up in the last commit and it doesn't copy any mod files into runtime/compiled at all besides scripts.rson. Otherwise previously it worked pretty nicely

p0358:
But yeah, after last commit it's borked and I need to take a break now, so pushed it out, idk how to debug the no copying of files

pg9182:
Two ideas:

- I could just do the opposite method and symlink files from the volume into the container (this is probably what I'll do, since I'm making a custom entrypoint anyways).
- It may be possible to patch wine to handle the redirections... I'll be looking into this more later.

p0358:
I call ready to merge

PR #21, opened 2021-12-29
Allow auth server bind address to be changed

barnabwhy:
adds convar `ns_server_bind_address` (which defaults to 0.0.0.0), allowing the server host to choose which local ip the server binds itself to.
Likely allows fixes for some issues with setups using multiple network adapters.


p0358:
Shouldn't this ideally synchronize to the bind address of the game's sockets? (I think it's controllable with `ip` cvar or `-ip` command line option, would need to look into whether the cvar is synchronized from the option and at which stage)

L1ghtman2k:
https://developer.valvesoftware.com/wiki/Command_Line_Options, `-ip` is the option

barnabwhy:
On the valve site it says
<html><body>


-ip | Does the same as the ipname Cvar
-- | --



</body>
</html>

So would it make sense to use the ipname Cvar to change the auth server bind address?

PR #31, opened 2022-01-02
NOP out call to VGUI_Shutdown during server DLL shutdown

geniiii:
Avoids a crash that occurs when quitting the game from the console.
Kind of a hacky workaround, but it seems to work fine.


abarichello:
Thanks!

PR #32, opened 2022-01-03
Add VS Build Tools instructions

JJRcop:
Maybe this is unnecessary, but adds some instructions for using the Visual Studio Build Tools to compile Northstar.

This allows people to compile the project who don't want to install the full IDE.


abarichello:
Thanks!

PR #34, opened 2022-01-04
Audio Override

HappyDOGE:
This commit adds audio override functionality.

Audio overrides should be located in the "audio" folder, which should be located inside the mod folder.
Audio override consists of a definition file (JSON) and a folder with override samples.
Override samples must be Stereo 16-bit WAV files and must be at least 1 second in duration.

Definition structure:
- EventId (string or an array of strings): Sound Event ID(s) to override.
- EventIdRegex (string or an array of strings): Sound Event ID(s) regular expressions, if matched the event will be overrided.
- AudioSelectionStrategy ("sequential" or "random"): Selection strategy for multiple override samples.

Additionally, "ns_print_played_sounds" convar has been added to print out sound events as they're requested.

Example mod: [dg.BoomMenuAccept.zip](https://github.com/R2Northstar/NorthstarLauncher/files/7807213/dg.BoomMenuAccept.zip)

### DIFF
--- a/NorthstarDedicatedTest/NorthstarDedicatedTest.vcxproj
+++ b/NorthstarDedicatedTest/NorthstarDedicatedTest.vcxproj
@@ -34,6 +34,7 @@
   </PropertyGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.props" />
   <ImportGroup Label="ExtensionSettings">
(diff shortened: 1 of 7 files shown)


x3Karma:
YESSS

HappyDOGE:
merged by request

PR #37, opened 2022-01-05
Add legal notices file to builds

JJRcop:
Like https://github.com/R2Northstar/Northstar/pull/101 but for the NorthstarLauncher repo instead

Adds a legal notices file to the root of the repo including all licenses and copyright information for third party libraries utilized by Northstar, along with Northstar's license itself. This file is copied into the build directory when compiling, making it easy to include in future published releases of Northstar.

If I missed anything, and this no longer has coverage of used third party libraries, let me know.

So, why in the NorthstarLauncher then?
 * Not fragmented from source
   * Being put in the Northstar repo distances it from the activity
   * This is where the referenced projects are actually used
   * Could lead to neglect if kept elsewhere
 * Properly version controlled along with releases
 * Copied along with all builds
   * Difficult to miss when packaging new releases
   * All future copies of the software comply with licenses by including them with binary builds, even developer ones

### DIFF
--- a/Northstar-Legal.txt
+++ b/Northstar-Legal.txt
@@ -1,0 +1,313 @@
+This file contains licenses and copyright notices of the Northstar client, and
+third-party libraries and software utilized by it.
+
+NORTHSTAR CLIENT
+================
+
+MIT License
+
+Copyright (c) 2021 R2Northstar
+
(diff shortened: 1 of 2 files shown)


abarichello:
Thanks

PR #39, opened 2022-01-06
Fix name spoofing

r3muxd:


BobTheBob9 (changes requested):
this is a great pr conceptually but i'd rather do the verification in authentication or somewhere where the web request isn't going to block main loop

barnabwhy:
If i'm not mistaken all this does is check if the origin username exists, and thus it surely only stops people setting non-existent usernames and not name spoofing as a whole

BobTheBob9:
> If i'm not mistaken all this does is check if the origin username exists, and thus it surely only stops people setting non-existent usernames and not name spoofing as a whole

yeah, but it helps at least, and prevents any invalid origin names from joining at all

BobTheBob9:
closing due to lack of activity

PR #41, opened 2022-01-07
LatencyFleX Implementation

KyleGospo:
Adds LatencyFleX, an open source and vendor agnostic input latency reduction technology similar to Nvidia's Reflex. Currently only works on Linux via Wine/Proton, but future versions may support Windows as well. Falls back to doing nothing if LatencyFleX isn't available.

See here for more information:
https://ishitatsuyuki.github.io/post/latencyflex/

### DIFF
--- a/NorthstarDedicatedTest/NorthstarDedicatedTest.vcxproj
+++ b/NorthstarDedicatedTest/NorthstarDedicatedTest.vcxproj
@@ -517,6 +517,7 @@
     <ClInclude Include="include\spdlog\version.h" />
     <ClInclude Include="keyvalues.h" />
     <ClInclude Include="languagehooks.h" />
+    <ClInclude Include="latencyflex.h" />
     <ClInclude Include="logging.h" />
     <ClInclude Include="main.h" />
     <ClInclude Include="masterserver.h" />
@@ -559,6 +560,7 @@
     <ClCompile Include="hooks.cpp" />
     <ClCompile Include="hookutils.cpp" />
     <ClCompile Include="keyvalues.cpp" />
+    <ClCompile Include="latencyflex.cpp" />
     <ClCompile Include="maxplayers.cpp" />
     <ClCompile Include="languagehooks.cpp" />
     <ClCompile Include="memalloc.cpp" />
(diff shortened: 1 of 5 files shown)


laundmo:
When you say thisnworks in Proton, does that mena it works on just Proton or Wine too?

as far as i understand this affects the rendering, do you know whether that is relevant for Dedicated Servers?

KyleGospo:
Completely irrelevant for Dedicated servers, it works with normal Wine as well so long as you install LatencyFleX

abarichello:
Can you provide data that this is indeed worth integrating into the project? I tried running your patch with their [custom mangohud config](https://github.com/ishitatsuyuki/LatencyFleX#mangohud-with-metric-support-optional) to graph latency but I couldn't get it to work, the graph just flatlines to zero when playing the game.

KyleGospo:
Can you add one of your logs with this enabled? To be clear, you did also add LFX=1 to your launch options, right?

KyleGospo:
Alright, took me a minute to get the best possible demonstration of this due to my CPU/GPU combo, but here we go:

With adaptive supersampling enabled and an FPS target of 5 and LatencyFleX disabled, I get 100% GPU usage with a fairly flat frametime:
Screenshot from 2022-01-11 14-06-20

Here is that same scene with LatencyFleX enabled, as you can see the GPU usage has dropped (To around the 95% number that is expected), and the frametime is now more variable (introduced jitter) as it tries to keep the frame queue clear.
Screenshot from 2022-01-11 14-06-53

You can also see evidence of it working when minimizing the game, with a low framerate upon returning until the algorithm is able to self-correct. I consider this a bug and am working to correct it.
Screenshot from 2022-01-11 14-07-16

KyleGospo:
The alt+tab bug mentioned above is now fixed upstream, no changes needed on our side.

abarichello (approved):
LGTM, I just need you to introduce a new section to the wiki explaining LatencyFlex and how to use it on Linux

KyleGospo:
Pull request created for the wiki at:
https://github.com/R2Northstar/NorthstarWiki/pull/18

Issue #23, opened 2021-12-30
Allow log file creation to be disabled


pg9182:

LMS007:
Hey @pg9182 , can you  fill in the feature request details here? Perhaps a why? Thanks!

JesperAppelberg:
Working on an implementation of this now. Will be a launch argument when it is fully implemented. As of right now however more work is needed. Expect it to be working in a few days though

Xavierhorwood:
I think I add what you wanted #44

abarichello:
Implemented in https://github.com/R2Northstar/NorthstarLauncher/commit/c9d66513a85d5325d3d002ca16c82d8302ec0ea2

pg9182:
Thanks!

PR #44, opened 2022-01-11
add disable logging files

Xavierhorwood:
does what #23 wants


abarichello:
I tried testing your patch but it crashes my game on launch for some reason

Xavierhorwood:
> I tried testing your patch but it crashes my game on launch for some reason

I will look it to that

Xavierhorwood:
> I tried testing your patch but it crashes my game on launch for some reason

ok so I found the issue with the crash and I am working on a fix

Xavierhorwood:
> 

I have fixed the crash but I had to remove the feature to disable dump files to fix the crash

abarichello:
Hey, your patch wasn't working because you need to check command line args after hooks are loaded. I created another commit (https://github.com/R2Northstar/NorthstarLauncher/commit/c9d66513a85d5325d3d002ca16c82d8302ec0ea2) to fix #23

PR #46, opened 2022-01-14
Allow UTF8 encoded server names and description, dedicated cmd window now titles the server name.

wolf109909:
Directly reading from the server cfg file directly to get server name and server description. This made server names and descriptions with Chinese characters work correctly. Demo:
WQ60RBLIOQO5Q%8HU9T{3DN
Clients don't need changes for this, as UTF8 decoding works just fine clientside.

I like to have server names displayed on the cmd title for better identification between multiple servers, so it works like that now. Demo:
image

This fork contains code from the NorthstarCN fork, which has to work with some master server modifications so I commented them off in this pr.


BobTheBob9 (changes requested):
parsing it directly from cfg seems like a bad idea, also not massive on the using namespace std and server name var could be in MasterServerManager rather than a global var
i think you've also included stuff for sending playername to masterserver as well here, which wasn't mentioned in your commit

i'd probably merge this if the extra stuff was dropped and you moved away from cfg parsing

wolf109909:
I'll close this here as the main fork already broke too much

Issue #50, opened 2022-01-20
dedicated: wineserver CPU usage is abnormally high


pg9182:
I'm looking into this in [`#wineserver cpu usage`](https://discord.com/channels/920776187884732556/933740994833289296) and will update this issue once I have definite details.

PR #53, opened 2022-01-20
Limit title update interval to 500ms

pg9182:
updates #50


pg9182:
On a 3-core i5-4460 VM, CPU time:

```
----------- ns      wine    ----------------------- ns      wine    - ns      wine
every 0.5s: 1:15.49 0:08.10 ... 50 real seconds ... 1:19.62 0:10.44 = 0:04.13 0:02.34
every tick: 1:20.66 0:10.76 ... 50 real seconds ... 1:25.64 0:14.05 = 0:04.98 0:03.29
                                                                      0:00.85 0:00.95
```

abarichello:
Thanks!

PR #58, opened 2022-01-26
Error logging for #148

MaxIsJoe:
Found a fix for https://github.com/R2Northstar/Northstar/issues/148

simply verify your game files twice if you ever have a game freeze when launching the northstar mod and the schannel error thing happens.
the game's console will alert you about this if it doesn't detect openssl and tell you about the fix if you ever encoutner this issue like i did.


abarichello (commented):
This isn't being logged to file btw, you should probably store the string result and log it later after spdlog initialization

PR #60, opened 2022-01-29
Add TextMsg logging

HappyDOGE:
CenterPrintAll serverside function now actually works, as well as some native engine prints (unknown command, some other commands etc.)
also adds a game-compatible bitbuf classes (read and write)


GeckoEidechse:
Wait, does that mean we can now hook chat messages on server?
E.g. have a mod listen for keywords like `!skip` and execute a corresponding command (vote skipping map)

PR #65, opened 2022-02-04
I/O + Mount All VPKs

EladNLG:
## Add I/O capability to the server squirrel VM.
- Squirrel can already overwrite existing files using the dev text buffer.

I think it may be nice to implement a buffer version later, but that is something that is beyond my skill and capabilities at the moment.

There are security issues with I/O capabilities that I was aware of - but I don't think that they matter since they are already achievable using the dev text buffer (which can access higher up directories - meaning that it can potentially override system files if `sv_cheats` is on.)

### Functions
`ReadFile( string modName, string path, string fileName )` - Reads a file at the specified path.
`WriteFile( string modName, string path, string fileName, string content )` - Writes/Creates a file (and it's folders if no such file exists). Places mod files separately to prevent conflicts and overwrites.
`GetAllFilesInFolder( string modName, string path )` - returns all files in a folder.
`GetAllFoldersInFolder( string modName, string path, bool recursive = false )` - returns all folders in a folder, recursively if set to.

### **Why?**
- Allows custom mods to save things _locally._ Whilst I know about pdiff and it's nice - it's also editable by clients by having them connect to a local server of their own and receiving the pdata, allowing for them to cheat.
- You cannot purely read files - the best you can normally do is write to a NUT file in the mod folder and reload the game, and then calling a function inside that file.

## Mount all VPKs at once, always.
This is functionality similar to Icepick that mounts all the vpks at once. Though precaching is still done manually to prevent any performance impact<sup>1</sup> (from what I noticed, at least).
### **Why?**
- Allows Multiplayer weapons in Singleplayer maps (for Co-Op)
- Allows scripts to use props from other maps (for making level editors more powerful and making your map of choice depend on aesthetics, not what props it has available.)

<sup>1</sup>There seems to be a noticable delay at startup of ~30 seconds? Though no impact in loading screens/in-game.

## Remove SpacesInLineCommentPrefix from .clang-format

My visual studio does not stop complaining about it, so away it goes.


EladNLG:
need to quickly add rpak loading ;)

wolf109909:
I like this idea! :D

PR #70, opened 2022-02-08
make playlist overrides work on private match

RomeoCantCode:
By removing IsDedicated check it allows to use playlistvaroverrides on servers launched through the game (private match / listen server).
It seems it does not have any other side effects on the client so far.


abarichello (approved):

abarichello:
Thanks

Issue #30, opened 2022-01-01
Compress old logfiles


lolPants:
At startup locate old logfiles and gzip them to save disk space.

abarichello:
https://github.com/R2Northstar/NorthstarLauncher/pull/73

PR #73, opened 2022-02-10
Compresses old logs

Legonzaur:
Uses [zlib ](https://zlib.net/) and [gzip-hpp](https://github.com/mapbox/gzip-hpp) to compress old `.txt` and `.dmp` logs

See #30


abarichello (approved):

GeckoEidechse:
Uhm, why did we merge this? From my understanding this is run on every Northstar launch, meaning that if any player crashes, relaunches the game and then later comes back to check the old log, they won't be able to check it out without knowing how to open gzip files.
It would make more sense to run it only if a log file is older then X days (preferably some time frame where the log likely concerns an older Northstar release) but at this point we could just delete the old logs (i.e. doing some kind of log rotation).

GeckoEidechse:
Also we should do a safety check that the file name roughly matches `nslogXYZ.txt` (we changed the date format a while back) to prevent accidentally messing with some other text files the player may have put into the `R2Northstar\logs` directory for one reason or another.

abarichello:
> Uhm, why did we merge this? From my understanding this is run on every Northstar launch, meaning that if any player crashes, relaunches the game and then later comes back to check the old log, they won't be able to check it out without knowing how to open gzip files. It would make more sense to run it only if a log file is older then X days (preferably some time frame where the log likely concerns an older Northstar release) but at this point we could just delete the old logs (i.e. doing some kind of log rotation).

That shouldn't be a problem, if they can't figure out 7zip they probably shouldn't be reading the log files anyways.

On another note the files aren't being compressed correctly, I got some errors that I thought were being caused by my Linux installation but apparently it's also happening on Windows.

GeckoEidechse:
Revert and let @Legonzaur fix?

GeckoEidechse:
Also I'd still prefer for it to only compress logs older then X days, to avoid constantly having to uncompress recent logs when investigating crashes during testing.

Together with doing pattern matching on file name, so `nslogYYYY-MM-DD hh-mm-ss.txt` and `nsdumpYYYY-MM-DD hh-mm-ss.dmp`.
(I think we can safely ignore logs created in the older format)

abarichello:
I'll investigate more tomorrow, I think the fix is because the outfile isn't being written as binary.

PR #86, opened 2022-02-18
Add check for player name correctness

barnabwhy:
Relies on R2Northstar/NorthstarMasterServer#42

Fails player auth on connect attempt if the connecting player has a different username to the one expected (as supplied by the master server)


barnabwhy:
This took too much effort

barnabwhy:
it seems git did some fuckery with the if statement and duplicated it?

NorthstarDedicatedTest/serverauthentication.cpp
```diff
@@ -152,39 +155,43 @@ bool ServerAuthenticationManager::AuthenticatePlayer(void* player, int64_t uid,
...
-		{
-			authFail = false;
-			// uuid
-			strcpy((char*)player + 0xF500, strUid.c_str());
 
-			// reset from disk if we're doing that
-			if (m_bForceReadLocalPlayerPersistenceFromDisk && !strcmp(authData.uid, g_LocalPlayerUserID))
+		if (!strcmp(strUid.c_str(), authData.uid) &&
+			!strcmp(name, authData.name)) // connecting client's uid is the same as auth's uid and name matches supplied one
+			if (!strcmp(strUid.c_str(), authData.uid) &&
```
barnabwhy, line 161: why is it duplicated here?

barnabwhy:
i’m going to shit myself at this rate
i hate git

barnabwhy:
I think I managed to fix the duplicated if statement but it probably needs testing, i can’t until tomorrow afternoon so if someone else can please do

NorthstarDedicatedTest/serverauthentication.cpp
```diff
@@ -152,7 +155,9 @@ bool ServerAuthenticationManager::AuthenticatePlayer(void* player, int64_t uid,
 	{
 		// use stored auth data
 		AuthData authData = m_authData[authToken];
-		if (!strcmp(strUid.c_str(), authData.uid)) // connecting client's uid is the same as auth's uid
+
+		if (!strcmp(strUid.c_str(), authData.uid) &&
+			!strcmp(name, authData.name)) // connecting client's uid is the same as auth's uid and name matches supplied one
```
pg9182, line 160: This check should be `!strcmp(strUid.c_str(), authData.uid) && (!*authData.name || !strcmp(name, authData.name))` to preserve compatibility with old master servers and to fail open.

pg9182 (commented):
Looks good in general.

NorthstarDedicatedTest/serverauthentication.cpp
barnabwhy: Fails open now :)

NorthstarDedicatedTest/serverauthentication.cpp
```diff
@@ -114,6 +114,9 @@ void ServerAuthenticationManager::StartPlayerAuthServer()
 					strncpy(newAuthData.uid, request.get_param_value("id").c_str(), sizeof(newAuthData.uid));
 					newAuthData.uid[sizeof(newAuthData.uid) - 1] = 0;
 
+					strncpy(newAuthData.name, request.get_param_value("username").c_str(), sizeof(newAuthData.name));
+					newAuthData.name[sizeof(newAuthData.name) - 1] = 0;
```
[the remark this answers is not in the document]
barnabwhy: Changed that

barnabwhy:
It now just changes people's player names to match their username instead of kicking

NorthstarDedicatedTest/serverauthentication.cpp
```diff
@@ -141,6 +144,19 @@ void ServerAuthenticationManager::StopPlayerAuthServer()
 	m_playerAuthServer.stop();
 }
 
+char* ServerAuthenticationManager::VerifyPlayerName(void* player, char* authToken, char* name)
+{
+	AuthData authData = m_authData[authToken];
+
+	bool nameAccepted = (!*authData.username || name == "" || !strcmp(name, authData.username));
```
pg9182, line 151: Uhhh... `name == ""` is incorrect in C/C++ for char pointers (it means you're checking if name points to the const string "", which will essentially never be true). You want to do `!*name` instead.

pg9182 (changes requested):

NorthstarDedicatedTest/serverauthentication.cpp
barnabwhy: oh i just realised name shouldn't be checked here
should !*authData.username cover when it's missing and ""?

> Uhhh... `name == ""` is incorrect in C/C++ for char pointers (it means you're checking if name points to the const string "", which will essentially never be true). You want to do `!*name` instead.

pg9182: >  should !*authData.username cover when it's missing and ""?

Yes, since `authData.username` is a fixed-size char buffer and strncpy will always append the null terminator unless the source is larger than the destination (in which case you'll still have it anyways since you manually set the last byte to null).

BobTheBob9 (changes requested):
can you verify this won't have issues if a client tries to connect without authenticating? doesn't seem like VerifyPlayerName ever checks whether auth data is actually present for the connecting client

barnabwhy:
@BobTheBob9 Added the fix, seems to work just fine

barnabwhy:
Why did the build fail?

pg9182:
It was merged with #90 (4f28a07d0562ca905bbcdb010b69604c330194bb).

NorthstarDedicatedTest/serverauthentication.cpp
```diff
@@ -141,6 +144,28 @@ void ServerAuthenticationManager::StopPlayerAuthServer()
...
+
+	if (!m_authData.empty() && m_authData.count(std::string(authToken)))
+	{
+		AuthData authData = m_authData[authToken];
+
+		bool nameAccepted = (!*authData.username || !strcmp(name, authData.username));
+
+		if (!nameAccepted && g_MasterServerManager->m_bRequireClientAuth && !CVar_ns_auth_allow_insecure->m_nValue)
+		{
+			strcpy(name, authData.username);
```
pg9182, line 159: Use strncpy and leave a comment explaining the 64-character limit. This will reduce the risk of regressions being introduced if the auth code is refactored.

NorthstarDedicatedTest/serverauthentication.cpp
```diff
@@ -141,6 +144,28 @@ void ServerAuthenticationManager::StopPlayerAuthServer()
...
+		}
+		return name;
+	}
+	else
+	{
+		return name;
+	}
```
pg9182, line 166: Unnecessary else statement. Also, you might as well just put the return at the end and take line 161 out of the block.

pg9182 (changes requested):

pg9182 (commented):
Looks good in general now, but I haven't actually tested it.

barnabwhy:
format check failure :(

Issue #82, opened 2022-02-17
Northstar crashes if R2Northstar/logs does not exist


pg9182:
As of 1a7b41cc722c05eb97006d29e124c17559c45324 (unreleased).

I haven't checked it yet, but my hunch is that it's a regression related to this https://github.com/R2Northstar/NorthstarLauncher/pull/73/files#diff-fc1e49c087e0aa0e7e432a10288ddbd465d6ed30ccb5946c2deea666386e974aR52.

GeckoEidechse:
@Legonzaur

Legonzaur:
Best thing to do would be to revert #76 and #73 and merge log rotation when it's done

GeckoEidechse:
As the relevant commit was reverted already I think this issue can be closed :P

PR #88, opened 2022-02-20
Log rotation

Legonzaur:
In following to #82, #76 and #73 

Removes .txt files and .dmp files older than x days (7 by default) based on file name

### DIFF
--- a/NorthstarDedicatedTest/logging.cpp
+++ b/NorthstarDedicatedTest/logging.cpp
@@ -10,10 +10,12 @@
 #include <Psapi.h>
 #include <minidumpapiset.h>
 #include "configurables.h"
+#include "logrotation.h"
 
 // This needs to be called after hooks are loaded so we can access the command line args
 void CreateLogFiles()
 {
+	logRotation(7);
 	if (strstr(GetCommandLineA(), "-disablelogs"))
 	{
 		spdlog::default_logger()->set_level(spdlog::level::off);
--- a/NorthstarDedicatedTest/logrotation.cpp
+++ b/NorthstarDedicatedTest/logrotation.cpp
@@ -1,0 +1,80 @@
+#include "pch.h"
+#include <filesystem>
+#include <string>
+#include <time.h>
+#include <iostream>
+#include <regex>
+#include "configurables.h"
+
+namespace fs = std::filesystem;
+using namespace std;
+
+const int seconds_in_day = 86400;
+
+double getLogAge(const fs::path path)
+{
+	time_t logtime;
+	time_t currenttime;
+	struct tm timeinfo = {0};
+	int year, month, day;
+	string filename = path.filename().string();
+	// Check if log has a correct name
+	if (regex_search(filename, regex("nslog([0-9]{2}-){2}[0-9]{4}\ [0-9]{2}(-[0-9]{2}){2}.txt")))
+	{
+		// Log naming before 1.4.0, which released on January 06 2022
+		year = stoi(filename.substr(11, 4));
+		month = stoi(filename.substr(8, 2));
+		day = stoi(filename.substr(5, 2));
+	}
+	else if (regex_search(filename, regex("nsdump([0-9]{2}-){2}[0-9]{4}\ [0-9]{2}(-[0-9]{2}){2}.dmp")))
+	{
+		// Dumps
+		year = stoi(filename.substr(12, 4));
+		month = stoi(filename.substr(9, 2));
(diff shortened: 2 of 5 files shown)


GeckoEidechse:
Personally I'd go for a higher value like 30 days before deleting logs (the size they accumulate is not very big) to ensure we don't lose any potentially relevant logs (beyond that we probably already release a new version of Northstar anyway, making old logs obsolete).

GeckoEidechse:
Also to avoid any regressions this time, let's test this with
- [ ] log directory missing
- [ ] device current time set to 1st Jan 2022
- [ ] device current time set to 31th December 2022
- [ ] ???
to catch possible edge cases ^^

NorthstarDedicatedTest/logging.cpp
```diff
@@ -10,10 +10,12 @@
 #include <Psapi.h>
 #include <minidumpapiset.h>
 #include "configurables.h"
+#include "logrotation.h"
 
 // This needs to be called after hooks are loaded so we can access the command line args
 void CreateLogFiles()
 {
+	logRotation(7);
```
pg9182, line 18: This should not happen if logs are disabled, and there should also be a separate option to disable log rotation (possibly passing it the number of days, with 0 being disabled).

NorthstarDedicatedTest/logrotation.cpp
```diff
@@ -0,0 +1,80 @@
+#include "pch.h"
+#include <filesystem>
+#include <string>
+#include <time.h>
+#include <iostream>
+#include <regex>
+#include "configurables.h"
+
+namespace fs = std::filesystem;
+using namespace std;
```
pg9182, line 10: ^ ...

NorthstarDedicatedTest/logrotation.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+double getLogAge(const fs::path path)
+{
+	time_t logtime;
+	time_t currenttime;
+	struct tm timeinfo = {0};
+	int year, month, day;
+	string filename = path.filename().string();
+	// Check if log has a correct name
+	if (regex_search(filename, regex("nslog([0-9]{2}-){2}[0-9]{4}\ [0-9]{2}(-[0-9]{2}){2}.txt")))
```
pg9182, line 22: The regular expressions should be compiled ahead of time, or at least only once.

NorthstarDedicatedTest/logrotation.cpp
```diff
@@ -0,0 +1,80 @@
...
+		return 0;
+	}
+
+	// Only reads days
+	timeinfo.tm_year = year - 1900;
+	timeinfo.tm_mon = month - 1;
+	timeinfo.tm_mday = day;
+
+	logtime = mktime(&timeinfo);
+	time(&currenttime);
```
pg9182, line 55: I would personally just return the time, then do the comparison in logRotation.

NorthstarDedicatedTest/logrotation.cpp
```diff
@@ -0,0 +1,80 @@
...
+	return difftime(currenttime, logtime);
+}
+
+void logRotation(int days = 7)
```
pg9182, line 59: This is incorrect; default arguments belong in the declaration (i.e., the header).

NorthstarDedicatedTest/logrotation.cpp
```diff
@@ -0,0 +1,80 @@
...
+{
+	string path = GetNorthstarPrefix() + "/logs";
+	struct stat info;
+	// Check if we can access log directory
+	if (stat(path.c_str(), &info) != 0 || !(info.st_mode & S_IFDIR))
+	{
+		return;
+	}
```
pg9182, line 67: Just check if the directory_iterator is valid rather than doing a separate check.

pg9182 (changes requested):
I don't personally think Northstar should handle log rotation, but if it must, there are a few problems with this implementation.

GeckoEidechse:
> I don't personally think Northstar should handle log rotation, but if it must, there are a few problems with this implementation.

Do it in Viper/VTOL instead?

Legonzaur:
Thank you for your advice, I'll see what I can do on Viper or VTOL instead then

NorthstarDedicatedTest/logrotation.cpp
abarichello: I think what pg meant is that these namespace shortening are helpful but they shouldn't really be introduced in the code base since it's a Good Practice ™️

PR #90, opened 2022-02-20
ConVar class rebuild

Mauler125:
This pull request contains the rebuild and mapped out ConVar class.
The codebase has been adjusted to use the new system instead.

This pull request also contains a full fix for (the now) currently added virtual function methods of 'CCvar' 
to be used throughout the SDK. See 'gameutils.cpp' L77 for example usage.
You can add more CCvar class methods to the SDK by adding the vtable index and prototype in 'cvar.cpp'.

I also added the 'Color' struct which is used in the ConVar class for returning values.

Update:
I also added the mapped out ConCommand struct. Note that this was taken from Apex, i expect all members to align with Titanfall 2 considering the sizes where identical and the command classes not being modified extensively in general. But if something does happen to misalign (unlikely), single step the ConCommand registration function called by the ConCommand constructor in engine.dll and align accordingly.

With this mapped out, consider rebuilding the constructor for ConCommands in the SDK instead.


NorthstarDedicatedTest/NorthstarDedicatedTest.vcxproj.filters
```diff
@@ -121,6 +121,12 @@
     <Filter Include="Header Files\include\libcurl">
       <UniqueIdentifier>{ea1e17a6-40b7-4e1b-8edb-e9ae704ce604}</UniqueIdentifier>
     </Filter>
+    <Filter Include="Source Files\Shared\Math">
+      <UniqueIdentifier>{59b0f68f-daa7-4641-b6fa-8464b56da2bb}</UniqueIdentifier>
+    </Filter>
+    <Filter Include="Header Files\Shared\Math">
+      <UniqueIdentifier>{44a83740-9d70-480d-9a7a-43b81f8eab9e}</UniqueIdentifier>
+    </Filter>
```
pg9182, line 129: Do we really want even more filters? Although, I guess we can just deal with them when we do the big refactor.

pg9182 (approved):
Code looks good and I like the implementation. PR doesn't seem to cause any obvious issues, and I trust @Mauler125. The formatting needs to be fixed, though.

NorthstarDedicatedTest/NorthstarDedicatedTest.vcxproj.filters
Mauler125: For the extra files i added to this project, i don't think there was a better filter. I could place them elsewhere (in a existing filter) if desired, but i recommend a full project restructure regardless in the future.

pg9182 (approved):

abarichello (approved):

PR #94, opened 2022-02-23
Added Global ban stuff

wolf109909:
it works
maybe :3

### DIFF
--- a/NorthstarDedicatedTest/bansystem.cpp
+++ b/NorthstarDedicatedTest/bansystem.cpp
@@ -218,9 +218,8 @@ void ClearBanlistCommand(const CCommand& args) { g_ServerBanSystem->ClearBanlist
 void InitialiseBanSystem(HMODULE baseAddress)
 {
 	g_ServerBanSystem = new ServerBanSystem;
-	g_ServerBanSystem->m_CurrentMessageIndex = 0;
 	g_ServerBanSystem->OpenBanlist();
-	IsBanMessageFileParsed = g_ServerBanSystem->ParseLocalBanMessageFile();
+	g_ServerBanSystem->ParseLocalBanMessageFile();
 	RegisterConCommand("ban", BanPlayerCommand, "bans a given player by uid or name", FCVAR_GAMEDLL);
 	RegisterConCommand("unban", UnbanPlayerCommand, "unbans a given player by uid", FCVAR_NONE);
 	RegisterConCommand("clearbanlist", ClearBanlistCommand, "clears all uids on the banlist", FCVAR_NONE);
--- a/NorthstarDedicatedTest/bansystem.cpp
+++ b/NorthstarDedicatedTest/bansystem.cpp
@@ -5,11 +5,134 @@
 #include "concommand.h"
 #include "miscserverscript.h"
 #include <filesystem>
+#include <iostream>
 #include "configurables.h"
+#include "masterserver.h"
+/*
+const char* BanMessages[] =
+{
+	"You farded and shidded, appeal at discord.gg/Northstar",
+	"Authentication error, please enable mat_global_lighting 1",
+	"uh owh! we madwe an oopsie woopsie >.< we ar workin vewy hawd two fwix it uwu",
+	"Error code: 80085. Please check https://laund.moe/ to solve your issue.",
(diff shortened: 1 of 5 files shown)


stllok:
fking 666

wolf109909:
EVERYTHING IS FIXED AND SHOULD BE WOKRING NOW UWU

barnabwhy:
loving the branch name :trollface:

wolf109909:
> loving the branch name :trollface:

XD

GeckoEidechse:
As stated in discussion on Discord I'm not to fond of having a "single source of truth" banlist due anyone finding themselves on that list pointing fingers at core Northstar contributors. Instead we should somehow/somewhere provide a public list of known offenders that contains their category of offence as well as other metadata (a sort of list of shame) and then enable server hosters to use a specific list using some commandline flag with the default being `none`.

The idea to this being that instead of having to blame core Northstar contributors, any offenders would have to instead complain to server hosters of which there are more of and as such the offenders do not have a single point of blame.

Additionally I'm growing increasingly found of "soft-bans" where offenders get nerfed based on their offence.
- So cheaters with aimbot would get massive damage reductions so that anyone attacked by them would have more then enough time to react and counteract.
- Speedhackers would get slowed down by being rubber-banded to a previous location.
- Chat spammers would get shadow-banned, so all their messages would never reach any other player.
- etc.

wolf109909:
Hi gecko! Thanks for the reply! :3
Initially I used json for the job, but later I realize for me I only need the uid part cuz we usually don't encounter other hacks than regular aimbots that should have 100% being banned. 
The other reason for parsing raw text stream is that I think json is kinda slow if it gets massive, thus I made it extremely fast and simple. 
If we move this to json and make a soft ban system base on that, it's better to also make a visual backend for editing the json file. (Which is actually one more reason for me to not use json here, I have to manually construct json string with my human hand if I wanna ban someone xDD)
But seriously speaking, I really like the softban idea. We can also do stuff like "ppl being punished can only see the servers that are meant for them to play(like a ttk extremely long, unsatisfying game.)"
For me , all I can do is to move this back to json format, make a callback in squirrel( which was actually done but removed) so you can have easier access to the ban informations. 
But I'm not any good at programming so whether this will finish really depends on my luck and free time uwu

GeckoEidechse:
I'm not sure if JSON is needed specifically but any form of "banlist" we provide (again I'd argue server hosters need to enable it specifically) should be fine grained so that server hosters can make informed decisions (compared to just listing UIDs) and follow some clearly defined standards so that others can create and host their own banlists, allowing server hosters to get the content from other/multiple sources. ^^

BobTheBob9:
personally, i think the best idea would just be to fail authentication for particularly aggregious cases, where there's pretty much no doubt that the person should get a full ban, i don't personally see much of a reason to provide a global banlist that's enforced on gameserver, if the assumption is that most servers will be using that banlist anyway

BobTheBob9:
> As stated in discussion on Discord I'm not to fond of having a "single source of truth" banlist due anyone finding themselves on that list pointing fingers at core Northstar contributors. Instead we should somehow/somewhere provide a public list of known offenders that contains their category of offence as well as other metadata (a sort of list of shame) and then enable server hosters to use a specific list using some commandline flag with the default being `none`.
> 
> The idea to this being that instead of having to blame core Northstar contributors, any offenders would have to instead complain to server hosters of which there are more of and as such the offenders do not have a single point of blame.
> 
> Additionally I'm growing increasingly found of "soft-bans" where offenders get nerfed based on their offence.
> 
>     * So cheaters with aimbot would get massive damage reductions so that anyone attacked by them would have more then enough time to react and counteract.
> 
>     * Speedhackers would get slowed down by being rubber-banded to a previous location.
> 
>     * Chat spammers would get shadow-banned, so all their messages would never reach any other player.
> 
>     * etc.

i'm not sure softbans would be such a good idea to be honest, while stuff like rubberbanding speedhackers is a good idea (though, not entirely sure it falls within the realm of softbans really, more just exploit prevention)
however, stuff like reducing damage i think should be left more to mods, rather than being shipped with northstar by default

GeckoEidechse:
Closing cause we decided (a while back) to not do global bans but instead want to make it easier to "import" bans. ^^

PR #98, opened 2022-02-25
Patch 1

r3muxd:
v,


BobTheBob9:
while the commit itself does seem pretty good (i'm assuming this is primarly cvar help/usage text, find command, that sorta thing), this really needs some cleaning up to be merged i think, especially since alot of the code isn't exactly self-documenting with the number of ida variable names in it

PR #100, opened 2022-03-04
RCON (Server & Client)

Mauler125:
This pull request contains:
* Full CNetAdr class rewrite in modern C++ (protocol agnostic)
* Full CSocketCreator rebuild (based on Valve's implementation, but protocol agnostic)
* RCON for server (dedicated only)
* RCON for client
* Non-game net console client to connect to dedicated server without needing any files of the game itself (just the netcon, link MFC static to the netconsole if maximum portability is desired, since then user doesn't need to install vc runtime)

Information regarding using RCON:
SERVER:
* The password must be at least 8 characters long (ConVar: `rcon_password`)
* To disable RCON, simply keep `rcon_password` empty.
* To send logs to connected net consoles, set `sv_rcon_sendlogs` to `1`, or send a `SERVERDATA_REQUEST_SEND_CONSOLE_LOG` message
* The RCON server will bind to the same port as the `hostport` ConVar used by the game for UDP sockets, both IPv4 and IPv6 (dual stack)
* The server will automatically ban IP's of spammers or hackers. It will not kick them out of game, just remove access to server remote access. Game kicking/banning could be implemented if desired.
* It is recommended that the user sets a `sv_rcon_whitelist_address`, this will never get banned for failed attempts, or get 'cleaned' when max sockets (defined by `sv_rcon_maxsockets`) gets reached.
* The server will automatically close the listen socket once a net console has successfully authenticated with the server
* The listen socket gets reopened if connection to net console gets closed in any way (for example, when your game crashes or when you issued a disconnect)
* Auto executes `rcon_server.cfg` on launch. You can put your configs in here.

CLIENT:
* To connect to server, in the console set 'rcon_address' to the address of the RCON server you wish to connect to (default: [127.0.0.1]:37015)
* Then enter command `rcon`. This will try to establish a connection to address stored in `rcon_address` ConVar.
* To authenticate with the server: `rcon PASS yourRCONpassword321`.
* To send commands to server: `rcon "your commands here"`.
* To disconnect from RCON server: `rcon_disconnect`.
* * Auto executes `rcon_client.cfg` on launch. You can put your configs in here.

NETCONSOLE:
* Same as `CLIENT`, but in this case you do not need to use the `rcon` command prefix to send queries to RCON server.
* Follow instructions in the terminal.

Additional information:
* Port forwarding is required for this to work. Its the same range as the UDP protocol (37015-37020), but it has to be opened for TCP as well (RCON ports are equal to the UDP port of the game, which you could retrieve with the 'status' command or various tools available for Windows).


wolf109909:
I love this. Thank you Amos :D

Vexbane:
Thanks for all your hard work Amos!

abarichello:
Had some problems while testing this, changing convars worked fine but running server scripts gave me this error: 
image
Left: dedicated server output | Right: NetConsole

abarichello:
Inviting others to test/review since there's a lot of code to cover

Mauler125:
> Had some problems while testing this, changing convars worked fine but running server scripts gave me this error: image Left: dedicated server output | Right: NetConsole

Hi, I fixed this issue on the latest commit. It should now work properly on commands, setting convars and running scripts
image

GeckoEidechse:
As someone with very little Windows dev experience, how do I actually go about building the NetConsole part? ^^

Mauler125:
> As someone with very little Windows dev experience, how do I actually go about building the NetConsole part? ^^

NetConsole gets build among with the rest of the solution. The resulting binary should be 'netcon.exe'.

GeckoEidechse:
Aight, so there's a lot of desire and need to get this merged so I drafted up a bit of a rough roadmap on what we should still take into consideration for merging:

- [ ] Mention NetConsole in `BUILD.md`

Build instructions aren't really needed as it just builds as part of Launcher when creating the build in VS. We should however update `BUILD.md` as part of this PR to include mention of NetConsole.


- [ ] Usage instructions for server hosters.

Assuming they have already downloaded the necessary files, what would be the steps they need to perform to get everything setup? Basically a quickstart guide to go from working server -> can RCON into server.

These instructions would then be added to wiki on merge to tell server hosters how to use RCON interface.


- [ ] Maybe some way of disabling RCON via cli flag and/or at compile time.

I'm thinking, what happens if we for some reason find critical bug in RCON?
We would obviously bump min version on master server to next release that fixes it so that unpatched servers cannot expose themselves to clients via server browser.
But what if no-ones available to fix the underlying issue. Say you're on holiday for example. If we can disable RCON at compile time we can make a "hotfix" that just disables it until we fix the underlying bug.
Ofc, if RCON is not exposed to outside via port-forwarding by hoster, this is not really an issue.

- [ ] Checking CI to not include NetConsole in release files.

This one's for me. Obviously we want to distribute RCON in some pre-compiled fashion but we shouldn't put it in the main release zip as its contents end up in the `Titanfall2` folder on players' machines. NetConsole should be kinda separate to avoid cluttering players `Titanfall2` folder.

pg mentioned wanting to change the way we build releases where we build inside Launcher and Mods on tagging and in release repo we just pull in the files from those releases on tag. This would help with distributing NetConsole for those that need it as we can just link to Launcher release files.\
However it's still gonna be a bit until that happens, so in the meantime I guess we can just chug the pre-compiled exe somewhere, assuming it doesn't change too often.


- [ ] Some preliminary testing using this branch with server hosters.

Basically, we make an "RCON release" for server hosters to test it and play around with it to get feedback. We do need the instructions mentioned above for it though.

- [ ] Reviews

I need 1-2 people with decent understanding of the Northstar codebase to review this PR

Mauler125:
RCON is fully disabled if password length is < 8 characters (user could leave the password ConVar 'rcon_password' empty), there is a boolean that gets tested every frame (CRConServer::m_bInitialized), it gets set by CRConServer::Init() when it has been properly initialized (a password of min 8 chars). No socket will be created and no parts of the RCON system is ran if password length was < 8 and/or m_bInitialized is false. You could still comment 'g_pRConServer->RunFrame()' to compile it out, but the system for disabling it very simple and will never fail so this isn't necessary.

For distributing 'netcon.exe' and perhaps any other uncommon binary, I recommend just putting them in the 'bin' folder to reduce clutter (this is what we do in r5reloaded). I also recommend putting the Northstar dll's etc in the bin folder since its irrelevant to the end user. Only the executable is relevant.

The instructions for running RCON is in the commit description, no external files are required. For anyone willing to format it for the wiki, the above should be all you need to get it running.

I updated the commit description to also mention port forwarding is required, its the same as the game's UDP ports, but TCP also needs to be forwarded. The same range (37015-37020) or further if you host more servers.

RaVenInTheDark:
Apologies if this is the wrong spot, but how easy would it be to rcon through some sort of external script, rather than netconsole/a client?

NorthstarDedicatedTest/squirrel.cpp
```diff
@@ -241,6 +243,17 @@ template <ScriptContext context> SQInteger SQPrintHook(void* sqvm, char* fmt, ..
...
 		spdlog::info("[{} SCRIPT] {}", GetContextName(context), buf);
+
+		if (IsDedicated())
+		{
+			if (CVar_sv_rcon_sendlogs->GetBool())
+			{
+				char sendbuf[1024] {};
+
+				snprintf(sendbuf, sizeof(sendbuf), "[%s SCRIPT] %s", GetContextName(context), buf);
+				g_pRConServer->Send(sendbuf);
```
BobTheBob9, line 254: why not just do this our spdlog sink to capture all logs, rather than doing it manually for different log sources?

Mauler125: > why not just do this our spdlog sink to capture all logs, rather than doing it manually for different log sources?

Where is your sink, is it by any change 'SourceConsoleSink::sink_it_()'?

BobTheBob9: oh nvm actually, you'd have to create a new spdlog sink
SourceConsoleSink is the console sink used for the vgui console, shouldn't be super hard to modify the code used for initializing it to make a new sink (https://github.com/R2Northstar/NorthstarLauncher/blob/main/NorthstarDedicatedTest/sourceconsole.cpp#L39)

laundmo:
> The server will automatically ban IP's of spammers or hackers.

Are we sure this behaviour is good? With how often home IPs can change, CGNAT, VPNs etc, IP bans can easily lead to banning the wrong IP. 

Would also like some clarification of "automatically" here: how does it detect spammers and hackers? why was automatic spam and hacker detection part of a rcon PR?

RaVenInTheDark:
> > The server will automatically ban IP's of spammers or hackers.
> 
> Are we sure this behaviour is good? With how often home IPs can change, CGNAT, VPNs etc, IP bans can easily lead to banning the wrong IP.
> 
> Would also like some clarification of "automatically" here: how does it detect spammers and hackers? why was automatic spam and hacker detection part of a rcon PR?

IMO this should be optional, with an option to clear the bans on reboot

Mauler125:
> > > The server will automatically ban IP's of spammers or hackers.
> > 
> > 
> > Are we sure this behaviour is good? With how often home IPs can change, CGNAT, VPNs etc, IP bans can easily lead to banning the wrong IP.
> > Would also like some clarification of "automatically" here: how does it detect spammers and hackers? why was automatic spam and hacker detection part of a rcon PR?
> 
> IMO this should be optional, with an option to clear the bans on reboot



> 

Bans are cleared upon reboot, they are stored in a runtime vector.
As for the second question, the system automatically bans the user from remote console access when they entered the wrong password, or ignored the no-auth message too often (which can be controlled by the ConVar's ''sv_rcon_maxfailures" and "sv_rcon_maxignores"), the server owner doesn't need to perform any actions manually to ban a user. You can set those ConVar's to a very high number if you wish to effectively 'disable' it (no netconsole (rcon client) data is remembered upon reboot).

Mauler125:
> > The server will automatically ban IP's of spammers or hackers.
> 
> Are we sure this behaviour is good? With how often home IPs can change, CGNAT, VPNs etc, IP bans can easily lead to banning the wrong IP.
> 
> Would also like some clarification of "automatically" here: how does it detect spammers and hackers? why was automatic spam and hacker detection part of a rcon PR?

The system automatically bans the user from remote console access when they entered the wrong password, or ignored the no-auth message too often (which can be controlled by the ConVar's ''sv_rcon_maxfailures" and "sv_rcon_maxignores"), the server owner doesn't need to perform any actions manually to ban a user. You can set those ConVar's to a very high number if you wish to effectively 'disable' it (no netconsole (rcon client) data is remembered upon reboot).

Mauler125:
> > > The server will automatically ban IP's of spammers or hackers.
> > 
> > 
> > Are we sure this behaviour is good? With how often home IPs can change, CGNAT, VPNs etc, IP bans can easily lead to banning the wrong IP.
> > Would also like some clarification of "automatically" here: how does it detect spammers and hackers? why was automatic spam and hacker detection part of a rcon PR?
> 
> IMO this should be optional, with an option to clear the bans on reboot

Bans are cleared upon reboot, they are stored in a runtime vector. For disabling, you can set the 'sv_rcon_maxfailures' or 'sv_rcon_maxignores' ConVar's to a very high number, or modify the code to implement a disable feature. I personally see no reason to implement one, since this occurrence is rare, and in case it happens, just restart the server (you are supposed to restart them every 3 hours anyways).

NorthstarDedicatedTest/squirrel.cpp
Mauler125: I will take a look into creating a sink dedicated to this.

laundmo:
> The system automatically bans the user from remote console access when they entered the wrong password, or ignored the no-auth message too often

Ah, this was a misunderstanding then. In the PR text i understood "spammers and hackers" to mean people spamming ingame chat or using a "hacked" game to gain a unfair advantage. I didn't even think you could mean brute force attacks with that.

GeckoEidechse (changes requested):
Found a bug in testing where starting a match is not possible on dedicated server from game client. Video:

https://user-images.githubusercontent.com/40122905/177223626-01f9b0c7-4adc-49cd-a056-db978b4c05fb.mp4

To reproduce:
1. Install Northstar v1.9.0
2. Get CI artifacts and copy `Northstar.dll` and `Northstar.exe` into Titanfall2 folder
3. Set `rcon_password` and `ns_server_password`
4. Start dedicated server via `r2ds.bat`
5. Start Northstar client on same machine
6. Join your own dedicated server via server browser
7. Press "start match"
8. Countdown timer gets stuck

**EDIT:**

If I set server to launch into game, e.g. by setting `+setplaylist ctf +mp_gamemode ctf +map mp_glitch` in dedi startup args and then try to join from client, I then error out on client with:

snapshot0015

Mauler125:
> Found a bug in testing where starting a match is not possible on dedicated server from game client. Video:
> out.mp4
> 
> To reproduce:
> 
>     1. Install Northstar v1.9.0
> 
>     2. Get CI artifacts and copy `Northstar.dll` and `Northstar.exe` into Titanfall2 folder
> 
>     3. Set `rcon_password` and `ns_server_password`
> 
>     4. Start dedicated server via `r2ds.bat`
> 
>     5. Start Northstar client on same machine
> 
>     6. Join your own dedicated server via server browser
> 
>     7. Press "start match"
> 
>     8. Countdown timer gets stuck
> 
> 
> **EDIT:**
> 
> If I set server to launch into game, e.g. by setting `+setplaylist ctf +mp_gamemode ctf +map mp_glitch` in dedi startup args and then try to join from client, I then error out on client with:
> 
> snapshot0015

Are the server and client running the same dll (the Northstar.dll obtained from the CI artifacts)? Or is the client running a different (from the main branch with no RCON) dll?

GeckoEidechse:
> Are the server and client running the same dll (the Northstar.dll obtained from the CI artifacts)? Or is the client running a different (from the main branch with no RCON) dll?

Both are running the same dll. I just ran client and server from the same Northstar/Titanfall2 install ^^

BulletMemes:
Had this bug while testing on a live server of mine, mind you this is running using pg's docker container, could be the culprit but not sure. 
https://youtu.be/UMVZ9qDZlWY

BulletMemes:
I've also noticed a few seperate things,

1. You can't initiate the RCON request without having a value in the rcon_password ConVar.
2. Running `rcon PASS PasswordHere` will not read the password from the arg if there is a different value in the rcon_password ConVar on the Client end.

Mauler125:
Do any of you have '-netconport' and/or '-netconpassword' set?

Also, could anyone try running the server on windows and run 'ns_server_password' and 'rcon_password' simultaneously? i can't auth to the master server for some reason, even after a full reinstall of the game and Northstar, and keep experiencing the timescale bug.
Also, just to make sure, is there any place in Northstar that opens a TCP socket using the same port as the UDP port of the game protocol?

None of these issues happen on r5reloaded, and the RCON implementation is identical, which is why this bug is interesting.

Mauler125:
> I've also noticed a few seperate things,
> 
>     1. You can't initiate the RCON request without having a value in the rcon_password ConVar.
> 
>     2. Running `rcon PASS PasswordHere` will not read the password from the arg if there is a different value in the rcon_password ConVar on the Client end.

Thanks for your feedback, i added the option to use the third argument from the command as rcon password. See latest commit.

BulletMemes:
> > I've also noticed a few seperate things,
> > ```
> > 1. You can't initiate the RCON request without having a value in the rcon_password ConVar.
> > 
> > 2. Running `rcon PASS PasswordHere` will not read the password from the arg if there is a different value in the rcon_password ConVar on the Client end.
> > ```
> 
> Thanks for your feedback, i added the option to use the third argument from the command as rcon password. See latest commit.

Oh, I thought that was built in from the start from looking at your guide in the PR 
image

Mauler125:
> > > I've also noticed a few seperate things,
> > > ```
> > > 1. You can't initiate the RCON request without having a value in the rcon_password ConVar.
> > > 
> > > 2. Running `rcon PASS PasswordHere` will not read the password from the arg if there is a different value in the rcon_password ConVar on the Client end.
> > > ```
> > 
> > 
> > Thanks for your feedback, i added the option to use the third argument from the command as rcon password. See latest commit.
> 
> Oh, I thought that was built in from the start from looking at your guide in the PR image

It was only implemented on the standalone netconsole client, not the game client.

RoyalBlue1:
image

While compiling this

GeckoEidechse:
Aight, for anyone wanting to test RCON here are some instructions on how to get the necessary files (no compilation etc required)

Before starting, make sure you have a **working and up-to-date Northstar install** and you're **logged into GitHub with your GitHub account** (downloading files from CI only works while logged into any GitHub account)!

In order to use RCON you'll need to download the modified NorthstarLauncher binaries.

1. Head to the bottom of the page of the PR and click on "_Show all checks_"
rcon1

2. For "_CI / build (pull_request)_" click on "_Details_"
rcon2


3. From here click on "_Summary_"
rcon3

4. And then click on "_NorthstarLauncher-XXXXXXX_"
rcon4

5. Once download, open the zip and copy `Northstar.dll` and `NorthstarLauncher.exe` to the Northstar install on your server. \
  `netcon.exe` is a standalone application to connect to your RCON-enabled server and can be used in any client. It does **NOT** have to be copied to your Titanfall2 install folder.
rcon5

:warning: In order to actually enable RCON on the server, [refer to the instructions at the top of the PR](https://github.com/R2Northstar/NorthstarLauncher/pull/100#issue-1159133773). For example RCON will **NOT** be enabled unless you set a password that fulfils the necessary requirements mentioned in the instructions.

GeckoEidechse:
> Auto executes 'rcon_server.cfg' on launch. You can put your configs in here.

What's the location of that file root Titanfall2 folder?

GeckoEidechse:
So on a new attempt I using the newest CI build at the time I now get

GeckoEidechse:
So on a new attempt I using the newest CI build at the time I now get:

image

Checking dedi console a (long enough) password is set:

image

Trying to connect on in-game client it says:
image

And, yes I verified I properly added the files from CI as logs say version is `dev` (I was running stock `v1.9.1` launcher before).

cpdt:
I have a few questions about the RCON protocol after writing my own client (https://github.com/cpdt/northstar-rcon-client)

 - What are `SERVERDATA_REQUEST_VALUE`, `SERVERDATA_REQUEST_SEND_REMOTEBUG`, `SERVERDATA_RESPONSE_VALUE`, `SERVERDATA_RESPONSE_UPDATE`, `SERVERDATA_RESPONSE_STRING` and `SERVERDATA_RESPONSE_REMOTEBUG` for? Seems like none of them are used.
 - Why is there both `SERVERDATA_REQUEST_SETVALUE` and `SERVERDATA_REQUEST_EXECCOMMAND`? The server handler for them is the same.
 - Why do requests and responses have a requestID that is always set to -1 and ignored on the other end?
 - A bit more nit-picky, but since we're using protobuf can't we have separate message types for each response and request type, with the fields they each actually need? That would be less error-prone and make the protocol far more self documenting.

NetConsole/netconsole.cpp
```diff
@@ -0,0 +1,420 @@
...
+				svFirstArg = svFirstArg.erase(svFirstArg.find(" "));
+
+				if (strcmp(svFirstArg.c_str(), "PASS") == 0) // Auth with RCON server.
+				{
+					std::string svSerialized = this->Serialize(svSecondArg, "", cl_rcon::request_t::SERVERDATA_REQUEST_AUTH);
+					this->Send(svSerialized);
+				}
+				else if (strcmp(svFirstArg.c_str(), "SET") == 0) // Set value query.
+				{
+					std::string svSerialized = this->Serialize(svFirstArg, svSecondArg, cl_rcon::request_t::SERVERDATA_REQUEST_SETVALUE);
```
cpdt, line 118: I think there's a bug here, `requestBuf` will always be "SET" so the server will try to set the `SET` convar. Probably need to split the second arg into the var name and value?

NorthstarDedicatedTest/sv_rcon.cpp
```diff
@@ -0,0 +1,478 @@
...
+	ConVar* pConVar = g_pCVar->FindVar(cl_request.requestbuf().c_str());
+	if (pConVar)
+	{
+		pConVar->SetValue(cl_request.requestval().c_str());
+	}
+	else // Execute command with "<val>".
+	{
+		Cbuf_AddText(Cbuf_GetCurrentPlayer(), cl_request.requestbuf().c_str(), cmd_source_t::kCommandSrcCode);
+		Cbuf_Execute();
+	}
```
cpdt, line 404: Since both `SETVALUE` and `EXECCOMMAND` can go here, this could be dangerous:

If I hypothetically send a `SETVALUE` request with `requestBuf="quit"` and `requestVal="potatos"`, there is likely no convar called `quit` so instead this will execute that as a command. 

I don't believe it should ever be possible to execute a command with a `SETVALUE` request.

cpdt (changes requested):

NorthstarDedicatedTest/include/protobuf/any.cc
```diff
@@ -0,0 +1,83 @@
+// Protocol Buffers - Google's data interchange format
```
GeckoEidechse, line 1: From @pg9182 on Discord:

> I recommend adding a `.gitattributes` to mark [protobuf source files] as vendored

NetConsole/netconsole.cpp
Mauler125: Thanks, probably a small mistake. I will take a look at this tomorrow.

NorthstarDedicatedTest/sv_rcon.cpp
Mauler125: There is no danger or undefined behavior here. If you run 'quit "potatos"' or 'exit 255 255 0 255', the engine would still call the 'quit' or 'exit' command callback which is just setting the statemachine's next state to HS_SHUTDOWN. The engine internally checks through the ConCommandBase class whether or not this is a ConVar through virtual method 'ConCommandBase::IsCommand()' (ConVar's and ConCommand's inherit this class for this reason, since they share a good part of the same system), if its a ConCommand it will run it as a such, even if you feed it 500 extra params and values since pCommandBase->IsCommand() returned true.

cpdt: Ah sorry, I didn’t mean to imply there’s undefined behaviour. What I mean is:

If I send a `SETVALUE` request with the variable set to `quit`, I would expect this to do nothing if that convar doesn’t exist - since I’m sending a `SETVALUE` and not an `EXECCOMMAND` request. Instead, as written it will fall back to running the variable as a command when it doesn’t exist, which seems counterintuitive. Hypothetical of course and it would be weird if someone tried to set the value of `quit`, but this could work for any other command too.

It’s not dangerous as in undefined behaviour, but as in it will run a command when you might not expect it to.

Would it make sense to handle SETVALUE and EXECCOMMAND separately?

GeckoEidechse:
> So on a new attempt I using the newest CI build at the time I now get:
> 
> image
> 
> Checking dedi console a (long enough) password is set:
> 
> image
> 
> Trying to connect on in-game client it says: image
> 
> And, yes I verified I properly added the files from CI as logs say version is `dev` (I was running stock `v1.9.1` launcher before).

Okay for some reason I cannot reproduce that issue anymore so I guess consider it solved ig :P

NetConsole/netconsole.cpp
Mauler125: This bug has been fixed as of commit 566d3dc33b9c8be39efe429f8ca848f3ff7580ea.

NorthstarDedicatedTest/sv_rcon.cpp
Mauler125: Yes, this makes sense. I changed the behavior accordingly.
The SETVALUE request will now only work on ConVar's as of commit 566d3dc33b9c8be39efe429f8ca848f3ff7580ea.

Mauler125:
> > So on a new attempt I using the newest CI build at the time I now get:
> > image
> > Checking dedi console a (long enough) password is set:
> > image
> > Trying to connect on in-game client it says: image
> > And, yes I verified I properly added the files from CI as logs say version is `dev` (I was running stock `v1.9.1` launcher before).
> 
> Okay for some reason I cannot reproduce that issue anymore so I guess consider it solved ig :P

This most likely happened because the hostport of the dedi isn't 37015. By default (when no args are passed to netconsole or the rcon command in-game), the netconsole will attempt to connect to localhost::37015. If your server happens to have port 37016 or anything non 37015, which happens if you have multiple servers running, or started a game client before the dedi, or could also happen if something else uses that socket, then you need to specify the correct ip and port manually.

NorthstarDedicatedTest/sha256.cpp
```diff
@@ -0,0 +1,129 @@
...
+	SHA256 ctx = SHA256();
+	ctx.init();
+	ctx.update(const_cast<unsigned char*>(reinterpret_cast<const unsigned char*>(input.c_str())), input.length());
+	ctx.final(digest);
+
+	char buf[2 * SHA256::DIGEST_SIZE + 1] {};
+	buf[2 * SHA256::DIGEST_SIZE] = 0;
+	for (int i = 0; i < SHA256::DIGEST_SIZE; i++)
+	{
+		sprintf_s(buf + i * 2, sizeof(buf), "%02x", digest[i]);
```
cpdt, line 126: > Exception Thrown
> Run-Time Check Failure #<span></span>2 - Stack around the variable 'buf' was corrupted.

Happens in the call to `CRConServer::SetPassword`.

This fixes it, although it's a bit weird that it's happening in the first place:

```suggestion
		sprintf_s(buf + i * 2, sizeof(buf) - i * 2, "%02x", digest[i]);
```

NorthstarDedicatedTest/sv_rcon.cpp
```diff
@@ -0,0 +1,569 @@
...
+				this->ProcessMessage(
+					this->Deserialize(std::string(reinterpret_cast<char*>(pData->m_RecvBuffer.data()), pData->m_nPayloadLen)));
+
+				pData->m_nPayloadLen = 0;
+				pData->m_nPayloadRead = 0;
+			}
+		}
+		else if (pData->m_nPayloadRead < sizeof(int)) // Read size field.
+		{
+			pData->m_RecvBuffer[pData->m_nPayloadRead++] = *pRecvBuf;
```
cpdt, line 383: > Expression: vector subscript out of range

Because `m_RecvBuffer` has 4 bytes reserved, but it's length is still 0. I was able to just change the `CConnectedNetConsoleData` constructor to initialize the vector to 4 elements instead of reserving space, but I'm not sure if that's the best way with how you're using the buffer.

cpdt (changes requested):
I ran into a couple of assertion failures when running a debug build of this:

DxsSucuk:
What exactly happends on a wrong password try?
Cause while playing around (I was using https://github.com/cpdt/northstar-rcon-client/blob/main/northstar-rcon-cli/src/main.rs) I found out that the Server does not tell me that I entered the wrong password and just continues keeping the connection normally open.

The external rcon client that I used has a check for authentication Issues unsure if that one is wrong or if Northstar even sends a rcon auth error?

cpdt:
@DxsSucuk If you try the included netcon.exe client, it works correctly:

```
Northstar TCP net console [Version 2.0.0.1]
Enter <IP> <PORT>:
[22:06:26] [info] Connected to: [127.0.0.1]:37015
sv_cheats 1
[22:06:29] [info] This server is password protected for console access. Authenticate with 'PASS <password>' command.
PASS test
[22:06:31] [info] Admin password incorrect.
```

So that's a bug in my RCON client, presumably the auth response strings that it was checking against have changed. I've tracked that in https://github.com/cpdt/northstar-rcon-client/issues/1.

DxsSucuk:
Thanks for the response @cpdt!

DxsSucuk:
I noticed that the server does not log that someone connected over RCON?
Is that wanted or?

NorthstarDedicatedTest/cl_rcon.cpp
```diff
@@ -0,0 +1,365 @@
...
+
+#include "pch.h"
+#include "convar.h"
+#include "concommand.h"
+#include "cvar.h"
+#include "rcon_shared.h"
+#include "sv_rcon.pb.h"
+#include "cl_rcon.pb.h"
+#include "cl_rcon.h"
+#include "net.h"
```
GeckoEidechse, line 15: For some reason, this include causes the CI build to fail.

Removing the include (and the one function call that needs it) causes the CI to build just fine (excluding formatting errors).

See also 1b6e0aa3d8f2774a68ef361141a0c4b7e411f3b2

1b6e0aa3d8f2774a68ef361141a0c4b7e411f3b2

NorthstarDedicatedTest/sha256.cpp
Mauler125: Thanks for the suggestions. The reason i used sprintf_s here was to suppress compiler warnings. On the r5sdk repo i reverted this change a long time back, but somehow missed porting this over to Northstar. It has been fixed as of commit 24f561f8bb656bd7a242d77c40651318e89ad2f8.

NorthstarDedicatedTest/sv_rcon.cpp
Mauler125: Thanks, small oversight. Fixed as of commit 24f561f8bb656bd7a242d77c40651318e89ad2f8.

NorthstarDedicatedTest/cl_rcon.cpp
Mauler125: Fixed as of 4573d5e3141915f7066f182b47dac6a1d7d7bdcb.
This was caused by commit 938d89713eb055340b5e8866de3000b61e4a0f13 from the Squirrel PR as found by Gecko.

GeckoEidechse (changes requested):
Before someone pings me again on Discord about RCON, the reason it's not getting merged currently is due to being flat-out broken in the sense that it causes host timescale issues.

More info can be found in this comment I left a while back that GitHub "helpfully" hides due to the amount of comments left on this PR: https://github.com/R2Northstar/NorthstarLauncher/pull/100#pullrequestreview-1027925594 
(Just retested on latest RCON commit and it still exists)

Obviously as long as only having RCON active (no RCON clients connected) already causes timescale issues making the game unplayable there's no point in merging until fixed.

The other issue is merge conflicts but those can always be resolved once we have a working RCON implementation. ^^

Mauler125:
Pull request has been closed due to lack of interest. The amount of work required for merging the local fixes for the time scale issue into the new code base outweighs the interest for RCON. The contributors of Northstar had also requested me to move this into a plug-in instead; this is a big task which I unfortunately no longer have time for.

GeckoEidechse:
> Pull request has been closed due to lack of interest. The amount of work required for merging the local fixes for the time scale issue into the new code base outweighs the interest for RCON.

Do note that interest for RCON support (or any protocol that allows for remotely interacting with server) is still very high. It wasn't tackled by core Northstar contributors at the time, due to higher priority tasks like codebase refactors and adding needed functionality to release new gamemodes, or lower hanging fruits that are easier to implement. There was also the assumption that this pull request was still being worked on to resolve the issues it had.

You're of course free to share your unpushed local fixes so that we can hopefully resolve the timescale issues together and push this PR over the finish line. ^^

&nbsp;

As for

> The contributors of Northstar had also requested me to move this into a plug-in instead; this is a big task which I unfortunately no longer have time for.

Despite the discussion, I'm not sure there was ever a definitive conclusion whether to do plugin or merge directly. The main issue was lack of a working implementation that didn't cause clients to desync due to server timescale slowing down as soon as RCON is active. An actually working version would have probably guided the way of how to implement.

Issue #101, opened 2022-03-04
Client chat hooks memory leak


cpdt:
In the `CHudChat::AddGameLine` hook:

https://github.com/R2Northstar/NorthstarLauncher/blob/13dcc7d9307b160131ed80832efda99001aa6c6d/NorthstarDedicatedTest/clientchathooks.cpp#L18-L28

If `squirrel->setupfunc` does not error but `IsFirstHud` returns false, `squirrel->call` will never be called. The function that `squirrel->setupfunc` pushes to the stack will never be popped, leaking memory.

This will happen roughly four times for each chat message, since `CHudChat::AddGameLine` is called for every chat hud and `IsFirstHud` returns false for all but one of them.

PR #103, opened 2022-03-05
Fix a memory leak for `setupfunc`

emma-miler:


emma-miler:
@cpdt and I talked about this. Does not fix #101, but fixes a related issue

cpdt (approved):

PR #106, opened 2022-03-05
DebugOverlay

r-ex:
DrawOverlay hook and reimplementation using existing render functions to recreate debugoverlay functionality for OVERLAY_LINE and OVERLAY_BOX


BobTheBob9:
is this behind a convar or smth or will it just draw these all the time?

abarichello:
I believe this is used by `DebugDrawBox` and `DebugDrawLine`, there's some dev functions that use it but they don't draw anything. Ex: [DebugTitanfall](https://github.com/R2Northstar/NorthstarMods/blob/846cd1efbb9e8c8de0f0c50ae90d271281c45719/Northstar.CustomServers/mod/scripts/vscripts/titan/_replacement_titans_drop.gnut#L49).

BobTheBob9:
yes it is, what i was asking was whether there was any way to toggle that functionality through a cvar or smth, since we probably don't want `DebugDrawBox`, `DebugDrawLine`, etc to render anything for non-developer users

r-ex:
iirc, DrawAllOverlays (which calls DrawOverlay) has "enable_debug_overlays" checked before running anything related to drawing, although it's set to 1 by default

abarichello:
> since we probably don't want `DebugDrawBox`, `DebugDrawLine`, etc to render anything for non-developer users

From what i have seen in vanilla scripts they always have a debug variable check before calling it

emma-miler:
> > since we probably don't want `DebugDrawBox`, `DebugDrawLine`, etc to render anything for non-developer users
> 
> From what i have seen in vanilla scripts they always have a debug variable check before calling it

This is true, but setting `DEV` breaks some scripts, so it would be better to have it as a convar

r-ex:
would it be bad to just set the value of `enable_debug_overlays` to 0 instead of the current default of 1 at launch?

that way you don't need to rely on vars being checked in scripts (as mod creators that use debug overlay to test would have to remember to put a var in to stop it from rendering at all times)

(DrawAllOverlays is at engine.dll+0xAB780. enable_debug_overlays is checked at the start)
image

emma-miler:
If that cvar works, then that looks good to me.

abarichello:
Could you fix the conflicts on this one, please?

PR #112, opened 2022-03-23
Implement KittenPopo exploit fixes (and sanity changes)

KittenPopo:


NorthstarDedicatedTest/sigscanning.cpp
```diff
@@ -5,7 +5,7 @@
 // note: sigscanning is only really intended to be used for resolving stuff like shared function definitions
 // we mostly use raw function addresses for stuff
 
-size_t GetDLLLength(HMODULE moduleHandle)
+size_t GetDLLength(HMODULE moduleHandle)
```
barnabwhy, line 8: Should actually be DLLLength, 3 Ls cause 2 in DLL and 1 in Length

README.md
```diff
@@ -1,12 +1,5 @@
-# NorthstarLauncher
-![Build Status](https://github.com/R2Northstar/NorthstarLauncher/actions/workflows/ci.yml/badge.svg)
+# NorthstarLauncher - ExploitFixes
```
barnabwhy, line 1: Don’t change readme

barnabwhy (changes requested):

barnabwhy:
other then those two things i don’t see anything wrong with this, haven’t done any testing though

NorthstarDedicatedTest/sigscanning.cpp
KittenPopo: Fixed, oops

KittenPopo:
clang format be like "nice indentation, it would be a shame if i were to ruin it"

GeckoEidechse:
> I forgot to exclude readme.md from pull request, not sure if I can change that .

Just change it back to original in another commit and we'll probably squash merge the whole thing anyway which should remove any indication the change ever happened in the first place ^^

KittenPopo:
> Just change it back to original in another commit and we'll probably squash merge the whole thing anyway which should remove any indication the change ever happened in the first place ^^

Ok fair enough, done

PR #113, opened 2022-03-24
Add version numbering to all master server requests

emma-miler:
Adds version numbers to all MS requests in the form of a custom user agent.
The version number is retrieved from the resources' file version.
@abarichello i changed the default version from `0.0.0.0` to `0.0.0.1` to make builds dev versions, so this needs to be changed in CI


emma-miler:
@pg9182 feel free to scream at me for violating the core fundamentals of WinApi or whatever

NorthstarDedicatedTest/version.cpp
```diff
@@ -0,0 +1,59 @@
...
+				DWORD dwLeftMost = HIWORD(dwFileVersionMS);
+				DWORD dwSecondLeft = LOWORD(dwFileVersionMS);
+				DWORD dwSecondRight = HIWORD(dwFileVersionLS);
+				DWORD dwRightMost = LOWORD(dwFileVersionLS);
+
+				// We actually use the rightmost integer do determine whether or not we're a debug/dev build
+				// If it is set to 1 (as in resources.rc), we are a dev build
+				// On github CI, we set this 1 to a 0 automatically as we replace the 0.0.0.1 with the real version number
+				if (dwRightMost == 1) {
+					sprintf(version, "%d.%d.%d.%d-dev", dwLeftMost, dwSecondLeft, dwSecondRight, dwRightMost);
```
pg9182, line 39: It needs to be `+dev` for it to sort correctly with semver.

NorthstarDedicatedTest/version.cpp
```diff
@@ -0,0 +1,59 @@
...
+	hResInfo = FindResource(hInst, MAKEINTRESOURCE(1), RT_VERSION);
+	if (hResInfo != NULL) {
+		dwSize = SizeofResource(hInst, hResInfo);
+		hResData = LoadResource(hInst, hResInfo);
+		if (hResData != NULL) {
+			pRes = LockResource(hResData);
+			pResCopy = LocalAlloc(LMEM_FIXED, dwSize);
+			if (pResCopy != 0) {
+				CopyMemory(pResCopy, pRes, dwSize);
+				VerQueryValue(pResCopy, L"\\", (LPVOID*)&lpFfi, &uLen);
```
pg9182, line 25: We're being explicit in the rest of northstar, so do FindResourceW and VerQueryValueW.

NorthstarDedicatedTest/masterserver.cpp
```diff
@@ -1109,14 +1117,15 @@ void MasterServerManager::WritePlayerPersistentData(char* playerId, char* pdata,
 		{
 			CURL* curl = curl_easy_init();
 			SetCommonHttpClientOptions(curl);
+			curl_easy_setopt(curl, CURLOPT_USERAGENT, &NSUserAgent);
```
pg9182, line 1120: Why is this not being done in `SetCommonHttpClientOptions`?

GeckoEidechse:
In the long run we should switch from using `File version` to `Product version` as the latter allows for more custom strings like including `-rc1` endings for example.

NorthstarDedicatedTest/version.cpp
GeckoEidechse: Done in 2b2cf439a16356daf3253d299143b279b33f4984

NorthstarDedicatedTest/masterserver.cpp
GeckoEidechse: Resolved in 3643deb64efe15ab38af27bd605b03d587cf1d5b

NorthstarDedicatedTest/version.cpp
GeckoEidechse: Resolved in 3643deb64efe15ab38af27bd605b03d587cf1d5b

NorthstarDedicatedTest/masterserver.cpp
GeckoEidechse: Resolved in 34da460021e723ef09e45fecd9e65018524fa1d7

NorthstarDedicatedTest/version.cpp
```diff
@@ -0,0 +1,59 @@
...
+				DWORD dwLeftMost = HIWORD(dwFileVersionMS);
+				DWORD dwSecondLeft = LOWORD(dwFileVersionMS);
+				DWORD dwSecondRight = HIWORD(dwFileVersionLS);
+				DWORD dwRightMost = LOWORD(dwFileVersionLS);
+
+				// We actually use the rightmost integer do determine whether or not we're a debug/dev build
+				// If it is set to 1 (as in resources.rc), we are a dev build
+				// On github CI, we set this 1 to a 0 automatically as we replace the 0.0.0.1 with the real version number
+				if (dwRightMost == 1) {
+					sprintf(version, "%d.%d.%d.%d+dev", dwLeftMost, dwSecondLeft, dwSecondRight, dwRightMost);
```
pg9182, line 39: It might be a good idea to replace this with a snprintf + proper error checking in the future, but it won't overflow with any reasonable version (and we control that), so it's probably ok to leave it for now.

pg9182 (approved):
Good enough.

pg9182:
As discovered during testing, we'll need to build the version number into the DLL so we can still extract it even when launched via vanilla titanfall 2 and the launcher proxy or wsock32 stub (i.e., steam).

PR #131, opened 2022-03-29
ExploitFixes update + misc.

KittenPopo:
**ExploitFixes:**
- Disabled flag checks for NET_SetConVar (will improve these in the future)
- CUserCmds are no longer reset if an angle is invalid (this can happen in vanilla)

**Misc:**
- Set default value for _net_chan_limit_msec_per_sec_ (prevents console spam because the default was 0, which means all netmsgs are flagged as over the limit)


GeckoEidechse:
Whoever merges this, please make sure that the commit message matches up with the actual commits, starting at https://github.com/R2Northstar/NorthstarLauncher/pull/131/commits/96747bcefd46cf3a061d57568d810125bb9234d1

PR #133, opened 2022-04-05
add audio changes

BobTheBob9:
functionally this works from what i've tested and removes alot of the hitching on startup from audio loads, however there's probably a bunch of ways it could be improved before merging


pg9182:
Why not a [`shared_mutex`](https://en.cppreference.com/w/cpp/thread/shared_mutex), which is designed exactly for this purpose?

BobTheBob9:
any other thoughts? if not i'll format and merge

pg9182:
Looks good, but I haven't tested this (and don't currently have time to).

PR #134, opened 2022-04-05
add rui_drawEnable

BobTheBob9:
allows rui rendering to be enabled/disabled with the cvar rui_drawEnable (this is something respawn can do on dev builds, with a cvar of the same name)
works fine in testing, haven't encountered any issues with it


BobTheBob9:
i hate github.com why is this not building it builds fine locally

BobTheBob9:
ok i was actually just braindead, should be good now

PR #137, opened 2022-04-09
allow mods to load biks that aren't already present in r2/media

BobTheBob9:
atm, modded videos will only be loaded if they are overriding a video that's present in r2/media already, nopping out a check in the engine video loading function forces the game to attempt to load it regardless of whether it's there, so it'll attempt to load modded biks

in testing that hasn't caused any issues with either vanilla, overriden or custom biks


EladNLG (approved):
Looks good to me ¯\_(ツ)_/¯

uniboi (approved):
Forgotten PR 💀

CaptainDiqhedd:
This should be like, merged, n stuff, it would be really epic n things. i would luv u forever. thank you for taking the time to read my comment. :))))

BobTheBob9:
I fr did just forget this

PR #139, opened 2022-04-11
more changes to mod rpak loading

BobTheBob9:
adds support for unloading mod rpaks, fixes a bug with loading entirely custom mod rpaks, and prevents all unnecessary rpaks from being loaded on dedicated servers


pg9182:
Looks fine, but it's not like we got a chance to look at it, lol.

PR #140, opened 2022-04-11
make toggleconsole FCVAR_DONTRECORD

BobTheBob9:


GeckoEidechse:
Prevents toggling console from being recorded in demos.

PR #147, opened 2022-04-14
Add run_on requirements for server and client

emma-miler:
Plugins must now specify whether or not they want to run on client and/or server.
DiscordRPC has already been updated

### DIFF
--- a/NorthstarDedicatedTest/dllmain.cpp
+++ b/NorthstarDedicatedTest/dllmain.cpp
@@ -158,13 +158,34 @@ bool LoadPlugins()
 		{
 			spdlog::error("{} does not have a version number in its manifest", pathstring);
 			continue;
-			// spdlog::info(manifestJSON["version"].GetString());
+		}
+		if (!manifestJSON.HasMember("run_on_client") || !manifestJSON.HasMember("run_on_server"))
+		{
+			spdlog::error("{} does not specify run_on_client and run_on_server. It will not be loaded!", pathstring);
+			continue;
 		}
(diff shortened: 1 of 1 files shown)


NorthstarDedicatedTest/dllmain.cpp
```diff
@@ -158,13 +158,34 @@ bool LoadPlugins()
 		{
 			spdlog::error("{} does not have a version number in its manifest", pathstring);
 			continue;
-			// spdlog::info(manifestJSON["version"].GetString());
+		}
+		if (!manifestJSON.HasMember("run_on_client") || !manifestJSON.HasMember("run_on_server"))
```
GeckoEidechse, line 162: Shouldn't it be `&&` instead of `||`

Cause currently a plugin that only specifies client (like DiscordRPC) will hit true on the second condition and enter the `if` clause

GeckoEidechse (changes requested):

NorthstarDedicatedTest/dllmain.cpp
emma-miler: No, this is checking if either of them *doesnt* exist

GeckoEidechse: Ah so it makes sure that keys are set regardless of content (`true`/`false`)? In that case I misread and it should be fine as is ^^

GeckoEidechse (approved):
Couldn't test it locally though cause out-of-town

GeckoEidechse:
Closes https://github.com/R2Northstar/Northstar/issues/256

emma-miler:
Superceded by new plugin API coming soon:tm:

PR #148, opened 2022-04-20
Implementing pdiff (mod persistence)

ASpoonPlaysGames:
Must be implemented alongside [NorthstarMasterServer/59](https://github.com/R2Northstar/NorthstarMasterServer/pull/59)

EDIT: just in case there is any confusion, this doesn't seem to require [NorthstarMasterServer/59](https://github.com/R2Northstar/NorthstarMasterServer/pull/59) and works fine with current MS build, however [NorthstarMasterServer/59](https://github.com/R2Northstar/NorthstarMasterServer/pull/59) does require this PR


ASpoonPlaysGames:
This is now ready for review, shouldn't break backwards compatibility (i've been using it myself just normally)

PR #150, opened 2022-04-22
Fix debug builds

emma-miler:
Properly initializes AuthData structs where used.
This does not get done automatically in debug mode because c++ is weird like that


emma-miler:
Fails clang-format due to debugoverlay not conforming. Fixed in #151

PR #152, opened 2022-04-23
Hardcode version number

### LINKED ISSUE #129: User-agent version incorrect when launching with alternative mechanisms
As discovered during testing, we'll need to build the version number into the DLL so we can still extract it even when launched via vanilla titanfall 2 and the launcher proxy or wsock32 stub (i.e., steam).

#113

GeckoEidechse:
This way, we don't get the wrong version number when launching Titanfall2.exe via `-northstar`
The hardcoded value should be overwritten by CI on release (https://github.com/R2Northstar/Northstar/pull/262)

Fixes #129

### DIFF
--- a/NorthstarDedicatedTest/version.cpp
+++ b/NorthstarDedicatedTest/version.cpp
@@ -4,62 +4,27 @@
 char version[16];
 char NSUserAgent[32];
 
+// Turning off clang-format here so it doesn't mess with style as it needs to be this way for CI
+// clang-format off
+#define NORTHSTAR_VERSION 0,0,0,1
+// clang-format on
+
 void InitialiseVersion()
 {
-	HRSRC hResInfo;
-	DWORD dwSize;
-	HGLOBAL hResData;
-	LPVOID pRes, pResCopy;
-	UINT uLen = 0;
-	VS_FIXEDFILEINFO* lpFfi = NULL;
-	HINSTANCE hInst = ::GetModuleHandle(NULL);
+	int northstar_version[] = {NORTHSTAR_VERSION};
 
-	hResInfo = FindResourceW(hInst, MAKEINTRESOURCE(1), RT_VERSION);
-	if (hResInfo != NULL)
+	// We actually use the rightmost integer do determine whether or not we're a debug/dev build
+	// If it is set to 1, we are a dev build
+	// On github CI, we set this 1 to a 0 automatically as we replace the 0,0,0,1 with the real version number
+	if (northstar_version[3] == 1)
 	{
-		dwSize = SizeofResource(hInst, hResInfo);
-		hResData = LoadResource(hInst, hResInfo);
-		if (hResData != NULL)
-		{
-			pRes = LockResource(hResData);
-			pResCopy = LocalAlloc(LMEM_FIXED, dwSize);
-			if (pResCopy != 0)
-			{
-				CopyMemory(pResCopy, pRes, dwSize);
-				VerQueryValueW(pResCopy, L"\\", (LPVOID*)&lpFfi, &uLen);
-
-				DWORD dwFileVersionMS = lpFfi->dwFileVersionMS;
-				DWORD dwFileVersionLS = lpFfi->dwFileVersionLS;
-
-				DWORD dwLeftMost = HIWORD(dwFileVersionMS);
-				DWORD dwSecondLeft = LOWORD(dwFileVersionMS);
-				DWORD dwSecondRight = HIWORD(dwFileVersionLS);
-				DWORD dwRightMost = LOWORD(dwFileVersionLS);
-
-				// We actually use the rightmost integer do determine whether or not we're a debug/dev build
-				// If it is set to 1 (as in resources.rc), we are a dev build
-				// On github CI, we set this 1 to a 0 automatically as we replace the 0.0.0.1 with the real version number
-				if (dwRightMost == 1)
-				{
-					sprintf(version, "%d.%d.%d.%d+dev", dwLeftMost, dwSecondLeft, dwSecondRight, dwRightMost);
-					sprintf(NSUserAgent, "R2Northstar/%d.%d.%d+dev", dwLeftMost, dwSecondLeft, dwSecondRight);
-				}
-				else
-				{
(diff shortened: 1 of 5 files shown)


GeckoEidechse:
Requires https://github.com/R2Northstar/Northstar/pull/262

GeckoEidechse:
So to give some background, I just opted to rip out the entire version resource reading part as to my knowledge we won't be able to fix that to read the "correct" version when running Titanfall2 with `-northstar`.

As such I also opted to make the number defined in `version.cpp` to be the more definitive one. The one defined in `resources.rc` is now basically just used to give the version of the Windows executable.

As per https://github.com/R2Northstar/Northstar/pull/262 the version number will be updated before compile inside CI when doing a release build.

It's not the nicest to have to define two version numbers in the code but it works ¯\\\_(ツ)\_/¯

pg9182:
> It's not the nicest to have to define two version numbers in the code but it works ¯\_(ツ)_/¯

You know you can make another ns_version.h header like this:

```cpp
#pragma once
#ifndef NORTHSTAR_VERSION
#define NORTHSTAR_VERSION 0,0,0,1
#endif
```

Then, include that in both this file and the rc. During the CI, you can use the a flag like `/DNORTHSTAR_VERSION=1,2,3,0` with msbuild, and you won't need to do any replacements manually.

Also, regarding the int array you create with the version, why not just include NORTHSTAR_VERSION in the fn args directly? It's less error-prone and more readable.

pg9182 (commented):
^

GeckoEidechse:
> You know you can make another ns_version.h header like this:

I did not but it seems more sensible so I'll look into it.

> Also, regarding the int array you create with the version, why not just include NORTHSTAR_VERSION in the fn args directly? It's less error-prone and more readable.

I didn't want to change the signature to avoid changing too much code at once.

NorthstarDedicatedTest/version.cpp
```diff
@@ -4,62 +4,27 @@
...
 void InitialiseVersion()
 {
-	HRSRC hResInfo;
-	DWORD dwSize;
-	HGLOBAL hResData;
-	LPVOID pRes, pResCopy;
-	UINT uLen = 0;
-	VS_FIXEDFILEINFO* lpFfi = NULL;
-	HINSTANCE hInst = ::GetModuleHandle(NULL);
+	int northstar_version[] = {NORTHSTAR_VERSION};
```
emma-miler, line 14: This could be turned into a `constexpr` which saves a bit of time on startup

NorthstarDedicatedTest/version.cpp
```diff
@@ -4,62 +4,27 @@
...
-			FreeResource(hResData);
-			LocalFree(pResCopy);
-		}
+		sprintf(version, "%d.%d.%d.%d+dev", northstar_version[0], northstar_version[1], northstar_version[2], northstar_version[3]);
+		sprintf(NSUserAgent, "R2Northstar/%d.%d.%d+dev", northstar_version[0], northstar_version[1], northstar_version[2]);
+	}
+	else
+	{
+		sprintf(version, "%d.%d.%d.%d", northstar_version[0], northstar_version[1], northstar_version[2], northstar_version[3]);
+		sprintf(NSUserAgent, "R2Northstar/%d.%d.%d", northstar_version[0], northstar_version[1], northstar_version[2]);
```
emma-miler, line 27: Bit of code duplication here, but that's fine if you want to keep it that way. Personally, i'd say turn NSUserAgent into an `std::string` and then initialize it with a format string and append the `+dev` part after, but thats up to you

emma-miler (changes requested):

NorthstarDedicatedTest/version.cpp
GeckoEidechse: So
```C
constexpr int northstar_version[] = {NORTHSTAR_VERSION};
```
?

emma-miler: yes, but you would put that in global scope, outside of the function. I'm not exactly sure on the semantics of `constexpr` mixed with `#define`, but i *think* that should be fine. The easiest way to check is to simply inspect the value of `northstar_version` in the binary, since it would be placed somewhere in the `.data` section

emma-miler: Okay so it turns out the semantics for `constexpr` are slightly different to normal construction. This should work however:
`constexpr int ns_version[4]{NORTHSTAR_VERSION};`

emma-miler: The above can be verified by running a shimmed version through gdb and running `print (int[4])ns_version` before ever starting execution.

GeckoEidechse:
> You know you can make another ns_version.h header like this:
> 
> ```c++
> #pragma once
> #ifndef NORTHSTAR_VERSION
> #define NORTHSTAR_VERSION 0,0,0,1
> #endif
> ```

Aight, so adding a new header file and using that works.

> Then, include that in both this file and the rc. During the CI, you can use the a flag like `/DNORTHSTAR_VERSION=1,2,3,0` with msbuild, and you won't need to do any replacements manually.

So the flag `/DNORTHSTAR_VERSION=1,2,3,0` does not work with `msbuild`. However I found that

```
set CL=/DNORTHSTAR_VERSION#1,2,3,0
msbuild /t:Rebuild /p:Configuration=Debug R2Northstar.sln
```

does except that for some reason the version is now correct in `version.cpp` (i.e. `v1.2.3`) but for some reason this is not the case for `resources.rc` (`v0.0.0+dev`).
I think we'll just stick to regex-ing the value for now ^^

NorthstarDedicatedTest/version.cpp
GeckoEidechse: Done in f3e897146521db5db3900bef9280299b48a5af6a

NorthstarDedicatedTest/version.cpp
GeckoEidechse: Eh, I'm too lazy to figure it out now... ^^

emma-miler (approved):
Looks good to me, the constexpr is properly being calculated at compile-time, and the strings are formatted correctly

PR #153, opened 2022-04-23
Write launcher version to log file

GeckoEidechse:
Should help sorting out old logs

### DIFF
--- a/NorthstarDedicatedTest/dllmain.cpp
+++ b/NorthstarDedicatedTest/dllmain.cpp
@@ -287,5 +287,8 @@ bool InitialiseNorthstar()
 	// run callbacks for any libraries that are already loaded by now
 	CallAllPendingDLLLoadCallbacks();
 
+	// Write launcher version to log
+	spdlog::info("NorthstarLauncher version: {}", version);
+
 	return true;
 }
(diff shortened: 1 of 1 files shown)


GeckoEidechse:
To give some background on why this particular location was chosen to add the line, I put it somewhere where it prints the version close to the top of the log file, logging to file in general should be initialised already, and I have access to `spdlog` without needing to additionally add the header.

emma-miler (commented):
As mentioned on discord, i would recommend moving this call to just after logging and version gets initialized, since we are "likely" to crash while injecting if there is any issue with that, which means we still dont get the benefit of seeing the version number in the logs

GeckoEidechse:
> As mentioned on discord, i would recommend moving this call to just after logging and version gets initialized, since we are "likely" to crash while injecting if there is any issue with that, which means we still dont get the benefit of seeing the version number in the logs

I tried that but for some reason logging wasn't fully initialised at that point and I'm unsure why. Could try moving it up further by trial-and-error but putting it right after logging initialisation definitely didn't work for some reason. It would only show up in console but not in logfile :/

emma-miler (approved):

PR #155, opened 2022-04-26
Fix duplicate script in scripts.rson + allow mods to force vpks to load through vpk.json

EladNLG:
See title.

NOTE: the duplicate script in scripts.rson fix doesn't take into account the RunOn of the two scripts.


EladNLG:
Should re-fork the repo, this is way too unorganized

PR #156, opened 2022-04-26
Add dependency constants

EladNLG:
See title.

NOTE: the fix doesn't take into account the script's RunOn field.


NorthstarDedicatedTest/scriptsrson.cpp
```diff
@@ -11,6 +11,7 @@ void ModManager::BuildScriptsRson()
 	spdlog::info("Building custom scripts.rson");
 	fs::path MOD_SCRIPTS_RSON_PATH = fs::path(GetCompiledAssetsPath() / MOD_SCRIPTS_RSON_SUFFIX);
 	fs::remove(MOD_SCRIPTS_RSON_PATH);
+	std::vector<std::string> scriptsAdded;
```
emma-miler, line 14: Could use a better variable name

NorthstarDedicatedTest/scriptsrson.cpp
```diff
@@ -37,15 +38,23 @@ void ModManager::BuildScriptsRson()
 				_coolscript.gnut
 			]*/
 
+			// check if the same script has been added already.
+			if (std::find(scriptsAdded.begin(), scriptsAdded.end(), script.Path) != scriptsAdded.end())
+			{
+				continue;
```
emma-miler, line 44: This would silently drop the script. Consider notifying the user about this.
This could create issues if a mod is relying on modifying a file for the rest of its behaviour, so we need it be logged

emma-miler (changes requested):

NorthstarDedicatedTest/scriptsrson.cpp
EladNLG: Removed this in favor of a dependency constant system.

NorthstarDedicatedTest/scriptsrson.cpp
EladNLG: See resolved conversation below.
(This fix was removed.)

wolf109909:
This intheory shoud fix all issues with CARMOD, cool!

ASpoonPlaysGames:
could this cause issues in respawn code where they do `#if CLIENT #else` or no?

EladNLG:
> could this cause issues in respawn code where they do `#if CLIENT #else` or no?

No. This only defines new constants and does not modify any existing ones.

NorthstarDedicatedTest/modmanager.cpp
```diff
@@ -333,6 +364,27 @@ void ModManager::LoadMods()
...
+						bool shouldSkip = std::find_if(mod.Vpks.begin(), mod.Vpks.end(), matchesV) != mod.Vpks.end();
+						if (shouldSkip)
+							continue;
+						if (!v->value.IsTrue())
+							continue;
+						ModVPKEntry& modVpk = mod.Vpks.emplace_back();
+						modVpk.m_bAutoLoad = true;
+						std::string path = "vpk/client_";
+						path += v->name.GetString(); // for some reason these have to be done separately but fine
+						modVpk.m_sVpkPath = path;
```
BobTheBob9, line 385: Why don't you call MountVPK here like we do for mod vpks?

BobTheBob9 (commented):
dependency stuff is great and very mergable, might be better to move vpk stuff to another pr for now

NorthstarDedicatedTest/modmanager.cpp
EladNLG: Moved to another PR for now

NorthstarDedicatedTest/keyvalues.cpp
```diff
@@ -88,6 +88,63 @@ void ModManager::TryBuildKeyValues(const char* filename)
...
+		{
+			std::ifstream filestream(f.owningMod->ModDirectory / "mod" / normalisedPath);
+			std::stringstream stringstream;
+
+			if (filestream.fail())
+			{
+				spdlog::warn("kv {} was overriden by a mod but no file was found!", ogFilePath);
+				return;
+			}
+			while (filestream.peek() != EOF)
```
EladNLG, line 106: This, should seriously be a utility function. I'm seeing it get used like 4 times over this project.

ASpoonPlaysGames (approved):
unsure about good practice and all that in cpp tbh but logic seems sound from first inspection?

BobTheBob9 (approved):
looks good overall

PR #160, opened 2022-05-01
Fix segfault on badly formed compressed packets

cpdt:
This is a rewrite of CLZSS::SafeUncompress but with an extra guard to block badly formed LZSS data that attempt to read segments from before the start of the output buffer, sometimes causing a segfault. The added check is lines 88-92, everything else should work identical to vanilla.


GeckoEidechse:
Prevents crashes from attacks but lag is obviously still happening.

GeckoEidechse:
(Note to self: Bob approved it in Discord chat)

PR #161, opened 2022-05-02
dont allow multiple of the same account to connect without a commandline arg

BobTheBob9:
probably should've done this earlier, unsure if changes much in practice really but probably just good to do


GeckoEidechse:
Clang-format errors :trollface:

GeckoEidechse:
Actually, it's just cause that branch is based on an older commit. Merging `main` into it should fix clang-format errors ^^

PR #162, opened 2022-05-02
make hidden and devonly convars visible

RoyalBlue1:
also includes find and list command for convars


BobTheBob9:
this is essentially just pull #98 but with the code cleared up a bit, regardless of whether it's a bit more readable i don't think prs that hook string manipulation funcs and such for stuff like this should be what we're merging in at all, there's lots of hacky hook code in the codebase already and we should be trying to get rid of it, not add more

BobTheBob9:
sorry if this is a tad late, only just got to properly looking through it

emma-miler:
We could probably do without most of the string manipulation functions in this PR, but other than that it looks fine to me. Tested working

PR #164, opened 2022-05-03
Feature - hot reload banlist

### LINKED ISSUE #163: Hosting a dedicated server and banning yourself prevents you from opening lobby
### Bug
Not sure if this is a bug at all, but something I discovered while playing around with https://github.com/R2Northstar/Northstar/issues/251.

If the server files are in the same folder as the game files, banning yourself from the server, bans you from the main lobby.

To my understanding, this happens because the [banlist is checked every time](https://github.com/R2Northstar/NorthstarLauncher/blob/6c8112a6c368dd36d21fd94689e9682bc3b012a5/NorthstarDedicatedTest/serverauthentication.cpp#L370) a user connects (and you connect to the lobby) and since you are in that list, you get banned.

### Steps to reproduce
  1. Start dedicated server
  2. ban yourself
  3. Restart the game
  4. Try to join the lobby

### Specifications
  - Northstar version: 1.7.0
  - Platform: Origin

dilirity:
This fixes #163 and hopefully https://github.com/R2Northstar/Northstar/issues/251.

Not sure how to go about listen servers, since I don't know the equivalent for [isDedicated()](https://github.com/R2Northstar/NorthstarLauncher/blob/6c8112a6c368dd36d21fd94689e9682bc3b012a5/NorthstarDedicatedTest/dedicated.cpp#L8).

First time working with C++ so be gentle.

### DIFF
--- a/NorthstarDedicatedTest/serverauthentication.cpp
+++ b/NorthstarDedicatedTest/serverauthentication.cpp
@@ -367,7 +367,7 @@ bool CBaseClient__ConnectHook(void* self, char* name, __int64 netchan_ptr_arg, c
 	if (!ret)
 		return ret;
 
-	if (!g_ServerBanSystem->IsUIDAllowed(nextPlayerUid))
+	if (IsDedicated() && !g_ServerBanSystem->IsUIDAllowed(nextPlayerUid))
 	{
 		CBaseClient__Disconnect(self, 1, "Banned from server");
 		return ret;
--- a/NorthstarDedicatedTest/bansystem.cpp
+++ b/NorthstarDedicatedTest/bansystem.cpp
@@ -36,17 +36,24 @@ void ServerBanSystem::ClearBanlist()
 	// reopen the file, don't provide std::ofstream::app so it clears on open
 	m_sBanlistStream.close();
 	m_sBanlistStream.open(GetNorthstarPrefix() + "/banlist.txt", std::ofstream::out | std::ofstream::binary);
+	m_sBanlistStream.close();
 }
 
 void ServerBanSystem::BanUID(uint64_t uid)
 {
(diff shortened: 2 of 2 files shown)


barnabwhy:
You could compare `std::filesystem::last_write_time` with the last time that Northstar read from or wrote to the file to reduce the frequency that the file is parsed. It might also be worth adding in unban writing.

barnabwhy:
Also I think adding the IsDedicated check is bad given we can't then use bans on listen servers, maybe just find a way to tell if you're in the lobby?

GeckoEidechse:
I'm updating branch to sync it with upstream.

GeckoEidechse (changes requested):
The issue with the current version as found in testing is that only reloads when banning/unbanning someone from console.

This excludes the case where someone would edit the `banlist.txt` file to ban someone. An example case would be where someone is hosting multiple servers that all use the same banlist, to avoid having to edit it for each server individually.
Therefore banlist should also be reloaded periodically based on some event or timer.

**EDIT:** A good place to reload banlist would be whenever a player connects (or disconnects).

GeckoEidechse:
Superseded by #233

NorthstarDedicatedTest/serverauthentication.cpp
```diff
@@ -388,7 +388,7 @@ bool CBaseClient__ConnectHook(void* self, char* name, __int64 netchan_ptr_arg, c
 	if (!ret)
 		return ret;
 
-	if (!g_ServerBanSystem->IsUIDAllowed(nextPlayerUid))
+	if (IsDedicated() && !g_ServerBanSystem->IsUIDAllowed(nextPlayerUid))
```
GeckoEidechse, line 391: Looking at this again I think this breaks banlist functionality completely for listen server ^^

PR #166, opened 2022-05-03
added simple check for command being emit

ScureX:
keeping this a draft cause you could just remove it and it'd work again so im looking into blocking it properly

issue: anyone can execute emit and thus spam servers with sounds. apparently a respawn fuck up.


BobTheBob9:
this wouldn't fix the issue at all, this only prevents you from entering the command into the source console clientside

PR #170, opened 2022-05-06
Block `emit` command when sv_cheats is disabled

emma-miler:


GeckoEidechse (commented):
Should the code in

- `NorthstarDedicatedTest/emit_blocker.cpp`
- `NorthstarDedicatedTest/emit_blocker.h`

maybe be moved into `NorthstarDedicatedTest/ExploitFixes.cpp` and `NorthstarDedicatedTest/ExploitFixes.h` respectively to not clutter up project with more source files?

GeckoEidechse:
Adding Bob and pg as reviewer cause y'all have a better understanding of Launcher codebase than I do :P

BobTheBob9:
looks fine to me

GeckoEidechse:
I still think it should be in an existing file (like exploit fixes) and not a new one just for the emit blocker but I also want this released so I'm merging anyway :I

PR #172, opened 2022-05-10
Implemented Json into the squirrel vm

RoyalBlue1:
Implemented JSON into squirrel vm. 
DecodeJSON takes JSON string and returns table. Encode Json takes table with string keys and returns JSON string. 
All values in table returned by Decode have to be casted first. For that purpose there are also functions included to cast from var to array or table.

Squirrel code for testing and as an example

```
void function jsonTest(){
	table data
	data["string"] <- "test"
	data["int"] <- 1234
	data["float"]<- 1.234
	data["array"]<- ["1","2","3"]
	table data1
	data1["string"] <- "test2"
	data1["int"] <- 4321
	data1["float"]<- 4.321
	data["table"]<- data1
	table data2 = DecodeJSON(EncodeJSON(data))
	printt("JSON Test",EncodeJSON(data2))
	table data3 = expect table(data2["table"])
	array arr = expect array(data2["array"])
	printt(string(data["string"]),string(data3["string"]),string(arr[1]))
	//output should be first the encoded json and then "test test2 2"
}
```


NorthstarDedicatedTest/scriptJson.cpp
```diff
@@ -0,0 +1,670 @@
...
+
+void sq_EncodeJSON_Table(rapidjson::Value* obj, SQTable* sqTable, rapidjson::Document* allocatorDoc);
+void sq_EncodeJSON_Array(rapidjson::Value* obj, SQArray* sqArray, rapidjson::Document* allocatorDoc);
+void serverSq_DecodeJSON_Table(void* sqvm, rapidjson::Value* obj);
+void serverSq_DecodeJSON_Array(void* sqvm, rapidjson::Value* arr);
+
+SQRESULT serverSq_DecodeJSON(void* sqvm)
+{
+	const char* json = ServerSq_getstring(sqvm, 1);
+	rapidjson::Document doc;
```
p0358, line 21: should use `rapidjson_document` instead of all `rapidjson::Document`

RoyalBlue1:
> Does it assume that root JSON value is an object? For example decoding an array string directly is meant to remain unsupported? Also is it something r2-exclusive that requires the values to be explicitly casted?

Yes to both questions. I have to hardcode the return type so it is either always array or always table. I chose table because that is more common and more flexible. And yes respawn added strong typing to squirrel exclusively for r2

NorthstarDedicatedTest/scriptJson.cpp
RoyalBlue1: Why? We already do it like that in modmanager.cpp and all the tutorial I could find do it like this as well. Also if I change it all member function can not be resolved anymore.

p0358: Because it's the version that uses Source's memory allocator. The goal was to avoid STL one everywhere, and this basically undoes that. I'm aware modmanager.cpp uses that, I told Bob to fix it but he ignored it eventually xD
Most projects have the default rapidjson::Document typedef with standard allocator, since they don't use custom allocators, and yeah in such case its usage changes a little bit, you can look into the places where rapidjson_document is used to see how the other stuff around was changed

p0358: It's defined in memalloc.h and memalloc.cpp; example usage: https://github.com/R2Northstar/NorthstarLauncher/blob/47ea109be0d73120ddfe6896494e1fce986515b3/NorthstarDedicatedTest/modmanager.cpp#L476

RoyalBlue1: Ok thanks for explaining. Is it better now?

BigSpice (commented):
Tried And Works on my end

emma-miler (approved):

F1F7Y:
I'm probably missing something, but whats the point of JSON support when we can't read files?

F1F7Y:
getting ignored like it's tinder

RoyalBlue1:
it can be used for convars also I was working on a Socket when I made this, but that is on the backburner for now until Rcon is merged.

GeckoEidechse (approved):
Confirmed working in testing.

PR #174, opened 2022-05-11
Log UID in a variety of places during auth process

GeckoEidechse:
Should help with getting a clearer picture regarding UID spoofing attacks.


EladNLG:
Shouldn't we probably tell users that their UIDs are logged?

barnabwhy:
should probably log in activateplayer

GeckoEidechse:
> Shouldn't we probably tell users that their UIDs are logged?

We've been logging UIDs since [`v1.5.1`](https://github.com/R2Northstar/Northstar/releases/tag/v1.5.1) via https://github.com/R2Northstar/NorthstarMods/pull/220 so this is nothing new.

Also keep in mind that UIDs are public information. Given an Origin playername one can use Origin APIs to extract that players UID. And as Origin playernames are required to be unique (albeit they can be changed by the user) this is not so different from just logging playernames which we've been doing since initial release :P

barnabwhy (approved):
Looks good to me. If any changes are needed it would be the "says UID" bits being changed to something more descriptive but that isn't a big deal.

PR #180, opened 2022-05-31
log the response to an attempt to auth with origin

ASpoonPlaysGames:
Idk why we don't already log the response to this, it means we don't get any errors in logs if we fail northstar origin auth
I put a check in place to avoid logging tokens as well


KittenPopo:
> Anything sensitive that could be in there? Just to make sure we don't have someone publicly upload their logs in let's say `#development` on the Northstar Discord as part of debugging an issue with Origin auth and with that leak some authentication token or sth.

But it doesn't log the token? https://github.com/R2Northstar/NorthstarLauncher/pull/180/commits/3710c5524323d3c353862db6276438ad17940ec9

KittenPopo (changes requested):
image


Why do we call `spdlog::error(...)` even when authentication succeeded? Am I missing something?

ASpoonPlaysGames:
> Why do you call spdlog::error(...) even when authentication succeeded? Am I missing something?

If the authentication succeeded, the response should have a token, so it shouldn't run that line.

If for some reason it succeeds, but doesn't send a token, it should probably be logged as an error anyway because a token is necessary for a successful auth response

KittenPopo:
> If the authentication succeeded, the response should have a token, so it shouldn't run that line.

Then shouldn't we not say `"Northstar origin authentication completed successfully!"`... 🤔

ASpoonPlaysGames:
That would probably be considered out of scope of the PR, as this PR is just about logging the response to the attempt (but not logging tokens)

In the event of getting a success:true but also no token back, it probably should come back as "Northstar origin authentication failed"

ASpoonPlaysGames:
I took the opportunity of the MS failing auth for people to test the PR a bit, got this as the result:
`
[error] Northstar origin authentication failed
[error] {"success":false,"error":{"enum":"STRYDER_RESPONSE","msg":"Couldn't parse stryder response"}}
[error] Failed reading masterserver response: got fastify error response
[error] {"success":false,"error":{"enum":"INVALID_MASTERSERVER_TOKEN","msg":"Invalid or expired masterserver token"}}
`

emma-miler:
Implemented elsewhere

PR #181, opened 2022-06-01
Set process mitigation policies for hardened security against potential RCE attacks

KittenPopo:
Windows has a set of features called "mitigation policies" for developers to use on their programs.
This allows the developer to prevent their program from being able to do certain "sketchy" things that it wouldn't need to normally, as a security measure in case of unauthorized code execution.

Since Titanfall 2 is a very exploitable game, and I've found quite a few potential RCE exploits, I think it is very important that Northstar have as many of these policies enabled as possible.

This pull request will cause the launcher to reboot itself immediately with various mitigation policies. I have tested Northstar with these policies, everything works fine.

**Policies Enabled:**
- Control Flow Guard (should be enabled by default in any release build, but just in case), prevents the control flow path from "escaping", which is a common attack on a program's security
- Disable all remote image loading (this should really never happen)
- Prevent system module loading from defaulting to modules in the default directory instead of system directories (I don't know why this is a thing by default, but, it's ridiculous for obvious reasons)
- Prevent manual syscalls (this is by far the most important, as this is enabled by default and would allow RCEs to call system functions manually straight into kernel in just 2 tiny instructions, _VERY_ bad)

image

This, along with my AntiRCE file check system in my other PR, should prevent small RCEs from being able to do much damage to a user, if any.

### DIFF
--- a/LauncherInjector/main.cpp
+++ b/LauncherInjector/main.cpp
@@ -276,8 +276,79 @@ HMODULE LoadDediStub(const char* name)
 	return h;
 }
 
+void StartSelfIfNeeded()
+{
+	if (strstr(GetCommandLineA(), "-selfstarted"))
+		return;
+
+#ifdef _DEBUG
+	return;
+#endif
+
+	// Start ourselves with various process mitigation policies
+	// See:
+	//  https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-setprocessmitigationpolicy
+	//  https://docs.microsoft.com/en-us/windows/security/threat-protection/override-mitigation-options-for-app-related-security-policies
+	//	https://www.ired.team/offensive-security/defense-evasion/preventing-3rd-party-dlls-from-injecting-into-your-processes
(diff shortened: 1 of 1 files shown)


GeckoEidechse:
Needs testing with Linux support and people using `-northstar` to start Northstar from Steam/Origin via wsock32.dll

barnabwhy:
> Needs testing with Linux support and people using `-northstar` to start Northstar from Steam/Origin via wsock32.dll

In theory you could simply add a check to the wsock proxy and make it reboot via northstarlauncher to ensure it ends up with the policies enabled

KittenPopo:
> Needs testing with Linux support and people using `-northstar` to start Northstar from Steam/Origin via wsock32.dll

Yea I did some more testing - it doesn't even seem to always enable the policies, even when set at startup. Sometimes it literally changes none of them, seemingly at random.
WinAPI moment... at least DEP is enabled by default for all x64 processes.

This PR needs some big improvements cause it's too inconsistent.

GeckoEidechse:
> This PR needs some big improvements cause it's too inconsistent.

Imma turn this PR into draft state then to prevent accidental merges. You should be able to change it back once ready <3

emma-miler:
Closing as stale

PR #168, opened 2022-05-04
Update to match recent LatencyFleX changes 

KyleGospo:
The main entry point for LatencyFleX was changed from latencyflex_wine.dll to latencyflex_layer.dll to prepare library for eventual windows release/to be more generic. This updates our implementation to match, while still supporting the previous libraries for existing users (As the official DXVK-NVAPI implementation of it does).

~~Currently in the process of testing, not ready to merge. Just opening this for visibility for now.~~

PR #183, opened 2022-06-07
LatencyFleX Implementation Update

KyleGospo:
Enhancement of previous PR #168 

Fixes a crash issue experienced in bottles when LatencyFleX fails to load due to an initialization error. Now handled gracefully.
As before, works with the original and modern release of LatencyFleX. Also features better logging than before.


pg9182 (approved):
LGTM

Issue #169, opened 2022-05-05
Server not retrying registration after failure


GeckoEidechse:
With https://github.com/R2Northstar/NorthstarMasterServer/pull/67 a server auto-restarting after a crash might be blocked by its ghosted hogging the `ip:port` combo until timeout. The problem is that if the restarted server doesn't retry for registration (with some timeout between attempts) it will never re-register and as such not display on server browser.

From a client's perspective it will look like the server never came back after crashing.

barnabwhy:
Here's what I know about this:
- If a server tries to register more than 30 seconds after the remnant of it's last heartbeat it should be added fine
- Calling server/remove_server from the dedicated server on close should fix this issue entirely

GeckoEidechse:
> Calling server/remove_server from the dedicated server on close should fix this issue entirely

That sadly wouldn't solve the case where a server hard-crashes and then gets restarted by e.g. docker-compose. In such a case due to the hard crash the server never had the chance to call `/server/remove_server`.

Simply retrying registration if it fails on start-up (depending on error response) seems like the simplest solution to me ^^
(We should obviously still have servers call `/server/remove_server` on normal shutdown.)

RomeoCantCode:
> Simply retrying registration if it fails on start-up (depending on error response) seems like the simplest solution to me ^^ (We should obviously still have servers call `/server/remove_server` on normal shutdown.)

https://github.com/RomeoCantCode/NorthstarLauncher/commit/37b38d6d52deef7f095bcc0314c4de7621a339ec
😄 

Also I think the server should check if it was gracefully shut down before. If not it should call `/server/remove_server` using the previous serverID (would require storing the serverID and like a gracefulshutdown flag on FS).

Also as a side note. If you gracefully shut down the server (using `quit` command) it will call `/server/remove_server` but it seems not doing so if you close the console / close server process in some other way.

GeckoEidechse:
> using the previous serverID (would require storing the serverID and like a gracefulshutdown flag on FS).

I doubt the server could save ID on crash if it can't call `/server/remove`. We could obviously store it before but we have edge cases like the docker container, where most files are immutable and not stored back to disk. I think the 30s wait to re-register server on crash is honestly fine, so as long as we retry we shouldn't really have an issue. ^^

GeckoEidechse:
[...]

PR #187, opened 2022-06-13
Attempt reconnect in case of DUPLICATE_SERVER

GeckoEidechse:
If gameserver receives DUPLICATE_SERVER response it will retry the auth request after a timeout of 10 seconds.
Retries will stop after 10 attempts.

Uses `GOTO`s to achieve this which is not particularly nice but I assume acceptable as we use the already as well in this part of the code.

For testing I spun a dedicated server, killed it after auth finished, then immediately created a new one. Output shows that it retried after failure until it was able to connect on the 2nd attempt. Then joining the server and playing around with it worked as expected.

screenshot

### DIFF
--- a/NorthstarDedicatedTest/masterserver.cpp
+++ b/NorthstarDedicatedTest/masterserver.cpp
@@ -14,6 +14,7 @@
 #include <cstring>
 #include <regex>
 #include "version.h"
+#include <chrono>
 // NOTE for anyone reading this: we used to use httplib for requests here, but it had issues, so we're moving to curl now for masterserver
 // requests so httplib is used exclusively for server stuff now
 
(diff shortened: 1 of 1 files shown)


emma-miler:
Fixes #169

NorthstarDedicatedTest/masterserver.cpp
```diff
@@ -903,6 +908,29 @@ void MasterServerManager::AddSelfToServerList(
 				{
 					spdlog::error("Failed reading masterserver response: got fastify error response");
 					spdlog::error(readBuffer);
+
+					// Check for enum member in JSON
+					if (serverAddedJson["error"].HasMember("enum") && retry_counter < 1)
```
GeckoEidechse, line 913:
```suggestion
					if (serverAddedJson["error"].HasMember("enum") && retry_counter < 10)
```

RoyalBlue1 (approved):

PR #189, opened 2022-06-14
fix status command and add UID to status output

TH3-S4LM0N:
Made this for rolesspawn cause github
It adds UIDs to the status command and makes it work again, like Counter-Strike: Source. I noticed it was missing in Northstar, so I broke out a debugger to figure out why. It turns out some commands are "engine client" commands, and the existing console command check didn't check for those. I tried to copy the codestyle I saw in other files, so please let me know if I made any mistakes.


NorthstarDedicatedTest/logging.cpp
```diff
@@ -473,18 +466,17 @@ void __fastcall Host_Status_PrintClient_Hook(__int64 client, char addresses, voi
...
+
 	if (nci)
 	{
-		rate = *(unsigned int*)(nci + 232);
-		loss = (int)(float)((*(float*)(6712 * 1 + nci + 500)) * 100.0);
-		ping = (int)(float)((*(float*)(6712 * 0 + nci + 508)) * 1000.0);
-		connected = COM_FormatSeconds(GetTimeConnected(nci));
+		int rate = *(unsigned int*)(nci + 232);
+		int loss = (int)(float)((*(float*)(6712 * 1 + nci + 500)) * 100.0);
+		int ping = (int)(float)((*(float*)(6712 * 0 + nci + 508)) * 1000.0);
```
p0358, line 474: Is the two-stage cast here needed rather than directly casting it to int?

NorthstarDedicatedTest/serverauthentication.cpp
```diff
@@ -511,7 +515,7 @@ char CGameClient__ExecuteStringCommandHook(void* self, uint32_t unknown, const c
 	ConCommand* command = g_pCVar->FindCommand(tempCommand.Arg(0));
 
 	// if the command doesn't exist pass it on to ExecuteStringCommand for script clientcommands and stuff
-	if (command && !command->IsFlagSet(FCVAR_CLIENTCMD_CAN_EXECUTE))
+	if (command && !command->IsFlagSet(FCVAR_CLIENTCMD_CAN_EXECUTE) && !CGameClient__IsEngineClientCommand(NULL, tempCommand))
```
p0358, line 518: It would probably be better and safer to try adding the flag `FCVAR_CLIENTCMD_CAN_EXECUTE` to the `status` command itself rather than this

GeckoEidechse:
Response to above comments on Discord by commit author:
https://discord.com/channels/920776187884732556/950322078945538058/986267580526374952

p0358 (approved):
It seems fine now, the list of allowed concommands *should* be harmless, and the commands that are allowed and where they're defined in engine.dll is now commented above the declaration of `CGameClient__IsEngineClientCommand` for easy review of what that command checks in case someone needs to revisit concommands stuff in the future

NorthstarDedicatedTest/logging.cpp
GeckoEidechse: Fixed in 9391bf3da1d048543cbe7a8a7d773ae6e6d83b51

NorthstarDedicatedTest/serverauthentication.cpp
GeckoEidechse: From @rolelessweapon on Discord:

> The same CGameClient::IsEngineClientCommand check exists in other Source engine games.
> You'd also have to add the `FCVAR_CLIENT_CAN_EXECUTE` command to ping as well, which would work, but feels clunky rather than reproducing the existing check

https://discord.com/channels/920776187884732556/950322078945538058/986267580526374952

GeckoEidechse:
Closing as @rolelessweapon now has sufficient access to open PRs under her own name.

@rolelessweapon please open new PR with the same branch and link to this PR.

PR #190, opened 2022-06-14
Disable DevTextBuffer by default

emma-miler:
This is code ported from an older PR that is still in development. Could possibly look at checking if `sv_cheats` is enabled, but adding any kind of branching in the inner function seems to crash on a stack canary for some reason.

### DIFF
--- a/NorthstarDedicatedTest/miscclientfixes.cpp
+++ b/NorthstarDedicatedTest/miscclientfixes.cpp
@@ -9,7 +9,6 @@ CrashingWeaponActivityFuncType CrashingWeaponActivityFunc1;
 typedef bool (*DevTextBufferDumpToFileType)(int a1);
 DevTextBufferDumpToFileType DevTextBufferDumpToFileClient;
 
-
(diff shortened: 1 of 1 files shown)


emma-miler:
As Amos pointed out in discord, the first parameter to the function is a pointer and uses `uint64_t` and not `int`, which is why it would crash whenever I added any code, as it would overflow. Working on a fix and adding a convar to change behaviour.

emma-miler:
In this case `char*` would be better right? since it is pointing to a `char*` that contains the path

Mauler125:
Yes, as it adds context. Its still pretty much the same otherwise.

cpdt (approved):

r-ex:
this isn't a complete fix - DevTextBufferWrite should be behind the same convar as you can use it to endlessly write into memory and cause an OOM crash on the client. (+ there's no point in being able to write into the buffer if you can't also dump to file)

RoyalBlue1:
isnt it supper easy to set a convar from squirrel. so this fix wont do anything

GeckoEidechse:
Superseded by #211

PR #191, opened 2022-06-14
Guard against GetEntByIndex read out of bounds

cpdt:
`0x4000` is used as a maximum for other code reading the same entity array, so seems fine to use here too.


emma-miler (approved):

cpdt:
I've had this deployed on my servers for the past few hours, no issues and no crashes.

NorthstarDedicatedTest/ExploitFixes.cpp
```diff
@@ -347,6 +347,20 @@ KHOOK(
...
+// GetEntByIndex (called by ScriptGetEntByIndex) doesn't check for the index being out of bounds when it's
+// above the max entity count. This allows it to be used to crash servers.
+typedef void*(__fastcall* GetEntByIndexType)(int idx);
+GetEntByIndexType GetEntByIndex;
+
+static void* GetEntByIndexHook(int idx)
+{
+	if (idx >= 0x4000)
+	{
+		return nullptr;
```
RoyalBlue1, line 359: maybe a error message here like "GetEntByIndex {} is out of bounds" or something like that

RoyalBlue1 (approved):

NorthstarDedicatedTest/ExploitFixes.cpp
emma-miler, line 359:
```suggestion
		spdlog::info("GetEntByIndex {} is out of bounds", idx);
		return nullptr;
```

GeckoEidechse: Resolved in 50c919ec9037e9a8bbc9fee888f35f66ce3379de

PR #192, opened 2022-06-14
Fix bug where emit wouldnt be blocked if uppercase

emma-miler:
temp fix while we wait for #178 to not suck i guess


NorthstarDedicatedTest/ExploitFixes.cpp
```diff
@@ -394,7 +397,34 @@ void DoBytePatches()
...
+	{
+		unsigned char c1 = static_cast<unsigned char>(s1[i]);
+		unsigned char c2 = static_cast<unsigned char>(s2[i]);
+		if (tolower(c1) != tolower(c2))
+			return false;
+		if (c1 == '\0' || c2 == '\0')
+			break;
+	}
+	return true;
+}
```
cpdt, line 413: Can't you use `strnicmp`?

emma-miler: Updated ^^

NorthstarDedicatedTest/ExploitFixes.cpp
```diff
@@ -394,7 +397,19 @@ void DoBytePatches()
 	}
 }
 
-void ExploitFixes::LoadCallback(HMODULE unused)
+char function_containing_emit_hook(uint64_t unknown_value, uint64_t command_ptr)
+{
+	char* command_string = *(char**)(command_ptr + 1040); // From decompile
```
cpdt, line 402: Not really important but I think `command_ptr` is a `CCommand`, and we do have a class definition for that so you don't need to read a byte offset.

cpdt (approved):

NorthstarDedicatedTest/ExploitFixes.cpp
emma-miler: ah neat. I think this was written before that code ever got merged. I know in popo's PR he uses that definition, but honestly i cant really be arsed to change it here, since this will be removed when that one gets merged anyway.

cpdt:
I've had this deployed on my servers for the past few hours, seems to be working great 👍

KittenPopo:
image

bruh what

emma-miler:
> image
> 
> bruh what

homie i literally already removed it the fuck you complaining about

tomb-msft:
> I've had this deployed on my servers for the past few hours, seems to be working great 👍

@GeckoEidechse

PR #194, opened 2022-06-14
Delete emit_blocker.h

RoyalBlue1:
I completly and totally eradicated emit_blocker.h because it isn't used anymore


0xMihir:
You need to remove it from the project too.

emma-miler (approved):

PR #195, opened 2022-06-15
Move LZSS exploit fix hook to ExploitFixes.cpp

KittenPopo:
For some reason, people seem to like to add exploit fixes outside of the dedicated file for exploit fixes.

Please move the new LZSS exploit patch hook into ExploitFixes.cpp!


GeckoEidechse:
@cpdt pinging cause this touches your stuff
(GitHub won't let me request review from you for some reason)

GeckoEidechse:
Note to self, when merging add co-author line crediting cpdt so that git blame looks proper if we need to find (original) author.

tomb-msft (approved):
LGTM

Fwiw this fix isn’t new though, it was sitting there for like two months 🙃

cpdt:
^ should probably log out of my work account lol

x3Karma:
cpdt irl isn't a bongo cat? :(

GeckoEidechse:
No issues in small testing and cpdt says it's good, so merging ^^

PR #196, opened 2022-06-15
Set default val for net_chan_limit_msec_per_sec

KittenPopo:
Setting a default value for _net_chan_limit_msec_per_sec_ will fix the issue of singleplayer games spamming warnings in the console.


GeckoEidechse (approved):
Works in testing. Previously spams console, afterwards no longer does.

pg9182 (approved):

PR #198, opened 2022-06-17
Use strtoull instead of strtoll for uint64s

KittenPopo:
I have no proof that this will actually ever cause a real problem, but it's definitely a good idea and is more proper.

### DIFF
--- a/NorthstarDedicatedTest/bansystem.cpp
+++ b/NorthstarDedicatedTest/bansystem.cpp
@@ -20,7 +20,7 @@ void ServerBanSystem::OpenBanlist()
 	{
 		std::string line;
 		while (std::getline(enabledModsStream, line))
-			m_vBannedUids.push_back(strtoll(line.c_str(), nullptr, 10));
+			m_vBannedUids.push_back(strtoull(line.c_str(), nullptr, 10));
 
 		enabledModsStream.close();
 	}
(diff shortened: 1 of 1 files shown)


KittenPopo:
> Note to self:
> 
> * `strtoll`: Convert string to long long integer
> * `strtoull`: Convert string to _unsigned_ long long integer

`strtoll` = `str to L L (long long)`
`strtoull` = `str to U L L (unsigned long long)`
`strtoul` = `str to L (long/int)`

Once you realize the pattern its pretty helpful xD

GeckoEidechse:
To hit those lines in testing I assume I just to put someone on banlist and then have them join server? ^^

KittenPopo:
> To hit those lines in testing I assume I just to put someone on banlist and then have them join server? ^^

... Yes. 

Not sure why a 1-character function usage switch of a super common STL function needs testing, but go for it xD

GeckoEidechse:
> > To hit those lines in testing I assume I just to put someone on banlist and then have them join server? ^^
> 
> ... Yes.
> 
> Not sure why a 1-character function usage switch of a super common STL function needs testing, but go for it xD

Eh, like the issue is not gonna be in a standard lib function. More so in the change from one function to the other. And like I don't expect anything to break either. It's mostly about establishing a pattern of testing every PR for the stuff it changes.

Anyway, noticed no issues in testing ^^

PR #199, opened 2022-06-17
Add and implement GET_OFFSET/GET_OFFSET_PTR macro

KittenPopo:
A lot of code in northstar seems to use this really ugly format of `(type*)((char*)address + offset)` or whatever.

I made two simple macros to make it simpler and more readable to convert/interpret pointers at offsets and replaced the vast majority of the goofy `(char*)` offsets with them.


KittenPopo:
Closed due to fair point by Ice: 
image

This shouldn't be a macro.

PR #202, opened 2022-06-19
fixed custom rpaks being loaded more than once

### LINKED ISSUE #201: Preloading 2 rpaks can cause infinite loop on launch
Each time an rpak gets loaded the preloaded rpaks get reloaded, having 2 of them causes an infinite loop of them causing each other to get loaded it seems

ASpoonPlaysGames:
Potentially fixes #201 ? Definitely fixes an issue where every custom rpak was being loaded every time any other rpak was loaded ever, which was preventing us from having more than 1 custom rpak

### DIFF
--- a/NorthstarDedicatedTest/rpakfilesystem.cpp
+++ b/NorthstarDedicatedTest/rpakfilesystem.cpp
@@ -74,7 +74,6 @@ void PakLoadManager::RemoveLoadedPak(int32_t pakHandle)
 	loadedPaks.erase(pakHandle);
 }
 
-
 void HandlePakAliases(char** map)
 {
 	// convert the pak being loaded to it's aliased one, e.g. aliasing mp_hub_timeshift => sp_hub_timeshift
@@ -145,8 +144,6 @@ void LoadCustomMapPaks(char** pakName, bool* bNeedToFreePakName)
 		}
 	}
 }
-
-
 
 LoadPakSyncType LoadPakSyncOriginal;
 void* LoadPakSyncHook(char* path, void* unknownSingleton, int flags)
--- a/NorthstarDedicatedTest/rpakfilesystem.cpp
+++ b/NorthstarDedicatedTest/rpakfilesystem.cpp
@@ -119,7 +119,6 @@ void LoadCustomMapPaks(char** pakName, bool* bNeedToFreePakName)
 	}
 }
 
-#include "pch.h"
 // these are hashed with STR_HASH
 std::unordered_map<int32_t, size_t> loadedPaks {};
 
@@ -141,7 +140,7 @@ void* LoadPakSyncHook(char* path, void* unknownSingleton, int flags)
 		LoadPreloadPaks();
 		LoadCustomMapPaks(&path, &bNeedToFreePakName);
 
-		bShouldLoadPaks = true; // why are we ever loading these more than once? i dont understand why we were setting this to true again
+		bShouldLoadPaks = true;
 	}
 
 	spdlog::info("LoadPakSync {}", path);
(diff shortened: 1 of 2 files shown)


NorthstarDedicatedTest/rpakfilesystem.cpp
```diff
@@ -165,7 +165,7 @@ int LoadPakAsyncHook(char* path, void* unknownSingleton, int flags, void* callba
 		LoadPreloadPaks();
 		LoadCustomMapPaks(&path, &bNeedToFreePakName);
 
-		bShouldLoadPaks = true;
+		// bShouldLoadPaks = true; // why are we ever loading these more than once? i dont understand why we were setting this to true again
```
BobTheBob9, line 168: wouldn't this break mod pak loading entirely after like, literally the first one the game loads?

BobTheBob9: we set bShouldLoadPaks to false at all to prevent recursive LoadPak calls, but it needs to be set back to true afterwards so subsequent calls will still try to load mod paks

BobTheBob9 (changes requested):
if the goal is preventing single paks from loading multiple times at once, this should be done through a hashset or something to track what paks are loaded at a given time, not just preventing non-preload paks from loading at all

NorthstarDedicatedTest/rpakfilesystem.cpp
ASpoonPlaysGames: ah ok, wasnt sure how non-preload paks were loaded, will fix

NorthstarDedicatedTest/rpakfilesystem.cpp
```diff
@@ -137,7 +141,7 @@ void* LoadPakSyncHook(char* path, void* unknownSingleton, int flags)
 		LoadPreloadPaks();
 		LoadCustomMapPaks(&path, &bNeedToFreePakName);
 
-		bShouldLoadPaks = true;
+		bShouldLoadPaks = true; // why are we ever loading these more than once? i dont understand why we were setting this to true again
```
BobTheBob9, line 144: remove comment

NorthstarDedicatedTest/rpakfilesystem.cpp
```diff
@@ -119,6 +119,10 @@ void LoadCustomMapPaks(char** pakName, bool* bNeedToFreePakName)
 	}
 }
 
+#include "pch.h"
+// these are hashed with STR_HASH
+std::unordered_map<int32_t, size_t> loadedPaks {};
```
BobTheBob9, line 123: would be better to use some kind of bidirectional map here but i dont think standard libs have one, could use 2 maps?
also there's no reason this shouldn't be in the PakLoadManager

NorthstarDedicatedTest/rpakfilesystem.cpp
```diff
@@ -119,6 +119,10 @@ void LoadCustomMapPaks(char** pakName, bool* bNeedToFreePakName)
 	}
 }
 
+#include "pch.h"
```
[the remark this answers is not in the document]
ASpoonPlaysGames: i was getting some errors when i didn't before for some reason, was likely just me misusing STR_HASH at the time though, and forgot to remove it

NorthstarDedicatedTest/rpakfilesystem.cpp
```diff
@@ -185,6 +212,12 @@ int LoadPakAsyncHook(char* path, void* unknownSingleton, int flags, void* callba
 UnloadPakType UnloadPakOriginal;
 void* UnloadPakHook(int pakHandle, void* callback)
 {
+	if (loadedPaks.find(pakHandle) != loadedPaks.end())
```
BobTheBob9, line 215: this should be done in PakLoadManager

PR #203, opened 2022-06-19
ExploitFixes: Fix Rejection of Legitimate UserCmds with invalid fields

KittenPopo:
For some reason, Respawn frequently uses UserCmds with NAN floats or whatever, and they are never rejected.

This occurs when grappling or when in thirdperson.

This PR doesn't reject CMDs with NAN floats anymore, but rather just sets the nan floats to zero to prevent a particular crash exploit.
This should not affect in-game play whatsoever. I will test this more thoroughly later.

**Summary of Changes:**
- Made ExploitFixes not reset a CMD if float components are NAN, instead just change any NAN floats to zero
- Removed unnecessary vector/angle structs, replaced with a single Float3 struct


GeckoEidechse:
This should also address issues related to grappling entities/pilots, right? ^^

PR #204, opened 2022-06-19
ExploitFixes: Fix respawn exploit and block other Portal 2 commands

KittenPopo:
Titanfall 2 has the Portal 2 command-processing system still built in.

This is what leads to the respawn exploit, sound playing exploit, etc.

The sound playing exploit was already fixed in a recent commit (although I despise the complete disregard for following established standards/formatting), my hook is of the same function and fixes other exploitative commands as well.

Sorry guys, no more respawning in front of your killers with a keybind lol. It was fun while it lasted.

### DIFF
--- a/NorthstarDedicatedTest/ExploitFixes.cpp
+++ b/NorthstarDedicatedTest/ExploitFixes.cpp
@@ -5,7 +5,6 @@
 #include "NSMem.h"
 #include "cvar.h"
 
-
 ConVar* ns_exploitfixes_log;
 #define SHOULD_LOG (ns_exploitfixes_log->m_Value.m_nValue > 0)
 #define BLOCKED_INFO(s)                                                                                                                    \
@@ -436,6 +435,7 @@ KHOOK(
 		std::string cmdStr = command->Arg(0);
 		for (char& c : cmdStr)
 			c = tolower(c);
+
 		for (const char* blockedCommand : blockedCommands)
 		{
 			if (cmdStr.find(blockedCommand) != std::string::npos)
--- a/NorthstarDedicatedTest/ExploitFixes.cpp
+++ b/NorthstarDedicatedTest/ExploitFixes.cpp
@@ -427,6 +427,8 @@ KHOOK(
 		"emit", // Sound-playing exploit lol
 		"pre_go_to_hub", // Oh dear
 		"erase_mp_progress", // No
+		"pre_go_to_calibration", // Executes a command for every single entity for some reason, nice one valve
+		"end_movie", // This is not gmod
 		"load_recent_checkpoint" // This is the instant-respawn exploit
 	};
 
--- a/NorthstarDedicatedTest/ExploitFixes.cpp
+++ b/NorthstarDedicatedTest/ExploitFixes.cpp
@@ -5,9 +5,7 @@
 #include "NSMem.h"
 #include "cvar.h"
 
-typedef char(__fastcall* function_containing_emit_t)(uint64_t a1, uint64_t a2);
-function_containing_emit_t function_containing_emit;
-ConVar* sv_cheats;
+
 ConVar* ns_exploitfixes_log;
 #define SHOULD_LOG (ns_exploitfixes_log->m_Value.m_nValue > 0)
 #define BLOCKED_INFO(s)                                                                                                                    \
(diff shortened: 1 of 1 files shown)


NorthstarDedicatedTest/ExploitFixes.cpp
```diff
@@ -412,16 +409,47 @@ void DoBytePatches()
 	}
 }
 
-char function_containing_emit_hook(uint64_t unknown_value, uint64_t command_ptr)
+KHOOK(
```
GeckoEidechse, line 412: Not sure what our current stance on `KHOOK` vs `ENABLER_CREATEHOOK` is. AFAIK we prefer offsets cause faster and Titanfall2 will likely never see updates any more anyway so pattern matching isn't really needed.

Then there's Bob's refactor which apparently changes the way we do hooks anyway.

And even if we use offsets, we should probably still paste the pattern as a comment above in the unlikely case Respawn does update `Titanfall2.exe` and we suddenly have to change all hooks again.

&nbsp;

Basically, I'm trying to say that someone with hooking experience in NS needs to take a look at this :P

KittenPopo: I have hooking experience in NS lol, I've been hooking functions in source games for years.

KHook supports offsets in my recent NSMem improvements PR, so it's usable everywhere.

Hooking WITHOUT KHook:
- Create a typedef for the function
- Create a global-ish variable to store the original function pointer
- Manually offset from the base address at a module init to get the pointer to hook
- Manually create and name a function for the hook function
- Manually call ENABLER_CREATEHOOK with a bunch of args when you want the hook to be set up

```
// Make typedef manually
typedef char(__fastcall* function_containing_emit_t)(uint64_t a1, uint64_t a2);

// Make global-ish variable manually for original function pointer
function_containing_emit_t function_containing_emit;

// Make hook function
char function_containing_emit_hook(uint64_t unknown_value, uint64_t command_ptr)
{
	...
	
	// Original call return
	return function_containing_emit(unknown_value, command_ptr);
}

// Create a load callback function declaration in a .h
void LoadCallback(HMODULE unused); 

// Create a load callback definition in .cpp
void Something::LoadCallback(HMODULE baseAddress)
{
	// Call ENABLER_CREATEHOOK (why is this a macro again?) with relevant args
	ENABLER_CREATEHOOK(
		hook, (char*)baseAddress + 0x5889A0, &function_containing_emit_hook, reinterpret_cast<LPVOID*>(&function_containing_emit));
}

// Import your header in dll main and add load callback among the sea of other load callbacks (aaaa)
AddDllLoadCallback("server.dll", Something::LoadCallback);
```

Hooking WITH khook:
- Use the KHOOK macro to make the function, fill in module name, offset or signature, func name, ret type, conversion, and args in the macro parenthesis 
- Write hook code in that macro-function
- An original function pointer is automatically made for you with o##TheFunctionName
- Hook function is automatically created as well, named hk##TheFunctionName
- No need to put any code related to that hook anywhere else, it's that easy 😎 

```
// Make your KHOOK macro in a cpp or something
KHOOK(
	SpecialClientCommand, // Function name
	("server.dll", "48 89 5C 24 ? 48"), // DLL and signature (or offset!)
	bool, __fastcall, (void* player, CCommand* command) // Function typedef args (pretty intuitive)
	)
{

	...

	// Original call (oSpecialClientCommand has been created for you and will be set automatically)
	return oSpecialClientCommand(player, command);
}


// That's it. That's all. It'll initialize automatically and work as intended.
```

I made KHook as a substantially simpler, cleaner, and far more readable/understandable alternative to the verbose hooking process that NS was using (and still does).

BobTheBob9: i think this is fine for now since it's what's already being used across exploitfixes anyway, my plan is to use a similar api to this for hooking globally once the refactor stuff is done
(also, why does github's resolve conversation button not post a comment if it's in the text box for one? weird behaviour)

BobTheBob9: also, sigs could be potentially good in the future anyway if we want to mess with older versions of tf2, or if somehow we get a dev/internal build (though this probably will not happen ever)

GeckoEidechse: > I have hooking experience in NS lol, I've been hooking functions in source games for years.
> 
> KHook supports offsets in my recent NSMem improvements PR, so it's usable everywhere.

I mostly wanted feedback from others as it's more of a design decision then a technical one :P
Anyway Bob says it's ok so fine by me.

Also, that response is so detailed, I'm stealing it for launcher documentation >:3

NorthstarDedicatedTest/ExploitFixes.cpp
```diff
@@ -412,16 +409,47 @@ void DoBytePatches()
...
+	if (sv_cheats->GetBool())
+		return oSpecialClientCommand(player, command); // Don't block anything if sv_cheats is on
+
+	// These are mostly from Portal 2 (sigh)
+	constexpr const char* blockedCommands[] = {
+		"emit", // Sound-playing exploit lol
+		"pre_go_to_hub", // Oh dear
+		"erase_mp_progress", // No
+		"pre_go_to_calibration", // Executes a command for every single entity for some reason, nice one valve
+		"end_movie", // This is not gmod
```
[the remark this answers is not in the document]
KittenPopo: Fair enough, done in new commit.
Some of them I'm not entirely sure and would need to reverse/test in Portal 2 - not entirely sure the extent to which they can actually do anything, but two of these execute a command and call a function for every single entity in the entity list without hesitation.

NorthstarDedicatedTest/ExploitFixes.cpp
```diff
@@ -412,16 +409,47 @@ void DoBytePatches()
...
+	static ConVar* sv_cheats = g_pCVar->FindVar("sv_cheats");
+
+	if (sv_cheats->GetBool())
+		return oSpecialClientCommand(player, command); // Don't block anything if sv_cheats is on
+
+	// These are mostly from Portal 2 (sigh)
+	constexpr const char* blockedCommands[] = {
+		"emit", // Sound-playing exploit lol
+		"pre_go_to_hub", // Oh dear
+		"erase_mp_progress", // No
```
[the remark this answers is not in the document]
KittenPopo: (Resolved in similar)

KittenPopo:
> Might also be of use to add the name/uid of the player trying to run the command to the log output.

Command-username logging, in general, should be implemented - someone else was suggesting that and I think it's a better idea to have that be it's own thing.

Also, I'm not sure it's going to be super helpful - random people might see this PR, and try these commands - doesn't necessarily mean they are an unhinged exploiter looking to destroy servers.

I do think we should have global command logging though.
Something like:
```cpp
KHOOK(OnPlayerExecuteCmd, ("server.dll", "69 04 20"), (void* player, const char* cmd)) {
    spdlog::info("Player \"{}\" executed command \"{}\"", GetPlayerName(player), cmd);
}
```

BobTheBob9:
> Might also be of use to add the name/uid of the player trying to run the command to the log output.

this can be in another pr, i think it's infinitely better to merge this now if we've shown it already works

PR #205, opened 2022-06-19
Implement AntiRCE: Malicious file access/write detection

KittenPopo:
This is a system I made a few months ago, from le old giant PR.

Currently, it only checks file access, to prevent Titanfall from accessing files it should never access.

If an attacker utilizes an RCE, reading/creating files is a great go-to for doing damage, spreading a malicious binary, or stealing user information (i.e. discord token).

AntiRCE stops any file access outside of Titanfall's directory and user data folder (apart from loading system fonts).

This PR still needs more testing to find any potential false positives, but has proved very effective thus far.

**Overview of AntiRCE system:**
- Hook filesystem_stdio file handle opening at the lowest level
- When this hook fires, determine the true path of the target file and compare it against Titanfall's installation and user data folders (also check file type)
- If the file is outside of where Titanfall should be able to access, or is a binary/executable file (i.e. RCE trying to overwrite game DLLs), then the whole process is halted, an error is displayed, and the game will be shut down after with a custom exit code (666).

**New Commit Addition:**
- Hook all file accesses via the lowest level before kernel transition (Nt/ZwCreateFile)
- Check if the file creation access has write permissions
- If it's write access, and the file has a binary extension, trigger the AntiRCE emergency detection

_NOTE: There is a very slight possibility that some wacky game-injected external system/overlay/whatever writes to a binary file, leading to a false-positive, however I believe this is extremely rare and would be very unusual._

**Testing:**
I have tested the AntiRCE system quite a bit myself, with as many overlay programs and injected (legimiate) software I could find - I have yet to find any false positives.
However, it is important that others test this PR incase there is any genuine false positive that needs to be checked for that I didn't encounter or think of.


RoyalBlue1:
I don't know if I missed something but can you maybe try to print some debug info like callstack or something when the hook is triggered

NorthstarDedicatedTest/AntiRCE.cpp
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	int pathLen = strlen(path);
+
+	const char* extension = strrchr(path, '.');
+	if (extension)
+	{
+		if (!readOnly)
+		{
+			constexpr const char* EVIL_EXTENTIONS[] = {".dll", ".exe", ".bat", ".cmd", ".com"};
```
p0358, line 44: add .ps1, .vbs, .vb, possibly .js and .sys, .msi, .scr, .lnk, .cpl

KittenPopo:
> I don't know if I missed something but can you maybe try to print some debug info like callstack or something when the hook is triggered

I think you might have missed something.

image
image
image

NorthstarDedicatedTest/AntiRCE.cpp
KittenPopo: Adding in new commits, thx

KittenPopo: Added in new commit, thanks!

NorthstarDedicatedTest/AntiRCE.cpp
```diff
@@ -0,0 +1,248 @@
+#include "pch.h"
+#include "AntiRCE.h"
+#include "cvar.h"
+
+// These file extensions should NEVER be written to by the game, no matter what
+constexpr const char* EVIL_EXTENTIONS[] = {".dll", ".exe", ".bat", ".cmd", ".com", ".ps1", ".vbs", ".vb",  ".sys", ".msi", ".src",
+										   ".lnk", ".ps1", ".vbs", ".vb",  ".js",  ".sys", ".msi", ".scr", ".lnk", ".cpl"};
```
pg9182, line 7: You should probably add `.hta` to the list.

NorthstarDedicatedTest/AntiRCE.cpp
```diff
@@ -0,0 +1,248 @@
...
+#endif
+
+	// clang-format off
+	MessageBoxA(NULL, (
+			"AntiRCE EMERGENCY SHUTDOWN:\n" + msg + 
+			"\n\nIt is HIGHLY reccommended that you notify northstar developers/staff of this issue (unless you are sure it's fine)."
+		).c_str(), 
+		"Northstar AntiRCE", 
+		MB_ICONERROR
+	);
```
GeckoEidechse, line 219: I get that it is important that we get informed of an e.g. out-of-bounds write but I'm not sure this is the best approach. A crash and remote logger like [Sentry.io](https://sentry.io/) might be a better solution but we don't have support for that yet.

For now, I'd just drop the beeping noise.

KittenPopo: i fix 👍

ASpoonPlaysGames:
Closing as stale

PR #206, opened 2022-06-21
Add returntype enum to squirrel native functions

RoyalBlue1:
This PR should only be merged by @BobTheBob9

It adds a return type enum to the registration of native functions in squirrel. This is required for to return types like asset without needing to expect cast the return value first.

It will set the enum value depinding on the string return type and will keep the 32 value set before when the enum type is unknown

### DIFF
--- a/NorthstarDLL/squirrel.cpp
+++ b/NorthstarDLL/squirrel.cpp
@@ -401,6 +401,6 @@ int GetReturnTypeEnumFromString(const char* name)
 	}
 	else
 	{
-		return SqReturnStringOrNull; // previous default value
+		return SqReturnDefault; // previous default value
 	}
 }
--- a/NorthstarDLL/squirrel.cpp
+++ b/NorthstarDLL/squirrel.cpp
@@ -381,25 +381,26 @@ ON_DLL_LOAD_RELIESON("server.dll", ServerSquirrel, ConCommand, [](HMODULE baseAd
 		FCVAR_GAMEDLL | FCVAR_GAMEDLL_FOR_REMOTE_CLIENTS | FCVAR_CHEAT);
 })
 
-int GetReturnTypeEnumFromString(const char* name) {
-	
+int GetReturnTypeEnumFromString(const char* name)
+{
 
 	static std::map<std::string, SQReturnTypeEnum> sqEnumStrMap = {
-		{"bool",	SqReturnBoolean },
-		{"float",	SqReturnFloat },
-		{"vector",	SqReturnVector },
-		{"int",		SqReturnInteger },
-		{"entity",	SqReturnEntity },
-		{"string",	SqReturnString },
-		{"array",	SqReturnArrays },
-		{"asset",	SqReturnAsset },
-		{"table",	SqReturnTable }
-	};
+		{"bool", SqReturnBoolean},
+		{"float", SqReturnFloat},
+		{"vector", SqReturnVector},
+		{"int", SqReturnInteger},
+		{"entity", SqReturnEntity},
+		{"string", SqReturnString},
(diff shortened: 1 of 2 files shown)


KittenPopo:
Manually running strcmp on every single possiblity gang ✊

KittenPopo (changes requested):
Consider using a map instead of manually using strcmp on every combination.

Example:
```cpp
int getReturnTypeEnumFromString(const char* name) {
	using SQReturnTypeEnum;

	static std::map<std::string, SQReturnTypeEnum> sqEnumStrMap = {
		{"bool",	SqBoolean },
		{"float",	SqFloat },
		{"vector",	SqVector },
		{"int",		SqInteger },
		{"entity",	SqEntity },
		{"string",	SqString },
		{"array",	SqArrays },
		{"asset",	SqAsset },
		{"table",	SqTable }
	};

	if (sqEnumStrMap.count(name)) {
		return sqEnumStrMap[name];
	} else {
		return SqStringOrNull;
	}
}
```
Also, Northstar doesn't seem to use camelCase for function names... Let me guess, Java programmer?
Doesn't matter a ton tho, I just hate Java lol

Also remember to run clang-format before making PRs so you can get that epic 2/2 checks passed victory royale!11!
Easy commands for this at the end of: https://github.com/R2Northstar/NorthstarLauncher/blob/main/README.md

RoyalBlue1:
> Consider using a map instead of manually using strcmp on every combination.
> 
> Example:
> 
> ```c++
> int getReturnTypeEnumFromString(const char* name) {
> 	using SQReturnTypeEnum;
> 
> 	static std::map<std::string, SQReturnTypeEnum> sqEnumStrMap = {
> 		{"bool",	SqBoolean },
> 		{"float",	SqFloat },
> 		{"vector",	SqVector },
> 		{"int",		SqInteger },
> 		{"entity",	SqEntity },
> 		{"string",	SqString },
> 		{"array",	SqArrays },
> 		{"asset",	SqAsset },
> 		{"table",	SqTable }
> 	};
> 
> 	if (sqEnumStrMap.count(name)) {
> 		return sqEnumStrMap[name];
> 	} else {
> 		return SqStringOrNull;
> 	}
> }
> ```
> 
> Also, Northstar doesn't seem to use camelCase for function names... Let me guess, Java programmer? Doesn't matter a ton tho, I just hate Java lol
> 
> Also remember to run clang-format before making PRs so you can get that epic 2/2 checks passed victory royale!11! Easy commands for this at the end of: https://github.com/R2Northstar/NorthstarLauncher/blob/main/README.md

Im using the full enum name because otherwise the names are very similar to other names. I should rename them.

Also this pr fails clang because the refactor branch fails in general its not my code thats doing it.

NorthstarDLL/squirrel.cpp
```diff
@@ -379,4 +379,28 @@ ON_DLL_LOAD_RELIESON("server.dll", ServerSquirrel, ConCommand, [](HMODULE baseAd
...
+		{"asset", SqReturnAsset},
+		{"table", SqReturnTable}};
+
+	if (sqEnumStrMap.count(name))
+	{
+		return sqEnumStrMap[name];
+	}
+	else
+	{
+		return SqReturnDefault; // previous default value
```
BobTheBob9, line 404: call me stupid but where is this actually defined?

KittenPopo: i mean, the build succeeded, so... search for it?

RoyalBlue1: It is part of the enum I defined in the header. Previously it was just set in the constructor of the SQFuncRegistration struct without a named value

PR #208, opened 2022-06-23
Various minor naming convention fixes

KittenPopo:
I have noticed a few functions here and there that don't comform to Northstar's "established" `PascalCase()` function naming.

I renamed them because it makes the sadness go away.

### DIFF
--- a/NorthstarDedicatedTest/clientchathooks.cpp
+++ b/NorthstarDedicatedTest/clientchathooks.cpp
@@ -23,7 +23,7 @@ static void CHudChat__AddGameLineHook(void* self, const char* message, int inbox
 		return;
 	}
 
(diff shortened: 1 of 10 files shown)


barnabwhy:
iirc squirrel functions were both PascalCase cause they were named in correspondence with the functions in squirrel

barnabwhy:
also this undoes at least 1 pr

KittenPopo:
> also this undoes at least 1 pr

What? Which PR?

KittenPopo:
> iirc squirrel functions were both PascalCase cause they were named in correspondence with the functions in squirrel

They... weren't pascal case? Ig if they're corresponding to squirrel funcs maybe do `SQFn_func_name `or somethin'?

uniboi (approved):
image
less go

PR #209, opened 2022-06-24
Log all clientcommands if `ns_should_log_all_clientcommands` is set

GeckoEidechse:
This is the native version of https://github.com/R2Northstar/NorthstarMods/pull/391. I re-used the same convar name as in that PR.

Setting `ns_should_log_all_clientcommands` to `1` will enable the logging. By default the value is `0`, i.e. no logging of client commands.

Sample output:

```
[2022-06-25 00:41:32.602] [info] XXXXXXXXX with UID YYYYYYYY sent stringcommand CC_RespawnPlayer Pilot
```

together with a command blocked by exploitfixes:

```
...
[2022-06-25 00:40:01.157] [info] XXXXXXXXX with UID YYYYYYYY sent stringcommand emit 
[2022-06-25 00:40:01.157] [info] [SERVER SCRIPT] ############################
[2022-06-25 00:40:01.157] [info] [SERVER SCRIPT] CommandString: load_recent_checkpoint was not added via AddClientCommandCallback but is being called in CodeCallback_ClientCommand
[2022-06-25 00:40:01.157] [info] [SERVER SCRIPT] ############################
[2022-06-25 00:40:01.158] [warning] Blocked exploititive client command "emit".
...
```

When `ns_should_log_all_clientcommands` is set to `0`, the 
```
[2022-06-25 00:40:01.157] [info] XXXXXXXXX with UID YYYYYYYY sent stringcommand ZZZZZ
```
part will not be printed.

Convar setting and logging worked as expected in testing.

I'm 100% sure about the placements of the convar check and spdlog print as well as the `´new ConVar()` registration, so if someone could make sure their spots are ok, would be much appreciated <3

### DIFF
--- a/NorthstarDedicatedTest/serverauthentication.cpp
+++ b/NorthstarDedicatedTest/serverauthentication.cpp
@@ -486,7 +486,7 @@ char CGameClient__ExecuteStringCommandHook(void* self, uint32_t unknown, const c
 	// Only log clientcommands if the convar `ns_should_log_all_clientcommands` equals 1
 	if (Cvar_ns_should_log_all_clientcommands->GetBool())
 	{
-		spdlog::info("{} with UID {} sent stringcommand {}", (char*)self + 0x16, (char*)self + 0xF500, pCommandString);
+		spdlog::info("{} (UID: {}) executed command: \"{}\"", (char*)self + 0x16, (char*)self + 0xF500, pCommandString);
 	}
 
 	if (CVar_sv_quota_stringcmdspersecond->GetInt() != -1)
--- a/NorthstarDedicatedTest/serverauthentication.cpp
+++ b/NorthstarDedicatedTest/serverauthentication.cpp
@@ -484,7 +484,7 @@ CCommand__TokenizeType CCommand__Tokenize;
 char CGameClient__ExecuteStringCommandHook(void* self, uint32_t unknown, const char* pCommandString)
 {
 	// Only log clientcommands if the convar `ns_should_log_all_clientcommands` equals 1
-	if (Cvar_ns_should_log_all_clientcommands->GetInt() == 1)
+	if (Cvar_ns_should_log_all_clientcommands->GetBool())
 	{
 		spdlog::info("{} with UID {} sent stringcommand {}", (char*)self + 0x16, (char*)self + 0xF500, pCommandString);
 	}
--- a/NorthstarDedicatedTest/serverauthentication.cpp
+++ b/NorthstarDedicatedTest/serverauthentication.cpp
@@ -486,7 +486,7 @@ char CGameClient__ExecuteStringCommandHook(void* self, uint32_t unknown, const c
 	// Only log clientcommands if the convar `ns_should_log_all_clientcommands` equals 1
 	if (Cvar_ns_should_log_all_clientcommands->GetInt() == 1)
 	{
-		spdlog::info("{} sent stringcommand {}", (char*)self + 0x16, pCommandString);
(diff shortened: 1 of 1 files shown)


NorthstarDedicatedTest/serverauthentication.cpp
```diff
@@ -480,6 +483,12 @@ CCommand__TokenizeType CCommand__Tokenize;
 
 char CGameClient__ExecuteStringCommandHook(void* self, uint32_t unknown, const char* pCommandString)
 {
+	// Only log clientcommands if the convar `ns_should_log_all_clientcommands` equals 1
+	if (Cvar_ns_should_log_all_clientcommands->GetInt() == 1)
```
BobTheBob9, line 487: use GetBool here

BobTheBob9 (changes requested):

NorthstarDedicatedTest/serverauthentication.cpp
GeckoEidechse: Changed with ea1b04b10ecb6af2be3a1409d8af44a8074759f7

NorthstarDedicatedTest/serverauthentication.cpp
```diff
@@ -480,6 +483,12 @@ CCommand__TokenizeType CCommand__Tokenize;
 
 char CGameClient__ExecuteStringCommandHook(void* self, uint32_t unknown, const char* pCommandString)
 {
+	// Only log clientcommands if the convar `ns_should_log_all_clientcommands` equals 1
+	if (Cvar_ns_should_log_all_clientcommands->GetBool())
+	{
+		spdlog::info("{} with UID {} sent stringcommand {}", (char*)self + 0x16, (char*)self + 0xF500, pCommandString);
```
KittenPopo, line 489: Formatting is kinda odd, doesn't make it super clear where the command starts and ends (?)

I'd recommend something like:
```cpp
spdlog::info("{} (UID: {}) executed command: \"{}\"", (char*)self + 0x16, (char*)self + 0xF500, pCommandString);
```

KittenPopo, line 489: If the command contains a `\n`, the client can now write their own console lines (we both know who would figure this out and abuse the hell out of it).

Maybe some sanity checks?
```c
std::string commandStr = (char*)self + 0x16;
		
// Clean indenting/newline characters that would mess up console
for (char& c : commandStr) {
	switch (c) {
		case '\n':
		case '\r':
		case '\t':
			c = ' ';
	}
}
```

KittenPopo (changes requested):

NorthstarDedicatedTest/serverauthentication.cpp
GeckoEidechse: Changed it to suggested version in 034f1f0b63d0c8dd0040f53c558a17509629a52a

GeckoEidechse: Doesn't seem to be necessary. From testing, commands are already escaped properly.

For example, I put in, `test\nhello\0abc%soh no` and it printed as:

```
[2022-06-25 12:45:09.344] [info] XXXXXXXXXXX (UID: YYYYYYYYY) executed command: "test\nhello\0abc%soh no"
```

emma-miler (approved):

NorthstarDedicatedTest/serverauthentication.cpp
KittenPopo: Hm ok lol, I wonder about other escape characters... 🤔 I'll test and let yk if I find anything.

PR #211, opened 2022-06-26
Block developer squirrel functions

RoyalBlue1:
Some Command Line Functions are still allowed
To allow functions again use command line parameter `-allowsquirreldevfunctions`

### DIFF
--- a/NorthstarDedicatedTest/squirrel.cpp
+++ b/NorthstarDedicatedTest/squirrel.cpp
@@ -464,7 +464,7 @@ template <ScriptContext context> void ExecuteCodeCommand(const CCommand& args)
 
 SQRESULT SQ_DevFuncStub(void* sqvm)
 {
-	spdlog::info("Blocked execution of squirrel developer function for security reasons. To re-enable them use start parameter "
+	spdlog::warn("Blocked execution of squirrel developer function for security reasons. To re-enable them use start parameter "
 				 "-allowSquirrelDevFunctions.");
 	return SQRESULT_NULL;
 }
--- a/NorthstarDedicatedTest/squirrel.cpp
+++ b/NorthstarDedicatedTest/squirrel.cpp
@@ -6,6 +6,7 @@
 #include "concommand.h"
 #include "modmanager.h"
 #include <iostream>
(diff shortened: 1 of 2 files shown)


NorthstarDedicatedTest/squirrel.cpp
```diff
@@ -447,4 +460,28 @@ template <ScriptContext context> void ExecuteCodeCommand(const CCommand& args)
...
+
+template <ScriptContext context> int64_t RegisterSquirrelFuncHook(void* sqvm, SQFuncRegistration* funcReg, char unknown)
+{
+	static std::set<std::string> allowedDevFunctions = {
+		"Dev_CommandLineHasParm",
+		"Dev_CommandLineParmValue",
+		"Dev_CommandLineRemoveParm",
+	};
+
+	if ((funcReg->devLevel == 1) && (!CommandLine()->CheckParm("-allowSquirrelDevFunctions")) &&
```
pg9182, line 479: For future reference: this is case-insensitive.

pg9182 (approved):

NorthstarDedicatedTest/squirrel.cpp
```diff
@@ -447,4 +460,29 @@ template <ScriptContext context> void ExecuteCodeCommand(const CCommand& args)
 		g_UISquirrelManager->ExecuteCode(args.ArgS());
 	else if (context == ScriptContext::SERVER)
 		g_ServerSquirrelManager->ExecuteCode(args.ArgS());
+}
+
+SQRESULT SQ_DevFuncStub(void* sqvm)
+{
+	spdlog::info("Blocked execution of squirrel developer function for security reasons. To re-enable them use start parameter "
```
GeckoEidechse, line 467: Maybe warn instead. Shouldn't usually be hit and that makes hitting it (for one reason or another) more obvious.

```suggestion
	spdlog::warn("Blocked execution of squirrel developer function for security reasons. To re-enable them use start parameter "
```

pg9182 (approved):

GeckoEidechse (approved):
Confirmed working in testing.

PR #215, opened 2022-06-29
Add -safe to dedicated launch arguments

rolelessweapon:
This should provide better dedicated performance.

### DIFF
--- a/NorthstarDedicatedTest/dedicated.cpp
+++ b/NorthstarDedicatedTest/dedicated.cpp
@@ -265,6 +265,7 @@ void InitialiseDedicated(HMODULE engineAddress)
 	// make sure it still gets registered
 
 	// add cmdline args that are good for dedi
+	CommandLine()->AppendParm("+fatal_script_errors 0", 0);
 	CommandLine()->AppendParm("-safe", 0);
 	CommandLine()->AppendParm("-nomenuvid", 0);
 	CommandLine()->AppendParm("-nosound", 0);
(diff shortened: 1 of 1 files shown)


BobTheBob9:
-safe shouldn't really do anything since we have the dx proxy dlls to prevent rendering calls anyway, and disabling fatal script errors probably shouldn't be default behaviour

GeckoEidechse:
We definitely don't wanna ignore fatal script errors as while that might reduce crashes, they will mask bugs from us and as such will bite as in the long run ^^

NorthstarDedicatedTest/dedicated.cpp
```diff
@@ -265,6 +265,8 @@ void InitialiseDedicated(HMODULE engineAddress)
 	// make sure it still gets registered
 
 	// add cmdline args that are good for dedi
+	CommandLine()->AppendParm("+fatal_script_errors 0", 0);
+	CommandLine()->AppendParm("-safe", 0);
```
GeckoEidechse, line 269:
```suggestion
	CommandLine()->AppendParm("-safe", 0);
```
Let's change it to just `-safe` and we can give it a better look but as Bob said, it might not be necessary either to use.

GeckoEidechse (changes requested):

rolelessweapon:
Sure, but that only disables script errors on dedicated, where those probably aren't wanted. I think fatal script errors should absolutely remain on in debug / listen mode, so developers can test their scripts, but there's no benefit to having the server crash because someone found a weird way to break a chat parser.

GeckoEidechse:
What exactly does `-safe` actually do? ^^

GeckoEidechse:
> What exactly does `-safe` actually do? ^^

@rolelessweapon

BobTheBob9:
closing as this due to inactivity and changes don't seem like they should really do much in practice

PR #216, opened 2022-06-29
Add StringToAsset function to squirrel

RoyalBlue1:
Adds function to turn a String into an asset. This functionality is needed for more customizable ui for example custom pictures for Playlists. For example [NorthstarMods/#322](https://github.com/R2Northstar/NorthstarMods/pull/322)

It also adds a function to set better values for the return type enum of the Native function struct
I already have a PR about that for refactor branch, but I think we need it sooner than that PR is finished.
The better values for the return enum are needed to return assets without needing to cast the result after return.


ASpoonPlaysGames:
Wait can you not do this in squirrel?
`string thing = "hello/world"; asset thing2 = compilestring( "return $\"" + thing + "\"")`
something like this?

uniboi:
> Wait can you not do this in squirrel? `string thing = "hello/world"; asset thing2 = compilestring( "return $\"" + thing + "\"")` something like this?

[That works](https://github.com/R2Northstar/NorthstarMods/blob/33c45fd0b343b0b5ae24c4bfc996447faa131178/Northstar.Custom/mod/scripts/vscripts/sh_message_utils.gnut#L367-L370=) but recompiling that statement every time you need an asset from a string isn't very efficient

ASpoonPlaysGames:
That makes sense, fair enough

RoyalBlue1:
also you would have to cast the result of compilestring or set the file to untyped

NorthstarDedicatedTest/scriptUtility.cpp
```diff
@@ -0,0 +1,28 @@
+#include "pch.h"
```
BobTheBob9, line 1: capitalization of filename should be more inline with other files

NorthstarDedicatedTest/scriptUtility.cpp
```diff
@@ -0,0 +1,28 @@
...
+#include "scriptUtility.h"
+#include "squirrel.h"
+
+SQRESULT ServerSq_StringToAsset(void* sqvm)
```
BobTheBob9, line 5: should be SQ_ not Sq_ to be inline with other squirrel funcs
could also template these funcs rather than having dupe ones for client/server

BobTheBob9 (changes requested):

NorthstarDedicatedTest/scriptUtility.cpp
RoyalBlue1: ok but we should also make a folder for all script additions. They slowly get out of hand

NorthstarDedicatedTest/scriptUtility.cpp
RoyalBlue1: are you sure with other functions that have client and server versions we also had ServerSq and ClientSq as prefix
image
but im making it a template anyway so it shouldn't matter

NorthstarDedicatedTest/scriptutility.cpp
```diff
@@ -0,0 +1,29 @@
+#include "pch.h"
+#include "scriptUtility.h"
```
GeckoEidechse, line 2:
```suggestion
#include "scriptutility.h"
```

Should be lowercase

NorthstarDedicatedTest/dllmain.cpp
```diff
@@ -45,6 +45,7 @@
 #include <string.h>
 #include "version.h"
 #include "pch.h"
+#include "scriptUtility.h"
```
GeckoEidechse, line 48:
```suggestion
#include "scriptutility.h"
```
same

GeckoEidechse (approved):
Works in testing.

`sv_cheats 1`
`script print(StringToAsset("abc"))`

image

PR #217, opened 2022-07-01
Add Postload support for rpaks

ASpoonPlaysGames:
Allows rpaks to be loaded directly after another rpak, this prevents some issues where preload was happening too early, and the rpak ended up waiting for something to be loaded that never was, causing an infinite loop. (this allows rpaks with materials to be loaded properly because they need common.rpak to be loaded for all of the shadersets)


EladNLG (commented):
Code clean, good feature, mergeable by me :D

GeckoEidechse:
Any way I could test this to confirm bug fixed? Or is the issue more on the probabilistic side?^^

ASpoonPlaysGames:
> Any way I could test this to confirm bug fixed? Or is the issue more on the random side? ^^
Figured I'd mention this here, just for completion's sake.
Yeah you can test it, I have sent you a zip of 2 mods on discord, one of which won't work, as the previous rpak loading methods aren't sufficient for loading it correctly, the other uses the new Postload section of the rpak.json and should load properly.

(when the rpak doesn't load properly it just sorta makes the game wait infinitely)

GeckoEidechse:
Sample mods for references:

https://cdn.discordapp.com/attachments/993647955150114956/993648964006068365/217-Mods.zip

- `Mod.BrokenWithoutPR` will never work and should highlight the broken load
- `Mod.UsesPR` should work on both current main and PR but only on PR actually loads files. Logs should say `LoadPakAsync ./R2Northstar/mods\Mod.UsesPR\paks\thing.rpak 2` on PR version of Launcher.


The difference between the two mods in `paks/rpak.json`:

`Mod.UsesPR`

image

vs `Mod.BrokenWithoutPR`

image

GeckoEidechse (approved):
From testing, PR fixes addressed issue.

Just need someone to review code changes ^^

RoyalBlue1 (approved):

PR #222, opened 2022-07-21
Fix a few issues with Json implementation to Squirrel VM

RoyalBlue1:
Fixed Issues from #172 
> Fixed using Server newTable in Client Function
> Fixed Client newTable using wrong offset


GeckoEidechse:
2 line PR, should be fine based on the conversations I saw ^^

PR #223, opened 2022-07-21
Add Safe I/O

EladNLG:
# Safe I/O
This PR allows Squirrel isolated access to a limited set of files, by having *file names* defined in the mod.json file.

It also adds many safety precautions to prevent malicious use, even if a malicious actor gains complete access to the Squirrel VM, to minimize damage (e.g. a 50,000 character limit per file, a forced JSON file extension, and disallowing access to files not defined in the mod.json file.)

## Documentation

Squirrel Functions:
```js
void function NSSaveFile( string mod, string file, table data ) // writes the table as JSON to the specified file. File has to be defined in the mod.json of your mod for this to work.
table function NSLoadFile( string mod, string file ) // reads the specified JSON file and returns it as a table. File has to be defined in the mod.json of your mod for this to work.
```

To allow the squirrel VM to access a file, in the mod.json, add an array named `"SaveFiles"` and add to there all of the **names** of the files you wish to access. You **cannot** use folders or non-ASCII characters.

### Example:
```jsonc
{
    // mod.json stuff...
    "SaveFiles": [
        "saveFile1", // valid
        "saves/saveFile2", // invalid
        "..\\..\\eladNLG\\muteMod\\saveFile3", // also invalid
    ]
}
```

Then, use `NSSaveFile` and `NSLoadFile` to save and load JSON data. The data persists between restarts, so this can be useful for saving stuff you want the client/server to save **locally,** like how many kills the player has gained in your server.

### Example:

```js
global function Mod_Init
global function MutePlayer

global table mutedPlayersData

void function Mod_Init()
{
    mutedPlayersData = NSLoadFile( "EladNLG.MuteMod", "eladNLG_muteMod_mutedPlayers" )
    AddClientCommandCallback( "mute", MutePlayer )
}

void function MutePlayer( entity player, array<string> args )
{
    if (!PlayerAllowedToMute(player))
        return false;
    if (args.len() < 1)
        return true;
    entity playerToMute = GetPlayerByName(args[0])
    
    // mute the player
    if (!("playerList" in mutedPlayersData))
        mutedPlayersData.playerList <- {}
        
    mutedPlayersData.playerList[playerToMute.GetUID()] <- true
    
    // save the file
    NSSaveFile( "EladNLG.MuteMod", "eladNLG_muteMod_mutedPlayers", mutedPlayersData )
}
```

## Reasons

Using ConVars for saving stuff, *especially* settings, just doesn't work. 

It resets each time you boot up vanilla, the length of the ConVar contents is very limited, and it does not scale well AT ALL if you want to save data on multiple players (for example, a mod that mutes players. After muting a ~40 players, the ConVar will not save all of the muted players, and some will be discarded - and though the same problem would exist with this - this would only happen after ~4,000 players, and if we wish, we can raise the limit even higher.)

### DIFF
--- a/NorthstarDLL/dedicated.cpp
+++ b/NorthstarDLL/dedicated.cpp
@@ -126,7 +126,7 @@ ON_DLL_LOAD_DEDI_RELIESON("engine.dll", DedicatedServer, ServerPresence, (CModul
 {
 	spdlog::info("InitialiseDedicated");
 
-	AUTOHOOK_DISPATCH_MODULE("engine.dll")
+	AUTOHOOK_DISPATCH_MODULE(engine.dll)
 
 	// Host_Init
 	// prevent a particle init that relies on client dll
--- a/NorthstarDLL/logging.cpp
+++ b/NorthstarDLL/logging.cpp
@@ -252,7 +252,7 @@ void InitialiseLogging()
 	spdlog::default_logger()->set_pattern("[%H:%M:%S] [%l] %v");
 }
 
-ON_DLL_LOAD_CLIENT_RELIESON("engine.dll", EngineSpewFuncHooks, ConVar, (CModule module))
+ON_DLL_LOAD_RELIESON("engine.dll", EngineSpewFuncHooks, ConVar, (CModule module))
 {
 	AUTOHOOK_DISPATCH_MODULE(engine.dll)
 


GeckoEidechse:
Damn that was quick, haha o_O

I definitely want some feedback from some modders about what of this PR is needed. I know that a lot of modders have been asking for a form of read/write persistence for a while now, so getting there in a timely manner is definitely a goal.

One thing I'm wondering is whether access to multiple files is actually necessary, or if a single large JSON file is enough for modders' needs. Especially as in the case second case, we don't have to mess with parsing any filenames etc but simply base the JSON filename on the mod name and make it a file inside the mod directory. (I'd still require a flag in `mod.json` whether mod wants to enable writing to file like your version implicitly does ^^)

ASpoonPlaysGames:
> having one file per mod, or even worse - one file per northstar profile means that if that file is corrupted, everything in there is lost.
> 
> In addition, if the character limit is reached, you can do nothing about it in the case of 1 file per mod, or 1 file per profile.
> 
> This character limit cannot be removed since it's added so that Squirrel cannot take up the entire hard drive if a malicious actor gains access to the VM.

I think the current way of doing this is the best option, malicious actors can't create new files dynamically, and are limited in the size of the existing files. I don't think we can really get much better than that

EladNLG:
I do think allowing the character limit to be changed in the mod.json would be a good idea too

so instead of
```jsonc
{
    // mod.json stuff...
    "SaveFiles": [
        "saveFile1", // valid, defaults to 50k character limit
    ]
}
```

You could also do

```jsonc
{
    // mod.json stuff...
    "SaveFiles": [
        {
            "FileName": "saveFile1",
            "CharLimit": 200000
        }, // valid
        "saveFile2" // still valid, defaults to 50k character limit
    ]
}
```
in case you want a larger file. There would be no option for -1/no limit since that would basically ruin the security but it'll allow for mods to set a higher limit for files they know are going to take more space.

GeckoEidechse:
> I do think allowing the character limit to be changed in the mod.json would be a good idea too

That could also be applied to the single file case ;)

GeckoEidechse:
Also, one thing that came to mind is that we probably wanna offer the `ns_resetpersistence` equivalent for resetting mod data store that wipes the JSON files / resets them to certain state. ^^

EladNLG:
IMO, not needed. Just go into R2Northstar

> > I do think allowing the character limit to be changed in the mod.json would be a good idea too
> 
> That could also be applied to the single file case ;)

that was unrelated to my argument about multiple files :)

though I do think it's still a good idea since it leads to cleaner code overall :3

I don't think an `ns_resetpersistence` equivalent is needed, but it can be done if we wish to do so.

ASpoonPlaysGames:
We shouldn't need an `ns_resetpersistence` equivalent, as the files can just be deleted or modified directly if needed

EladNLG:
**VOTE HERE**

👍 if you want access to multiple files
💯 if you don't mind having only one file to work with

GeckoEidechse:
Considering that this now marked as ready, I would much appreciate if any modders can give this a try and play around with it, make some sample mods that uses or try to break it ;)

Instructions on how to download launcher PR binaries (so you don't have to compile from source) can be found in the wiki: 
https://r2northstar.gitbook.io/r2northstar-wiki/modding-and-development/development/testing#northstarlauncher

Code reviews are also welcome <3

taskinoz:
There was a discussion about allowing mods to read other mods files in the discord. Having it be an option would probably be good in the way the character limit is set up in case sensitive data is stored and another mod ships with a malicious mod or plugin.

ASpoonPlaysGames:
> There was a discussion about allowing mods to read other mods files in the discord. Having it be an option would probably be good in the way the character limit is set up in case sensitive data is stored and another mod ships with a malicious mod or plugin.

Since we dont really have a way to tell which mod a script comes from, I don't really see a way that we can stop a mod from reading another mod's saved data?

taskinoz:
> Since we dont really have a way to tell which mod a script comes from, I don't really see a way that we can stop a mod from reading another mod's saved data?

Thats a good point

EladNLG:
> > Since we dont really have a way to tell which mod a script comes from, I don't really see a way that we can stop a mod from reading another mod's saved data?
> 
> Thats a good point

In addition, dependency constants only makes figuring this out easier. Though, there is no way to know what other files have been registered by other mods, so unless you hard-coded the files in, you can't do anything buy brute force your way until you find a file that's available.

EladNLG:
Anyways, you can now override the default character limit.
```jsonc
{
    // mod.json stuff...
    "SaveFiles": [
        {
            "FileName": "saveFile1",
            "CharLimit": 200000 // overrides the char limit to 200k because saveFile1 contains a lot of data 8)
        }, // valid
        "saveFile2" // still valid, defaults to 50k character limit
    ]
}
```

GeckoEidechse:
> Anyways, you can now override the default character limit.

It might still make sense to have an upper bound hardcoded into launcher to avoid the case of a lazy modder just cranking the value to `999999999999999` and then getting the writing to disk equivalent of a memory leak ^^

Something like 1GB should be more than enough as upper limit and the Squirrel<->JSON conversion might die anyway at that size :P

EladNLG:
@GeckoEidechse upper bound is 2billion per file :)

EladNLG:
> @GeckoEidechse upper bound is 2billion per file :)

Update, we will be introducing a global limit and a per-mod limit to prevent malicious mods from filling up the hard drive.

GeckoEidechse:
> In addition, dependency constants only makes figuring this out easier. Though, there is no way to know what other files have been registered by other mods, so unless you hard-coded the files in, you can't do anything buy brute force your way until you find a file that's available.

Actually, could we do something where launcher checks whether two mods try to register the same file?

uniboi:
> Actually, could we do something where launcher checks whether two mods try to register the same file?

This shouldn't be an issue? The way I understood this is that every mod has it's own folder

EladNLG:
> > > Actually, could we do something where launcher checks whether two mods try to register the same file?
> > 
> > 
> > This shouldn't be an issue? The way I understood this is that every mod has it's own folder
> 
> Based on discussion above my assumption is that we cannot actually tell which mod currently running Squirrel code is coming from...

This wouldn't help mitigate that.

EladNLG:
> Based on discussion above my assumption is that we cannot actually tell which mod currently running Squirrel code is coming from...

I don't think this would help - and it would only make the experience using this worse, since now you have to make sure there are no conflicts between this mod's file names and the other mods file names, despite there being no conflict at all. 

If there was a way to detect which mod is currently running the code **in NorthstarLauncher** I *would* be able to prevent mods from writing into each other's files - but as of now nothing cannot be done.

uniboi:
You could keep track which mod is the last to register which script file and with stack infosvyouvcan figure out which mod is executing code

RoyalBlue1:
> If there was a way to detect which mod is currently running the code in NorthstarLauncher I would be able to prevent mods from writing into each other's files - but as of now nothing cannot be done.

we wanted to to something like that for better debugging anyway. we could make a map storing last overwrite for each file

pg9182:
For the record, out of all the IO-related ideas so far, I think I like this one best.

Also:
- We need to ensure we wrap filesystem operations with try/catch where necessary, especially the write ones.
- It may be better to flush output files on a separate thread.

EladNLG:
Hi, the merge conflicts have become a nightmare for me to solve, so as for now I'm closing this and rewriting the thing from scratch.

PR #224, opened 2022-07-22
Status command overhaul

rolelessweapon:
Makes the status command include UIDs and makes it functional.


NorthstarDedicatedTest/logging.cpp
```diff
@@ -407,6 +407,86 @@ bool CClientState_ProcessPrint_Hook(__int64 thisptr, __int64 msg)
...
+	}
+
+	if (hours > 0)
+	{
+		snprintf(string, sizeof(string), "%2i:%02i:%02i", hours, minutes, seconds);
+	}
+	else
+	{
+		snprintf(string, sizeof(string), "%02i:%02i", minutes, seconds);
+	}
```
GeckoEidechse, line 442: Having it consistently print `hh:mm:ss` would be preferred in terms of parsing over changing format in case the server has been online <1h

NorthstarDedicatedTest/logging.cpp
```diff
@@ -407,6 +407,86 @@ bool CClientState_ProcessPrint_Hook(__int64 thisptr, __int64 msg)
...
+
+typedef double(__fastcall* GetTimeConnectedType)(__int64 thisptr);
+GetTimeConnectedType GetTimeConnected;
+
+typedef char*(__fastcall* GetAddressType)(__int64 thisptr);
+GetAddressType GetAddress;
+
+const char* COM_FormatSeconds(int seconds)
+{
+	static char string[64];
```
GeckoEidechse, line 419: Would need @pg9182's opinion her on best practices and all but passing string pointer to function instead of allocating a `static` on heap would probably be preferred.

rolelessweapon: This function is from the CS:GO source leak.

NorthstarDedicatedTest/logging.cpp
GeckoEidechse: (Function was copied from leaked CS:GO source so I guess we're keeping it this way.)

GeckoEidechse:
Might wanna fix the failing format check at least ;)

rolelessweapon:
Fixed by changing FCVAR_CLIENTCMD_CAN_EXECUTE to FCVAR_GAMEDLL_FOR_REMOTE_CLIENTS. (Although I didn't know it existed at first in the project, sorry... >.<)

Alystrasz:
There's been some work on logging since last year, @rolelessweapon do you still want to merge this?

F1F7Y:
I'd be interested in this. Considering the age, do you want to resolve the conflicts or can i just redo this on my own branch and credit you in the PR @rolelessweapon ??

ASpoonPlaysGames:
stale, closing

PR #225, opened 2022-07-22
Several new server base classes

Mauler125:
Implemented:
* CServer
* CClient
* CNetChan
* A few smaller ones

Not all members align, however a lot do (structs have been taken from R5R).
CServer is aligned entirely
CClient is fairly well aligned (name, netchan, etc).
CNetChan also seems ok for the most part but towards the end i added padding as they did something (needs reversing), but you can obtain the addresses and signonstate, etc.

Example usage:
```cpp
CClient* pClient = g_pClient->GetClient(i);
spdlog::info("{}", pClient->GetServerName());
spdlog::info("{}", pClient->GetNetChan()->GetAddress());
```

I recommend replacing `GetPlayerByIndex()` with this entirely, and also fixup some of the fields like where OriginID is since i couldn't find it but i think you had some offsets to it, so you can add the members at this offset of the structure.
Persistence data should also be added still, but i left these parts as i don't know the exact sizes and how Northstar uses them.

I also added a few fixme comments.

### DIFF
--- a/NorthstarDedicatedTest/NorthstarDedicatedTest.vcxproj
+++ b/NorthstarDedicatedTest/NorthstarDedicatedTest.vcxproj
@@ -113,15 +113,20 @@
     <ClInclude Include="bits.h" />
     <ClInclude Include="buildainfile.h" />
     <ClInclude Include="chatcommand.h" />
+    <ClInclude Include="client.h" />
     <ClInclude Include="clientchathooks.h" />
     <ClInclude Include="debugoverlay.h" />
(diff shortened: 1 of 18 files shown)


F1F7Y:
> @Mauler125 is this worth keeping open? (given all the merge conflicts that have accrued)

This is pre refactor so i doubt any merge conflicts can be easily solved. I can confirm that most of this lines up, but as mentioned in the original comment, not all fields we use are in this pr.

ASpoonPlaysGames:
Feel free to reopen if you ever fix the conflicts and such

Issue #212, opened 2022-06-26
Figure out building Northstar on Linux


GeckoEidechse:
We use `msbuild`,  which is Windows only?

We should figure out how to compile on Linux to make developing on Linux easier.

Note that this means still creating a Windows executable. As Titanfall2 is a Windows only game there's no point in creating a Linux-native version of Northstar as that version would have to interface with Wine/Proton instead and hook functions there which require a complete rewrite just for Linux. Given that Northstar works fine running under Wine/Proton, there's no point in putting in the huge amount of effort required for working Linux native builds.

p0358:
* https://github.com/paleozogt/MSVCDocker
* https://github.com/madduci/docker-msvc-cpp
* https://github.com/mstorsjo/msvc-wine

Highly suggesting the last one (it defaults to VS2022 contrary to what readme says). Is the most painless since it downloads files automatically without requiring Vagrant Windows image to presetup.

The only other option is to use a different Linux-native compiler, which I don't really recommend due to possible ABI incompatibilities that will not necessarily be nice to debug for people trying this approach, though @pg9182 disagreed with my opinion on this before...

pg9182:
I am planning to open a series of PRs based on my own configs:
- #226
- Additional non-critical fixes for compiler warnings and header include cleanups.
- Reorganize third-party dependencies.
- Meson build script (cross-platform).
- CI builds on Linux.

The end goal will be to have everything in Northstar compiled with llvm-mingw-msvcrt (instead of msvc) using meson build scripts (instead of msbuild), with all CI being done on Linux.

Benefits:
- Simplify developing Northstar on Linux.
- Give us better compiler warnings from clang.
- Allow us to use modern compiler features.
- Allow us to use standard C++ features rather than MSVC extensions.
- Unify launcher builds.
- Allow the docker container to build ns components from source.
- Preserve ABI compatibility.
- Simplify the source layout.
- Simplify including third-party libraries.

Costs:
- Slightly larger launcher binaries.
- A few extra one-time steps for setting up a development environment on Windows (install LLVM toolchain, install Meson, generate vcxproj).

apophis42:
[...]

PR #226, opened 2022-07-23
Fix most clang/mingw issues

pg9182:
- Fix include case.
- Replace MSVC-specific align with standard alignas.
- Type fixes.
- Delete operator noexcept.
- A few other minor issues.
- clang-format everything.
- Use c++20 instead of c++17.
- Rewrite ERROR macro for launcher_wsock32_proxy.
- Use a plain ifstream for the audio.cpp wavStream.

Note: When compiling with clang, you'll need -municode.

Related to #212.

### DIFF
--- a/LauncherInjector/LauncherInjector.vcxproj
+++ b/LauncherInjector/LauncherInjector.vcxproj
@@ -56,7 +56,7 @@
       <SDLCheck>true</SDLCheck>
       <PreprocessorDefinitions>_DEBUG;_CONSOLE;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <ConformanceMode>true</ConformanceMode>
-      <LanguageStandard>stdcpp17</LanguageStandard>
+      <LanguageStandard>stdcpp20</LanguageStandard>
       <AdditionalOptions>/F8000000 %(AdditionalOptions)</AdditionalOptions>
     </ClCompile>
     <Link>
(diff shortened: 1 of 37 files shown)


pg9182:
This PR will be followed by a few more:
- Additional non-critical fixes for compiler warnings and header include cleanups.
- Reorganize dependencies.
- Meson build script (cross-platform).
- CI builds on Linux.

pg9182:
Also see https://clang.llvm.org/docs/MSVCCompatibility.html.

GeckoEidechse:
That's quite cool actually :D

I'm assuming current PR would require no modification in CI or from current development setups?

GeckoEidechse:
Also, cause this touches a lot of files, would it be possible to separate this up into further smaller PRs? ^^

I'm assuming stuff like 

> - Fix include case.
> - Type fixes.

etc. could all be separate PRs which is a bit of pain in the butt to do as the person submitting the PR but would allow for a lot faster review as well as more legible commit history IMO.

I might be alone on that stance though idk :P

pg9182:
> I'm assuming current PR would require no modification in CI or from current development setups?

Nope, this just fixes incorrect/non-standard c++.

> Also, cause this touches a lot of files, would it be possible to separate this up into further smaller PRs? ^^

Theoretically, but there's not much benefit to it; note that most of it is just from clang-format. I'm splitting the PRs by functionality and impact rather than by every tiny change to the code.

> etc. could all be separate PRs which is a bit of pain in the butt to do as the person submitting the PR but would allow for a lot faster review as well as more legible commit history IMO.

Yes, since I did most of this a while ago, and the stuff I did yesterday wasn't in any particular order. I actually think it's better this way since you can easily skip commits when, let's say, bisecting.

GeckoEidechse (approved):
Ran the CI made build in my client for a bit and noticed no new issues.

Also tested local compilation via msvc which still works as expected.

BobTheBob9:
this looks great, i'll merge if i can confirm it builds fine on linux

BobTheBob9 (approved):

Issue #207, opened 2022-06-21
Allow comments in `banlist.txt`


GeckoEidechse:
We should allow comments (and empty lines) in `banlist.txt`, basically just skipping the line with the comment when reading in UIDs. This way hosters can annotate in banlist why someone was banned or use it to enable/disable bans of a certain player by just turning that line into a comment.

Example
```python
12345678

# This person was banned for not being cool enough
23456789

# This person was also banned but then I decided to unban them again. Leaving UID as comment to reban if necessary
# 34567890

```

I used `#` as the comment marker here cause it's the common comment indicator in config files, bash scripts, and Python.
We can obviously also use something else like `//` or `;` (pls no)

GeckoEidechse:
Completed with #227

PR #227, opened 2022-07-24
Allow comments and newlines in `banlist.txt` 

ScureX:
added support for
- empty lines
- comments which are prefixed with `#`
in `banlist.txt` like it was suggested by @GeckoEidechse in #207 

Example:
possible banlist:
image

saves these UIDs:
image

Most of the stuff i did has a comment explaining my view on it but feel free to question it. :)

---

> also istg do NOT mention any commits i know i fucked up and forgor abt shitty changes i made a while back which i had to revert i hate it here 💀 

enjoy this cat :)) <3
image

### DIFF
--- a/NorthstarDedicatedTest/bansystem.cpp
+++ b/NorthstarDedicatedTest/bansystem.cpp
@@ -8,6 +8,7 @@
 #include "configurables.h"
 
 const char* BANLIST_PATH_SUFFIX = "/banlist.txt";
+const char BANLIST_COMMENT_CHAR = '#';
 
 ServerBanSystem* g_ServerBanSystem;
 
@@ -20,7 +21,29 @@ void ServerBanSystem::OpenBanlist()
 	{
 		std::string line;
 		while (std::getline(enabledModsStream, line))
+<<<<<<< Updated upstream
 			m_vBannedUids.push_back(strtoll(line.c_str(), nullptr, 10));
+=======
+		{
+			// ignore line if first char is #
+			if (line.front() == BANLIST_COMMENT_CHAR)
+				continue;
+
+			// remove tabs which shouldnt be there but maybe someone did the funny
+			line.erase(std::remove(line.begin(), line.end(), '\t'), line.end());
+			// remove spaces to allow for spaces before uids 
+			line.erase(std::remove(line.begin(), line.end(), ' '), line.end());
+
+			// check if line is empty to allow for newlines in the file
+			if (line == "")
+				continue;
+
+			// for inline comments like: 123123123 #banned for unfunny
+			std::string uid = line.substr(0, line.find(BANLIST_COMMENT_CHAR));
+
+			m_vBannedUids.push_back(strtoull(uid.c_str(), nullptr, 10));
+		}
+>>>>>>> Stashed changes
 
 		enabledModsStream.close();
 	}
@@ -38,10 +61,20 @@ void ServerBanSystem::ClearBanlist()
 	m_sBanlistStream.open(GetNorthstarPrefix() + "/banlist.txt", std::ofstream::out | std::ofstream::binary);
 }
 
+bool firstBanThisSession = true;
 void ServerBanSystem::BanUID(uint64_t uid)
 {
+	// if we do comments stuff etc theres a big chance that we fuck with the formatting and there wont be a newline at the end. if we then
+	// ban someone that uid will just be added directly after the last one, leading to invalid uids and confusion, so we just make a newline
+	// every time we ban someone for the first time
+	if (firstBanThisSession)
+	{
+		m_sBanlistStream << std::endl;
+		firstBanThisSession = false;
+	}
+	
 	m_vBannedUids.push_back(uid);
-	m_sBanlistStream << std::to_string(uid) << std::endl;
+	m_sBanlistStream << std::to_string(uid) << std::endl; 
 	spdlog::info("{} was banned", uid);
 }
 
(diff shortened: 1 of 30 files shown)


NorthstarDedicatedTest/bansystem.cpp
```diff
@@ -20,7 +21,25 @@ void ServerBanSystem::OpenBanlist()
 	{
 		std::string line;
 		while (std::getline(enabledModsStream, line))
-			m_vBannedUids.push_back(strtoull(line.c_str(), nullptr, 10));
+		{
+			// ignore line if first char is #
+			if (line.front() == BANLIST_COMMENT_CHAR)
```
AnActualEmerald, line 26: Maybe change this to
```cpp
if (line == ""  || line.front() == BANLIST_COMMENT_CHAR)
```

NorthstarDedicatedTest/bansystem.cpp
```diff
@@ -20,7 +21,25 @@ void ServerBanSystem::OpenBanlist()
...
+				continue;
+
+			// remove tabs which shouldnt be there but maybe someone did the funny
+			line.erase(std::remove(line.begin(), line.end(), '\t'), line.end());
+			// remove spaces to allow for spaces before uids
+			line.erase(std::remove(line.begin(), line.end(), ' '), line.end());
+
+			// check if line is empty to allow for newlines in the file
+			if (line == "")
```
AnActualEmerald, line 35: Remove this as it's checked for with the other check up top

AnActualEmerald (changes requested):
Everything looks good to me otherwise, I can try to test this in a couple hours when I get home

NorthstarDedicatedTest/bansystem.cpp
ScureX: the problem i have with your 2 changes is that this would allow lines that arent `""` but instead `" "` or `"\t"` or any combination of whitespaces really to pass through and cause errors

AnActualEmerald: Ah touche, I think in my head I was thinking of it as a `is_whitepace` check

AnActualEmerald (approved):

NorthstarDedicatedTest/bansystem.cpp
```diff
@@ -20,7 +21,25 @@ void ServerBanSystem::OpenBanlist()
 	{
 		std::string line;
 		while (std::getline(enabledModsStream, line))
-			m_vBannedUids.push_back(strtoull(line.c_str(), nullptr, 10));
+		{
+			// ignore line if first char is # or line is empty
+			if (line == ""  || line.front() == BANLIST_COMMENT_CHAR)
```
GeckoEidechse, line 26:
```suggestion
			if (line == "" || line.front() == BANLIST_COMMENT_CHAR)
```

This is where format check failed btw

GeckoEidechse (approved):
Comments worked in testing.

NorthstarDedicatedTest/bansystem.cpp
```diff
@@ -38,8 +57,18 @@ void ServerBanSystem::ClearBanlist()
 	m_sBanlistStream.open(GetNorthstarPrefix() + "/banlist.txt", std::ofstream::out | std::ofstream::binary);
 }
 
+bool firstBanThisSession = true;
 void ServerBanSystem::BanUID(uint64_t uid)
 {
+	// if we do comments stuff etc theres a big chance that we fuck with the formatting and there wont be a newline at the end. if we then
+	// ban someone that uid will just be added directly after the last one, leading to invalid uids and confusion, so we just make a newline
+	// every time we ban someone for the first time
+	if (firstBanThisSession)
```
BobTheBob9, line 66: should probably only do this if the last char isn't a newline

BobTheBob9 (changes requested):

NorthstarDedicatedTest/bansystem.cpp
ScureX: did it hopefully the most efficient way

NorthstarDedicatedTest/bansystem.cpp
```diff
@@ -40,6 +59,14 @@ void ServerBanSystem::ClearBanlist()
...
 void ServerBanSystem::BanUID(uint64_t uid)
 {
+	// checking if last char is \n to make sure uids arent getting fucked
+	std::ifstream fsBanlist(GetNorthstarPrefix() + "/banlist.txt");
+	std::string content((std::istreambuf_iterator<char>(fsBanlist)), (std::istreambuf_iterator<char>()));
+	fsBanlist.close();
+
+	if (content.back() != '\n')
+		m_sBanlistStream << std::endl;
+
```
GeckoEidechse, line 69: Soooo that didn't work in testing :/

GeckoEidechse (changes requested):

NorthstarDedicatedTest/bansystem.cpp
GeckoEidechse: Had like an commented out UID in file with no newline

```
# AAAAAA
```

and on ban it looked like
```
# AAAAAABBBBBB
```

GeckoEidechse: Nevermind, issue was on my end. Banlist only writes `\n` but my text editor expected `\r\n`, resulting in it not showing line breaks. We can talk about switching to `\r\n` but that should be a separate PR considering that current pre-PR implementation also only just wrote `\n`.

GeckoEidechse (approved):
Worked in testing.

NorthstarDedicatedTest/bansystem.cpp
ScureX: silly little windows 8 user

GeckoEidechse: :trollface:

BobTheBob9 (approved):

PR #228, opened 2022-07-24
Fixed `unban` not overwriting `banlist.txt`

### LINKED ISSUE #165: unban command not working
https://github.com/R2Northstar/NorthstarLauncher/blob/6c8112a6c368dd36d21fd94689e9682bc3b012a5/NorthstarDedicatedTest/bansystem.cpp#L56

Removed from list in memory, but not file.

ScureX:
This closes #165. 
In the current state it only works with #227!

When unbanning it comments out the uid and adds a comment with the local time when the person was unbanned. Gives a nice overview i think, but it only works with #227 since that adds the comment functionality.
image

---

### DIFF
--- a/NorthstarDedicatedTest/bansystem.cpp
+++ b/NorthstarDedicatedTest/bansystem.cpp
@@ -28,7 +28,7 @@ void ServerBanSystem::OpenBanlist()
 	}
 
 	// open write stream for banlist // dont do this to allow for all time access
-	//m_sBanlistStream.open(GetNorthstarPrefix() + "/banlist.txt", std::ofstream::out | std::ofstream::binary | std::ofstream::app);
+	// m_sBanlistStream.open(GetNorthstarPrefix() + "/banlist.txt", std::ofstream::out | std::ofstream::binary | std::ofstream::app);
 }
 
 void ServerBanSystem::ReloadBanlist()
--- a/NorthstarDedicatedTest/bansystem.cpp
+++ b/NorthstarDedicatedTest/bansystem.cpp
@@ -27,24 +27,8 @@ void ServerBanSystem::OpenBanlist()
 		enabledModsStream.close();
 	}
 
-	// open write stream for banlist // dont do this to allow for all time access
-	//m_sBanlistStream.open(GetNorthstarPrefix() + "/banlist.txt", std::ofstream::out | std::ofstream::binary | std::ofstream::app);
-}
-
-void ServerBanSystem::ReloadBanlist()
-{
-	std::ifstream fsBanlist(GetNorthstarPrefix() + "/banlist.txt");
-
-	if (!fsBanlist.fail())
-	{
-		std::string line;
-		// since we wanna use this as the reload func we need to clear the list
-		m_vBannedUids.clear();
-		while (std::getline(fsBanlist, line))
-			m_vBannedUids.push_back(strtoull(line.c_str(), nullptr, 10));
-
-		fsBanlist.close();
-	}
+	// open write stream for banlist
+	m_sBanlistStream.open(GetNorthstarPrefix() + "/banlist.txt", std::ofstream::out | std::ofstream::binary | std::ofstream::app);
 }
 
 void ServerBanSystem::ClearBanlist()
@@ -54,15 +38,12 @@ void ServerBanSystem::ClearBanlist()
 	// reopen the file, don't provide std::ofstream::app so it clears on open
 	m_sBanlistStream.close();
 	m_sBanlistStream.open(GetNorthstarPrefix() + "/banlist.txt", std::ofstream::out | std::ofstream::binary);
-	m_sBanlistStream.close();
 }
 
 void ServerBanSystem::BanUID(uint64_t uid)
 {
 	m_vBannedUids.push_back(uid);
-	m_sBanlistStream.open(GetNorthstarPrefix() + "/banlist.txt", std::ofstream::out | std::ofstream::binary);
 	m_sBanlistStream << std::to_string(uid) << std::endl;
-	m_sBanlistStream.close();
 	spdlog::info("{} was banned", uid);
 }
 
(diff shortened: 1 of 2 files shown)


ScureX:
Leaving this here for everyone that thinks this may be a computation problem:
image

considering our player counts, 200 lines taking <1ms and how rare it is to unban a user i think this is not a problem. if we suddenly get more players than vanilla we can think about improving this

GeckoEidechse:
I gotta say I really like the idea of reusing your previous PR to just comment out bans. ^^

Also, making it dependent on the previous PR is a smart way to get that PR merged faster :trollface:

GeckoEidechse:
Just to be sure, you tested, banning, unbanning, re-banning, and re-unbanning the same UID?

NorthstarDedicatedTest/bansystem.cpp
```diff
@@ -52,9 +54,75 @@ void ServerBanSystem::UnbanUID(uint64_t uid)
...
+
+				std::ostringstream unbanComment;
+
+				//{y}/{m}/{d} {h}:{m}
+				unbanComment << " # unban date: ";
+				unbanComment << now->tm_year + 1900 << "/"; // this lib is so fucking awful
+				unbanComment << std::setw(2) << std::setfill('0') << now->tm_mon + 1 << "/";
+				unbanComment << std::setw(2) << std::setfill('0') << now->tm_mday << " ";
+				unbanComment << std::setw(2) << std::setfill('0') << now->tm_hour << ":";
+				unbanComment << std::setw(2) << std::setfill('0') << now->tm_min;
```
GeckoEidechse, line 104: Why do you hate [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) :(

ScureX: ok first of all let me rant about how fukcing god awful and insulting this lib is they give you the fukcing yeaRS SINCE 1900 WHO THE FUCK DOES THAT???? now to the standard: we can switch this around however we want, replace the `/` with a `-`, i dont really care, i just thought this looked neat.

ScureX:
> I gotta say I really like the idea of reusing your previous PR to just comment out bans. ^^

spoke to wolf about it, i feel like this is a really nice thing so the uid isnt just "gone" but you have an overview of what you did and when

> Just to be sure, you tested, banning, unbanning, re-banning, and re-unbanning the same UID?

those 2 prs arent on the same branch so i only tested their independant quirks 💀 ill make a new branch where i merge the 2 and test it fully!

what im like 99.9% sure of if going to happen:
- banning:
>just adds the uid at the bottom of a file, probably with one line too many if you havent fucked with the banlist thru the editor and only used commands, but rather 1 line too much than 2 uids sticking together (had that happen before)

- unbanning:
> since i go thru the file in a loop and check every line for the uid that should be unbanned, even duplicate, active uids should be commented out, just in different places

- re-banning:
> re-banning will NOT comment out the unbanned uid! it wil just add the uid at the bottom again, which i honestly prefer. that way you have like a timeline of what happened when and you can recheck and see how often someone has been unbanned (like a "you have x chances till perm ban" thing maybe)

- re-unbanning
> re-unbanning will (because of the loop) once again just comment out every line that has the right (active) uid

@GeckoEidechse i will test this all again tho to be 100% sure and ping u here again :)

NorthstarDedicatedTest/bansystem.cpp
GeckoEidechse: At least replacing `/` with `-` would be preferred personally, it's also more consistent with the way we format date in logs and minidump ^^

GeckoEidechse:
> * re-banning:
> 
> 
> > re-banning will NOT comment out the unbanned uid! it wil just add the uid at the bottom again, which i honestly prefer. that way you have like a timeline of what happened when and you can recheck and see how often someone has been unbanned (like a "you have x chances till perm ban" thing maybe)

This is what I would expect and also prefere myself personally.

>  * re-unbanning
> 
> > re-unbanning will (because of the loop) once again just comment out every line that has the right (active) uid

This is what I'm looking for. I haven't gotten the chance to read fully through the code so I just want to make sure that I comments out the right UID and doesn't just somehow grab the earlier one and just add another `#` to the already commented out UID ^^

ScureX:
> This is what I'm looking for. I haven't gotten the chance to read fully through the code so I just want to make sure that I comments out the right UID and doesn't just somehow grab the earlier one and just add another `#` to the already commented out version of the UID ^^

```c++
std::string modLine = line; // copy the line into a free var that we can fuck with, line will be the original

// remove tabs which shouldnt be there but maybe someone did the funny
modLine.erase(std::remove(modLine.begin(), modLine.end(), '\t'), modLine.end());
// remove spaces to allow for spaces before uids
modLine.erase(std::remove(modLine.begin(), modLine.end(), ' '), modLine.end());

// ignore line if first char is # or empty line, just add it
if (line.front() == BANLIST_COMMENT_CHAR || modLine == "")
{
  banlistText.push_back(line);
  continue;
}
```

to spare you from reading through the whole file: this is the relevant part. it will do that for every line unless the first non-space/non-tab character is `#` to make sure its an active uid. then it just compares uids and comments out or not. that way it will comment out all the active uids while ignoring comments :P

ScureX:
alright so i fixed a small issue where unbanning a uid with inline comment wouldnt work because of whitespaces.

ive tested banning, unbanning, re.banning and re.unbanning without issues but id be glad if others could confirm!

@GeckoEidechse 
example of a full file with the 2 uids at the bottom following `banning, unbanning, re-banning and adding a comment, and re-unbanning`:
image

ScureX:
for testing purposes i merged the 2 pr branches into this branch https://github.com/ScureX/NorthstarLauncher/tree/commentsAndUnbanPRs so if anyone is doing a review or testing shit feel free to use it :)

ScureX:
alright so i made the banlist editable by closing it everywhere, it gets reloaded when a player joins. not really a need to reload it when a player gets banned since on join it get checked anyways

ScureX:
moved hot reload to #233

GeckoEidechse (approved):
Worked in testing by banning, unbanning, and re-banning a player a bunch of times with them (trying to) join in-between.

GeckoEidechse:
Bob said it's chill :D

PR #230, opened 2022-07-31
Allow Postload to be used with modded rpaks properly

### LINKED ISSUE #229: Post load rpak paths not relative to the folder they are in
This causes issues if you want to postload a modrpak after another modded rpak such as:
`{
	"Postload": {
		"betapulse_preload.rpak": "common.rpak",
		"betapulse.rpak": "betapulse_preload.rpak"
	}
}`

To do this currently you would need to do this:
`{
	"Postload": {
		"betapulse_preload.rpak": "common.rpak",
		"betapulse.rpak": "R2Northstar\mods\MODNAME\paks\betapulse_preload.rpak"
	}
}`

This isn't ideal as 'R2Northstar' might not always be the base folder, meaning the end user would have to adjust the path if they wanted it in a preset or something similar.

ASpoonPlaysGames:
Closes #229 
This allows Postload to use relative paths for the target rpak, meaning modded rpaks can be used as valid Postload targets.

This is now valid, and will cause the following rpak load order: `common.rpak -> betapulse_preload.rpak -> betapulse.rpak`
`"Postload": {
		"betapulse_preload.rpak": "common.rpak",
		"betapulse.rpak": "./betapulse_preload.rpak"
	}`

Importantly, `common.rpak` is still a valid Postload target, making this not a breaking change for rpak mods

`"betapulse.rpak": "betapulse_preload.rpak"` would also work in this example, however it would load the rpak if another mod loaded an rpak called `betapulse_preload.rpak` which could cause problems.


ASpoonPlaysGames:
Testing: (ignore example.rpak, thats just another rpak mod i had enabled when i took the screenshot)

unknown
unknown

Alystrasz:
Is this still an issue?
If yes, could you describe a way to reproduce it with current build + verify that this PR fixes it?

ASpoonPlaysGames:
> Is this still an issue?
> If yes, could you describe a way to reproduce it with current build + verify that this PR fixes it?

Kinda? I don't think anyone actually runs into it in practice, but being unable to load an rpak specifically after another modded rpak is annoying in theory.

For testing, see the above comments, nothing about that stuff has changed

PR #232, opened 2022-08-08
Add parsing region field from JSON server list and exposing it to Squirrel

GeckoEidechse:
mp_lobby0017

Adds capability to parse `region` field proposed in https://github.com/R2Northstar/NorthstarMasterServer/pull/84 and adds it to server object.

Further it exposes a function to Squirrel to read server region via `NSGetServerRegion` just like we do with other parts of the server object, e.g. `NSGetServerName`.

Also, like with other fields in the server object in the JSON file, a missing `region` key causes the server object to be skipped as "malformed".

As a sidenote, I implemented this by basically copying other existing code and adjusting it accordingly, so if you notice something odd, the chance is high that it was just a result of me copying code and less a distinct design decision :P

Depends on ~~https://github.com/R2Northstar/NorthstarMasterServer/pull/84~~ https://github.com/R2Northstar/Atlas/pull/12 (merged)

Dependent: https://github.com/R2Northstar/NorthstarMods/pull/479


GeckoEidechse:
Set to draft as the PR it depends on is not yet merged.

uniboi:
What's blocking this?

GeckoEidechse:
> What's blocking this?

...

> Depends on https://github.com/R2Northstar/NorthstarMasterServer/pull/84

GeckoEidechse:
Aight, PR is now brought back up-to-date with `main`. My issue rn in implementation is that I'm not sure how I implement optional JSON args, given that as per https://github.com/R2Northstar/Atlas/pull/12 server objects can have no region field.

NorthstarDLL/masterserver.cpp
```diff
@@ -231,9 +235,10 @@ void MasterServerManager::RequestServerList()
 					if (!serverObj.HasMember("id") || !serverObj["id"].IsString() || !serverObj.HasMember("name") ||
 						!serverObj["name"].IsString() || !serverObj.HasMember("description") || !serverObj["description"].IsString() ||
 						!serverObj.HasMember("map") || !serverObj["map"].IsString() || !serverObj.HasMember("playlist") ||
-						!serverObj["playlist"].IsString() || !serverObj.HasMember("playerCount") || !serverObj["playerCount"].IsNumber() ||
-						!serverObj.HasMember("maxPlayers") || !serverObj["maxPlayers"].IsNumber() || !serverObj.HasMember("hasPassword") ||
-						!serverObj["hasPassword"].IsBool() || !serverObj.HasMember("modInfo") || !serverObj["modInfo"].HasMember("Mods") ||
+						!serverObj["playlist"].IsString() || !serverObj.HasMember("region") || !serverObj["region"].IsString() ||
+						!serverObj.HasMember("playerCount") || !serverObj["playerCount"].IsNumber() || !serverObj.HasMember("maxPlayers") ||
+						!serverObj["maxPlayers"].IsNumber() || !serverObj.HasMember("hasPassword") || !serverObj["hasPassword"].IsBool() ||
+						!serverObj.HasMember("modInfo") || !serverObj["modInfo"].HasMember("Mods") ||
```
GeckoEidechse, line 241: Uh, someone double check that I didn't remove some vital check here :P

ASpoonPlaysGames: I hate this if statement so much

NorthstarDLL/masterserver.cpp
```diff
@@ -231,9 +235,10 @@ void MasterServerManager::RequestServerList()
 					if (!serverObj.HasMember("id") || !serverObj["id"].IsString() || !serverObj.HasMember("name") ||
 						!serverObj["name"].IsString() || !serverObj.HasMember("description") || !serverObj["description"].IsString() ||
 						!serverObj.HasMember("map") || !serverObj["map"].IsString() || !serverObj.HasMember("playlist") ||
-						!serverObj["playlist"].IsString() || !serverObj.HasMember("playerCount") || !serverObj["playerCount"].IsNumber() ||
-						!serverObj.HasMember("maxPlayers") || !serverObj["maxPlayers"].IsNumber() || !serverObj.HasMember("hasPassword") ||
-						!serverObj["hasPassword"].IsBool() || !serverObj.HasMember("modInfo") || !serverObj["modInfo"].HasMember("Mods") ||
+						!serverObj["playlist"].IsString() || !serverObj.HasMember("region") || !serverObj["region"].IsString() ||
```
GeckoEidechse, line 238: As region fields are optional per @pg9182's mention, this check should not be in here.

The thing is, I don't know how to properly parse the JSON later so that a non-existing JSON field results in a empty string for that line in the serverbrowser

NorthstarDLL/masterserver.cpp
ASpoonPlaysGames: Looks good to me, though

NorthstarDLL/masterserver.cpp
ASpoonPlaysGames: Either you can have the region field simply never not exist, and have Atlas send some sort of value that signifies null (empty string, etc.).

Or you can just do something like:
`region = serverObj.HasMember("region) ? serverObj["region"].GetString() : "";`
To set an empty string if `region` is not present

GeckoEidechse: oh yeah, that should work ^^

Imma check once pg updated Atlas in regards to no sending region for certain servers ^^

pg9182: @GeckoEidechse?

pg9182 (approved):
Code looks good, did some basic testing.

PR #235, opened 2022-08-14
Remove potentially clashing define

GeckoEidechse:
The correct type is `uint8_t`, not `uint8` which is non-standard and is Microsoft specific.

The `#define` caused an error for the RCON PR (#100).

The commit in this PR was cherry-picked as 4573d5e3141915f7066f182b47dac6a1d7d7bdcb from #100


GeckoEidechse:
Tagging @RoyalBlue1 as it was introduced in https://github.com/R2Northstar/NorthstarLauncher/pull/172

RoyalBlue1 (approved):

GeckoEidechse:
@RoyalBlue1 said merging is fine ^^

PR #237, opened 2022-08-17
Add log colours because they are pretty

ASpoonPlaysGames:
image

### DIFF
--- a/NorthstarDedicatedTest/logging.cpp
+++ b/NorthstarDedicatedTest/logging.cpp
@@ -234,6 +234,9 @@ void InitialiseLogging()
 	hExceptionFilter = AddVectoredExceptionHandler(TRUE, ExceptionFilter);
 
 	AllocConsole();
+	// these two lines are responsible for stuff to not show up in the console sometimes, from talking about it on discord
+	// apparently they were meant to make logging work when using -northstar, however from testing it seems that it doesnt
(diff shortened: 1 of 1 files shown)


GeckoEidechse:
Soo pwetty :eyes: 

I assume colour is only written to in console and we don't get any ANSI colour escape characters in logfile, right? ^^

ASpoonPlaysGames:
> I assume colour is only written to in console and we don't get any ANSI colour escape characters in logfile, right? ^^

image
Not 100% sure that this is the correct log, but all of the ones i have around the time i made the pr and uploaded that image are the same, no ANSI colour escape chars in sight as far as i can tell

GeckoEidechse:
So turns out colour printing doesn't seem to work on Windows 8.1 xD
It just prints in the usual white.

So I need someone on 10/11 to test to confirm this as working :,D

Alystrasz (approved):
Tested on Windows 10.
LGTM

GeckoEidechse (approved):
Colours confirmed working when testing on Windows 8.1 :3

GeckoEidechse:
Also confirmed again, no ANSI colour codes or similar in text logs ^^

GeckoEidechse:
Reverted in `v1.9.6` due to causing some issues with client freezes, mostly related to Viper. The exact cause is still unknown.

PR #244, opened 2022-08-24
Fix line endings (hopefully)

GeckoEidechse:
`CRLF` -> `LF`


GeckoEidechse:
Draft still cause I did it on Linux and `--renormalize` wouldn't work so I had to use a custom bash script instead.

GeckoEidechse:
Should be all good now

PR #251, opened 2022-08-24
Gitignore, ignore file name case sensitivity change

GeckoEidechse:
Ignore more commits in git blame


GeckoEidechse:
Set to draft as I'm anticipating more commits to ignore and I don't want to make a PR for each one

Issue #182, opened 2022-06-05
CURL revocation fails on Windows


faky1337:
I don't know when or why but on some client the revocation checks fail and the NorthstarLauncher is not able to establish an ecrypted connection using https. It seems related to schannel tho.

The error message (in the console window): `curl: (35) schannel: next InitializeSecurityContext failed: Unknown error (0x80092012)`.
Sometimes it seems not to throw any error message (but that could be the issue with a user not being capable of finding the console window? since it was not logged in any logfiles).
Also the error message gets translated to the clients OS language (german Windows => german error message).

adding `curl_easy_setopt(curl, CURLOPT_SSL_OPTIONS, CURLSSLOPT_NO_REVOKE);` on all requests seems  to fix the issue.

Not sure how much this would be related to security but I assume in case a certificate gets revoked the client would not be aware of this when disabling it in CURL?

- [ ] (Move CURL stuff to own function so we don't have to edit all occurances?)
- [ ] add switch like `-nosslrevocation` to optionally disable revocation thingy

p0358:
They should be asked out about their Windows version. I think curl uses WinSSL and thus by extension system's root trust store. And outdated/non-updated Windowses might have their stores outdated and incompatible with some newly issued certificates.

Although it should technically then be "expired", not "revoked"...

If this is indeed the issue, then one workaround could be shipping an up-to-date trusted CA list and using that instead

ASpoonPlaysGames:
Closing as stale

PR #254, opened 2022-08-25
add switch to disable revocation checks in libcurl

faky1337:
Will disable revocation checks if started with -nosslrevocation. Handy for some people who have issues with revocation checks in Windows Schannel. #182 
The cause for the issue is really not known and nobody was able to figure out what the issue might be.
This would be a workaround that people could add to their startup txt.

As far as I understand, the checks just try to look up if an issued certificate was revoked later. When removing the CA for northstar.tf I was still not able to connect so it will surely not disable the entire certificate checks.


pg9182:
Before we add this, can we also try `CURLSSLOPT_REVOKE_BEST_EFFORT`?

Alystrasz:
Would #182 be addressed by updating libcurl (#414)?

ASpoonPlaysGames:
Closing as stale (and conflicts)

PR #258, opened 2022-09-01
Change prints to C++ style

BobTheBob9:


GeckoEidechse (approved):
Works as expected in testing

PR #265, opened 2022-09-03
Refactor: Use in-file macros rather than global funcs for registering dll load callbacks/

GeckoEidechse:
This adds the refactor of dll load callbacks from the big refactor PR.

I'm putting it as draft for now cause it only works to some degree atm.

In particular it seems to still use the old load system in some places and me having very little launcher experience, I'm struggling to figure out the exact cause of the issue.

In a nutshell, I wanna get the refactored dll load callbacks working and merged as it makes up a decent chunk of file change and if merged makes the diff between refactor and main a lot more readable ^^

So any help in getting DLL load callbacks added properly would be appreciated <3

### DIFF
--- a/NorthstarDLL/audio.cpp
+++ b/NorthstarDLL/audio.cpp
@@ -1,4 +1,5 @@
 #include "pch.h"
+#include "hooks.h"
 #include "audio.h"
 #include "dedicated.h"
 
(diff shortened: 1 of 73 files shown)


GeckoEidechse:
Closed cause refactor got merged.

PR #266, opened 2022-09-04
Rename some members of MasterServerManager

GeckoEidechse:
Part of refactor. I basically took the changes from 037113d4b8032cd3715d879c2f4b08f66fa2be43 and added them onto main, excluding the debug build fix as that was related to a commit on refactor that is not yet present on main.


GeckoEidechse (commented):
Found no issues in testing.

PR #269, opened 2022-09-08
Add clang-format ignores for AUTOHOOK

GeckoEidechse:
The proper solution would be to have some way in clang-format config file that specifies that our `AUTOHOOK` formatting should be considered correct but that takes more investigation and simply ignoring formatting in AUTOHOOK macros is the lower hanging fruit.


GeckoEidechse:
Fun fact, vscode regex replace goes hard. I just used `AUTOHOOK(.+)\n(.+)\n` to find `AUTOHOOK` macros and then `// clang-format off\nAUTOHOOK$1\n$2\n// clang-format on\n` to add format ignores. Bit of manual clean-up at the end but turned the whole thing into a 5 minute job done ^^

GeckoEidechse:
Implemented in ba26be9bea801f219fa6340178fd6b13f02d574e

PR #270, opened 2022-09-11
Shut down server on crash and add server performance warning

rolelessweapon:
Shuts down server on crash, disconnecting all clients.
image

### DIFF
--- a/NorthstarDLL/logging.cpp
+++ b/NorthstarDLL/logging.cpp
@@ -209,6 +209,7 @@ long __stdcall ExceptionFilter(EXCEPTION_POINTERS* exceptionInfo)
 	}
 
 	logged = true;
+	reinterpret_cast<void (*)(char*)>((char*)GetModuleHandleA("engine.dll") + 0x118F00)((char*)GetModuleHandleA("engine.dll") + 0x12A53D40); // shut down server on crash.
 	return EXCEPTION_EXECUTE_HANDLER;
 }
 
--- a/NorthstarDLL/dedicated.cpp
+++ b/NorthstarDLL/dedicated.cpp
@@ -64,11 +64,21 @@ void RunServer(CDedicatedExports* dedicated)
 
 	// main loop
 	double frameTitle = 0;
+	float lastOverloadWarning = 0.f;
 	while (g_pEngine->m_nQuitting == EngineQuitState::QUIT_NOTQUITTING)
 	{
 		double frameStart = Plat_FloatTime();
 		g_pEngine->Frame();
+		float nextTickTime = Plat_FloatTime();
 
+		if (((nextTickTime - frameStart) > Cvar_base_tickinterval_mp->GetFloat()) && ((nextTickTime - lastOverloadWarning) >= 15.f))
+		{
+			spdlog::info(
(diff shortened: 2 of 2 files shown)


GeckoEidechse:
Just so I understand correctly, the purpose behind this PR is that the client gets the server shut-down signal so instead of freezing and timing out, they get disconnected properly from server, right? ^^

rolelessweapon:
Yes.

rolelessweapon:
Reference from source: https://pastebin.com/NAZGMGps

p0358:
Overall it's a good idea, but I'd suggest employing a few changes similar to what I did myself below in TFOR:
```cpp
if (Util::IsServerAlive())
{
	spdlog::error("Game server was running, shutting it down to kick out the connected clients...");
	MemoryAddress(enginedllBaseAddress).Offset(0x6398A8).PatchString("#DISCONNECT_SERVERCRASHED", sizeof("#DISCONNECT_SERVERSHUTTINGDOWN"));
	auto sv = static_cast<uint64_t>(enginedllBaseAddress + 0x27C5C80);
	auto CBaseServer_Shutdown = reinterpret_cast<void(__fastcall*)(uint64_t thisptr)>(enginedllBaseAddress + 0x20EDF0); // it checks sv.IsActive() inside anyways
	CBaseServer_Shutdown(sv);
}
```
* consider showing some log line before (and maybe after), as this is an operation that can potentially cause another crash itself (but that requires `sv.IsActive()` check which the func that's called checks itself before kicking clients; and it doesn't matter *that* much in Northstar to log it as a crash dump is already saved at that point)
* consider patching `#DISCONNECT_SERVERSHUTTINGDOWN` string to something like `#DISCONNECT_SERVERCRASHED` so that people know the server encountered an issue and don't assume the server owner is playing with them
* consider naming the function and maybe argument (assigning it to variable first), so it's more clear/readable what's being called

Not saying they're a must, but they're my personal recommendations

p0358:
But overall I tested this method and it works, so I approve 👍 definitely better than being stuck on resuming connection for several seconds if we can tell the clients it's not gonna happen, the risk of subsequent crash should hopefully be reasonably low here

NorthstarDLL/logging.cpp
```diff
@@ -209,6 +209,7 @@ long __stdcall ExceptionFilter(EXCEPTION_POINTERS* exceptionInfo)
 	}
 
 	logged = true;
+	reinterpret_cast<void (*)(char*)>((char*)GetModuleHandleA("engine.dll") + 0x118F00)((char*)GetModuleHandleA("engine.dll") + 0x12A53D40); // shut down server on crash.
```
BobTheBob9, line 212: would probably be best if this wasn't called on every server death, and was instead only called if server was active, and if it hadn't already been called, could probably just set a bool here to track whether it's been called already and not call it after that, in case we crash inside server shutdown and get stuck in a crash loop

BobTheBob9 (changes requested):

pg9182:
Looks like a good change overall, and I second @p0358's recommendations.

Does this handle the engine error crashes (it doesn't seem like it would, since those manifest as a messagebox which blocks the game loop)?

ASpoonPlaysGames:
@rolelessweapon is this something that is going to be worked on?

PR #272, opened 2022-09-15
Fix LSX error for good

p0358:
* fixes LSX error for good by adding `EAConnectionId` environment variable if it's missing, I confirmed this alone is enough to mitigate the LSX error, as I was lucky enough to have it occur 100% of the time when using NorthstarLauncher.exe :)
  * this method of checking environment variable existence is attested by Microsoft Docs, arguments to GetEnvironmentVariableA are nullptr and 0 as we aren't interested in the old value, these arguments are optional and thus can be set to these values
  * I didn't use `getenv` as the compiler would mutilate me for using "unsafe" function, and `getenv_s` requires taking output, so the method above is better
  * `(void)` cast is used on `_putenv` to explicitly discard the return value and avoid compiler warning
  * wsock and launcher proxy don't set this, as they're expected to be used with executable that's ran from within Origin and has always the environment variable pre-set, this only applies to NorthstarLauncher.exe
  * the actual value of `EAConnectionId` interestingly seems irrelevant, I could enter some random text in it and it would still mitigate the LSX error, it appears it just needs to exist and not be empty
* ~~set current working directory to the exe path -- this is normally expected to be set to that, but under certain circumstances where you create a shortcut to NorthstarLauncher.exe or otherwise use non-direct method of launching it (3rd party game launchers etc), the working directory might have been wrong, causing issues like the log file being attempted to be written in another location, due to use of relative paths~~ --> moved to #275 
* ~~catch an error when log file cannot be created instead of silently crashing, which can happen if the path isn't writable and confuse the user (after the MessageBox is shown, the game will continue launching, as non-writable log file isn't a fatal issue)~~  --> moved to #274

After merging can consider https://github.com/R2Northstar/Northstar/issues/220 obsolete. Of course I tested all these changes and they behave as expected.

### DIFF
--- a/NorthstarLauncher/main.cpp
+++ b/NorthstarLauncher/main.cpp
@@ -334,6 +334,8 @@ int main(int argc, char* argv[])
 		return 1;
 	}
 
+	SetCurrentDirectoryW(exePath);
+
 	bool noOriginStartup = false;
 	bool dedicated = false;
 	bool nostubs = false;
@@ -352,6 +354,11 @@ int main(int argc, char* argv[])
 	{
 		EnsureOriginStarted();
 	}
+
+	// ensure no LSX error
+	GetEnvironmentVariableA("EAConnectionId", nullptr, 0);
+	if (GetLastError() == ERROR_ENVVAR_NOT_FOUND)
+		(void)_putenv("EAConnectionId=Origin.OFR.50.0001452");
 
 	if (dedicated && !nostubs)
 	{
--- a/NorthstarDLL/logging.cpp
+++ b/NorthstarDLL/logging.cpp
@@ -19,14 +19,23 @@ void CreateLogFiles()
 	}
 	else
 	{
-		// todo: might be good to delete logs that are too old
-		time_t time = std::time(nullptr);
-		tm currentTime = *std::localtime(&time);
-		std::stringstream stream;
+		try
+		{
+			// todo: might be good to delete logs that are too old
+			time_t time = std::time(nullptr);
+			tm currentTime = *std::localtime(&time);
+			std::stringstream stream;
 
-		stream << std::put_time(&currentTime, (GetNorthstarPrefix() + "/logs/nslog%Y-%m-%d %H-%M-%S.txt").c_str());
-		spdlog::default_logger()->sinks().push_back(std::make_shared<spdlog::sinks::basic_file_sink_mt>(stream.str(), false));
-		spdlog::flush_on(spdlog::level::info);
+			stream << std::put_time(&currentTime, (GetNorthstarPrefix() + "/logs/nslog%Y-%m-%d %H-%M-%S.txt").c_str());
+			spdlog::default_logger()->sinks().push_back(std::make_shared<spdlog::sinks::basic_file_sink_mt>(stream.str(), false));
+			spdlog::flush_on(spdlog::level::info);
+		}
+		catch (...)
+		{
+			spdlog::error("Failed creating log file");
+			MessageBoxA(
+				0, "Failed creating log file! Make sure the profile directory is writable.", "Northstar Warning", MB_ICONWARNING | MB_OK);
+		}
 	}
 }
 
--- a/loader_launcher_proxy/dllmain.cpp
+++ b/loader_launcher_proxy/dllmain.cpp
@@ -128,6 +128,8 @@ extern "C" __declspec(dllexport) int LauncherMain(HINSTANCE hInstance, HINSTANCE
 			return 1;
 		}
 
+		SetCurrentDirectoryW(exePath);
+
 		bool loadNorthstar = ShouldLoadNorthstar();
 
 		if (loadNorthstar)
--- a/loader_wsock32_proxy/dllmain.cpp
+++ b/loader_wsock32_proxy/dllmain.cpp
@@ -39,6 +39,8 @@ BOOL WINAPI DllMain(HINSTANCE hInst, DWORD reason, LPVOID)
 			return true;
 		}
 
+		SetCurrentDirectoryW(exePath);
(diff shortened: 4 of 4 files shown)


p0358:
Uh, wait, I noticed some issues with the first change, don't merge it yet xd Perhaps I should have split them into two PRs ergh

p0358:
NVM the fix is all good, the issue I had was on my PC only and was related to outdated DNS xD Free to merge

NorthstarLauncher/main.cpp
```diff
@@ -353,6 +355,11 @@ int main(int argc, char* argv[])
 		EnsureOriginStarted();
 	}
 
+	// ensure no LSX error
+	GetEnvironmentVariableA("EAConnectionId", nullptr, 0);
+	if (GetLastError() == ERROR_ENVVAR_NOT_FOUND)
+		(void)_putenv("EAConnectionId=Origin.OFR.50.0001452");
```
GeckoEidechse, line 361: Can you give me like an ELI5 why this fixes LSX errors? ^^"

GeckoEidechse (commented):
3 separate PRs would have been nicer but I think I can deal with it :P

x3Karma:
No more tickets on Discord!!!! (real)

ASpoonPlaysGames (approved):
Would be very nice to stop the LSX error completely, LGTM

ASpoonPlaysGames:
Also would like to note that the amount of detail given in this PR is great :)

GeckoEidechse (changes requested):
So testing this by doing the same I can do to force an LSX error on my system by launching Northstar with Origin closed, I simply get stuck on the following.

screenshot

p0358:
Well, definitely not due to my changes, as the code for `LSX: connect()` is called before my changes. Does this stuck always happen or only randomly now? Did it eventually get unstuck when Origin finished opening or not?

The socket connection being stuck on LSX port might be an Origin issue and require an unrelated refactor to the waiting function... As it stands, if it turns out to be buggy like here, one could consider using it only if EADesktop.exe is running and OriginClientService.exe is not, as according to the comments in code the latter running is enough to launch the game successfully...

GeckoEidechse:
> Well, definitely not due to my changes, as the code for `LSX: connect()` is called before my changes. Does this stuck always happen or only randomly now? Did it eventually get unstuck when Origin finished opening or not?

Seems to be consistent. `LSX: connect()` is from this PR AFAIK #263

p0358:
Can you insert some extra log lines around those parts?

https://github.com/R2Northstar/NorthstarLauncher/blob/ebb4bd4ef099d77b0bf4dcbf63b0862fb58c29e8/NorthstarLauncher/main.cpp#L353-L361

For example:
```cpp
	if (!noOriginStartup && !dedicated)
	{
		EnsureOriginStarted();
	}
	std::cout << "Before GetEnvironmentVariableA" << std::endl;
	// ensure no LSX error
	GetEnvironmentVariableA("EAConnectionId", nullptr, 0);
	std::cout << "Before GetLastError" << std::endl;
	if (GetLastError() == ERROR_ENVVAR_NOT_FOUND) {
		std::cout << "Before _putenv" << std::endl;
		(void)_putenv("EAConnectionId=Origin.OFR.50.0001452");
		std::cout << "After everything" << std::endl;
}
```

p0358:
Then if `"Before GetEnvironmentVariableA"` isn't printed, it would mean it's stuck on connecting and that it 100% wouldn't be related to my changes. And if that was the case, you'd need to temporarily comment out `AwaitOriginStartup();` and then compare if my changes help against LSX error, nothing that `AwaitOriginStartup` in such case would need a refactor due to unrelated hang issue on connect.

In either case, `AwaitOriginStartup` has insufficient logging, you don't know if it's stuck on connect, on awaiting data, or finished successfully...

GeckoEidechse:
> Then if `"Before GetEnvironmentVariableA"` isn't printed, it would mean it's stuck on connecting and that it 100% wouldn't be related to my changes.

Yup, it doesn't get printed.

> And if that was the case, you'd need to temporarily comment out `AwaitOriginStartup();` and then compare if my changes help against LSX error

Commenting out `AwaitOriginStartup();` I'm back to LSX error again (:

screenshot

ASpoonPlaysGames:
Forgotten PR? What needs to be done?

GeckoEidechse:
> Forgotten PR? What needs to be done?

Didn't fix the issue it was trying to last time I checked (https://github.com/R2Northstar/NorthstarLauncher/pull/272#pullrequestreview-1111611458) ^^

p0358:
> Didn't fix the issue it was trying to fix last time I checked

Didn't fix for you, it fixed another case though for potentially different LSX error, but you didn't merge expecting some panacea for everything xd

If you decide to merge it, I'd need to resolve the current conflict by re-creating the commit and ditching the other changes that were already merged with different PRs I guess

GeckoEidechse:
> Didn't fix for you, it fixed another case though for potentially different LSX error, but you didn't merge expecting some panacea for everything xd

I mean I thought it was supposed to fix LSX for good :>

> Remember yours happened due to launching the game too early before Origin was ready and re-launching it later fixed it.

Just to be clear. It got stuck when launching with completely closed. To my knowledge that's the most common reason for LSX error. Northstar is supposed to boot Origin, wait for it to init and then LSX auth, but we never got that to work based on testing I did.

Anway, feel free to resolve merge conflicts and we can give it another shot ^^

ASpoonPlaysGames:
Closing as stale

GeckoEidechse:
Re-opening as I might look at this myself again. Although I will most likely cherry-pick and make new PR. So opening it again primarily so I don't miss it ^^

PR #274, opened 2022-09-16
Display error if logs path isn't writeable

p0358:
catch an error when log file cannot be created instead of silently crashing, which can happen if the path isn't writable and confuse the user (after the MessageBox is shown, the game will continue launching, as non-writable log file isn't a fatal issue)

[ separated from #272 ]


GeckoEidechse (approved):
Confirmed working in testing

screenshot

BobTheBob9 (commented):
looks great

PR #275, opened 2022-09-16
Set current directory properly

p0358:
set current working directory to the exe path -- this is normally expected to be set to that, but under certain circumstances where you create a shortcut to NorthstarLauncher.exe or otherwise use non-direct method of launching it (3rd party game launchers etc), the working directory might have been wrong, causing issues like the log file being attempted to be written in another location, due to use of relative paths

[ separated from #272 ]

### DIFF
--- a/NorthstarLauncher/main.cpp
+++ b/NorthstarLauncher/main.cpp
@@ -334,6 +334,8 @@ int main(int argc, char* argv[])
 		return 1;
 	}
 
+	SetCurrentDirectoryW(exePath);
+
 	bool noOriginStartup = false;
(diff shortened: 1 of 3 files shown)


GeckoEidechse (approved):
Confirmed working in testing.

I used my own tool [FlightCore](https://github.com/GeckoEidechse/FlightCore/) but removed the part where it `cd`s into the Titanfall2 directory. Without this PR Northstar launches in current dir, with, it launches in Titanfall2 dir.

BobTheBob9 (commented):
looks fine

PR #278, opened 2022-09-24
Replace XInput library with updated one

emma-miler:
This code adds better compatibility with newer controllers and different controller types

Also adds ASLR to the loaded library


pg9182:
This breaks docker since it tries to load xinput and hangs, presumably due to the messagebox.

NorthstarDLL/hooks.cpp
```diff
@@ -210,14 +210,29 @@ void CallAllPendingDLLLoadCallbacks()
...
-		CallLoadLibraryACallbacks(lpLibFileName, moduleAddress);
+		HMODULE moduleAddress = LoadLibraryExAOriginal("XInput9_1_0.dll", hFile, dwFlags);
+		if (moduleAddress)
+		{
+			CallLoadLibraryACallbacks(lpLibFileName, moduleAddress);
+		}
+		else
+		{
+			MessageBoxA(0, "Could not find XInput9_1_0.dll", "Northstar", MB_ICONERROR);
+			exit(-1);
```
pg9182, line 223: Uhh, this should be 1, not -1, but whatever.

PR #279, opened 2022-09-29
Allow rpaks to use starpaks that are in the mod's folder

ASpoonPlaysGames:
Before the starpaks had to be inside the r2 directory somewhere, this PR adds a field to the rpak.json that registers starpak paths as being relative to the mod/paks directory instead.

I'm making this PR now, however I am going to try and read the starpak paths from the rpak header instead of using the json here, so yeah


ASpoonPlaysGames:
Did the reading from the rpak file thing, as far as i'm concerned this is complete and ready

IJARika:
works as intended, tested myself.  see clip for example

https://streamable.com/ug1kdw

NorthstarDLL/rpakfilesystem.cpp
```diff
@@ -283,6 +283,48 @@ void* ReadFullFileFromDiskHook(const char* requestedPath, void* a2)
...
+
+				bool found = false;
+
+				for (size_t hash : mod.StarpakPaths)
+				{
+					if (hash == hashed)
+					{
+						starpakPath = mod.ModDirectory.string() + "\\paks\\" + starpakPath;
+						allocatedNewPath = new char[starpakPath.size() + 1];
+						strncpy(allocatedNewPath, starpakPath.c_str(), starpakPath.size());
```
IcePixelx, line 311: strncpy_s

NorthstarDLL/modmanager.cpp
```diff
@@ -426,6 +426,42 @@ void ModManager::LoadMods()
...
+					std::string str = "";
+					for (uint16_t i = 0; i < starpaksSize; i++)
+					{
+						if (buf[i] != 0x00)
+						{
+							str += buf[i];
+						}
+						else
+						{
+							if (str != "")
```
IcePixelx, line 456: !str.empty()

IcePixelx (changes requested):

NorthstarDLL/modmanager.cpp
ASpoonPlaysGames: Out of curiosity, what is the difference between the two?

pg9182: One uses compare (a branch), one checks the size field directly (a load). Minor difference, but it can add up.

NorthstarDLL/rpakfilesystem.cpp
pg9182: I would actually simply make allocatedNewPath a c++ string, then call c_str when passing it to stuff within the function.

pg9182:
docker: `ghcr.io/pg9182/northstar-dedicated-ci:1-tf2.0.11.0-ns1.9.7-ci.59de24a5.12.3197971268-nslauncher.6e71dcdc.3197262890`

pg9182 (approved):
LGTM. Code is okay, and it doesn't break in wine.

NorthstarDLL/rpakfilesystem.cpp
```diff
@@ -258,28 +258,62 @@ void*, __fastcall, (const char* pPath, void* pCallback))
...
+		}
+	LOG_STARPAK:
+		spdlog::info("LoadStreamPak: {}", filename.string());
+	}
 
 	void* ret = ReadFileAsync(pPath, pCallback);
-	if (allocatedNewPath)
-		delete[] allocatedNewPath;
 
 	return ret;
```
BobTheBob9, line 318: we should just return ReadFileAsync if we dont need to do anything between ReadFileAsync and return

NorthstarDLL/rpakfilesystem.cpp
```diff
@@ -258,28 +258,62 @@ void*, __fastcall, (const char* pPath, void* pCallback))
...
+			// hash the starpakPath to compare with stored entries
+			size_t hashed = STR_HASH(starpakPath);
+
+			// loop through all loaded mods
+			for (Mod& mod : g_pModManager->m_LoadedMods)
+			{
+				// ignore non-loaded mods
+				if (!mod.m_bEnabled)
+					continue;
+				// loop through the stored starpak paths
```
BobTheBob9, line 298: should probably leave a newline between continue and the comment here

NorthstarDLL/rpakfilesystem.cpp
```diff
@@ -258,28 +258,62 @@ void*, __fastcall, (const char* pPath, void* pCallback))
...
+	else if (path.extension() == ".starpak")
+	{
+		// code for this is mostly stolen from above
+
+		// unfortunately I can't find a way to get the rpak that is causing this function call, so I have to
+		// store them on mod init and then compare the current path with the stored paths
+
+		// game adds r2\ to every path, so assume that a starpak path that begins with r2\paks\ is a vanilla one
+		// modded starpaks will be in the mod's paks folder
+		if (path.string().find("r2\\paks\\") != 0 && path.string().find("r2/paks/") != 0)
```
BobTheBob9, line 285: i feel like there has to be a better way to do this check than to stringify the path and do string compares on it, like you could check the directory of the file directly or smth

BobTheBob9 (changes requested):

NorthstarDLL/rpakfilesystem.cpp
```diff
@@ -258,28 +258,62 @@ void*, __fastcall, (const char* pPath, void* pCallback))
...
+				// loop through the stored starpak paths
+				for (size_t hash : mod.StarpakPaths)
+				{
+					if (hash == hashed)
+					{
+						// construct new path
+						allocatedNewPath = (mod.m_ModDirectory / "paks" / starpakPath).string();
+						// set path to the new path
+						pPath = allocatedNewPath.c_str();
+						goto LOG_STARPAK;
```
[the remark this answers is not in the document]
ASpoonPlaysGames: Fairly certain I am, the loop through all mods as well as the loop through starpak paths in each mod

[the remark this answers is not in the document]
BobTheBob9: oh i fully misread this nvm, would recommend adding a newline before LOG_STARPAK: tho

NorthstarDLL/rpakfilesystem.cpp
ASpoonPlaysGames: decided to use the directory iterator and just check the second directory (first one is always r2, gets added by engine)

BobTheBob9 (approved):

NorthstarDLL/rpakfilesystem.cpp
```diff
@@ -258,30 +258,64 @@ void*, __fastcall, (const char* pPath, void* pCallback))
...
+	else if (path.extension() == ".starpak")
+	{
+		// code for this is mostly stolen from above
+
+		// unfortunately I can't find a way to get the rpak that is causing this function call, so I have to
+		// store them on mod init and then compare the current path with the stored paths
+
+		// game adds r2\ to every path, so assume that a starpak path that begins with r2\paks\ is a vanilla one
+		// modded starpaks will be in the mod's paks folder
+		if (*++path.begin() != "paks")
```
BobTheBob9, line 285: this is a nice trick but not super readable, would be better to use something more immediately readable here

BobTheBob9 (changes requested):

BobTheBob9 (approved):

PR #281, opened 2022-09-30
Add Named Pipe Client with Squirrle Integration

Neoministein:
This pull request adds a Named Pipe Client into the Northstar Launcher which allows it to connect to a Named Pipe Server.

This feature is currently being used for the [R2Northstar-TelemetrySystem](https://github.com/Neoministein/R2Northstar-TelemetrySystem) project to get a high volume of data out of Northstar to generate heatmaps and live web-based minimaps.

### Current Implementation details

Configuration:
```
The named pipe server is turned off by default but can be enabled with a startup argument:
+ns_enable_named_pipe 1
```

Open Named Pipe Client:
```
1. Squirrel calls NSOpenNamedPipe(string openingMessage, string closingMessageIfOpen)
2. The Launcher tries to connect to a named pipe at "\\\\.\\pipe\\GameDataPipe"
3. On connection the Launcher expects a byte array of 512. If it is smaller it needs to be null-terminated.
4. The byte array contains the name of a designated pipe which will be used for all further communication.

This is done so only one NamedPipeServer is needed for x amount of Northstar instances.
```

Send a message:
```
1. Squirrel calls NSSendToNamedPipe(string textToSend)
2. If connected and enabled it sends the specified string through the named pipe.
3. There is no specified length so the Named Pipe Server would need to look out for null-termination for when the message ends.
```

Close Named Pipe Client:
```
1. Squirrel calls NSCloseNamedPipe(string closingMessage)
2. The connection to the named pipe is terminated.
```

Developed by [@Aragami-delp], [@Neoministein]

### DIFF
--- a/NorthstarDLL/namedpipeclient.cpp
+++ b/NorthstarDLL/namedpipeclient.cpp
@@ -27,7 +27,7 @@ HANDLE GetNewPipeInstance()
 	HANDLE thisMatchPipe = INVALID_HANDLE_VALUE;
 	if (isGeneralConnected)
 	{
-		TCHAR chBuf[BUFF_SIZE]; // TODO: Length of general pipe ids + terminating zero
+		TCHAR chBuf[BUFF_SIZE]; // Server message should either be BUFF_SIZE or Length of general pipe ids + terminating zero
 		DWORD cbRead;
 		if (ReadFile(
 				generalPipe, // pipe handle
--- a/NorthstarDLL/namedpipeclient.cpp
+++ b/NorthstarDLL/namedpipeclient.cpp
@@ -48,12 +48,12 @@ HANDLE GetNewPipeInstance()
 
 void SendMessageToPipe(string message)
 {
-	bool success = false;
+	//bool success = false;
 	DWORD read;
-	do
-	{
-		success = WriteFile(hPipe, message.c_str(), message.length(), &read, nullptr);
-	} while (!success);
+	/*do
+	{*/
+		/*success = */WriteFile(hPipe, message.c_str(), message.length(), &read, nullptr);
+	/*} while (!success);*/
 }
 
 SQRESULT SQ_SendToNamedPipe(void* sqvm)
--- a/NorthstarDLL/namedpipeclient.cpp
+++ b/NorthstarDLL/namedpipeclient.cpp
@@ -1,0 +1,142 @@
+#pragma once
+#include "pch.h"
+#include "namedpipeclient.h"
+#include "squirrel.h" // Squirrel
+
+// Named Pipe Stuff
+#include <windows.h>
+#include <conio.h>
+#include <tchar.h>
+#include <string>
+#include <atlstr.h>
+
+using namespace std;
+
+#define GENERAL_PIPE_NAME TEXT("\\\\.\\pipe\\GameDataPipe")
+#define BUFF_SIZE 512
+
+HANDLE hPipe;
+bool isConnected = false;
+
+ConVar* Cvar_ns_enable_named_pipe;
+
+HANDLE GetNewPipeInstance()
+{
+	HANDLE generalPipe = CreateFile(GENERAL_PIPE_NAME, GENERIC_READ, 0, nullptr, OPEN_EXISTING, 0, nullptr);
+	bool isGeneralConnected = generalPipe != INVALID_HANDLE_VALUE;
+	HANDLE thisMatchPipe = INVALID_HANDLE_VALUE;
+	if (isGeneralConnected)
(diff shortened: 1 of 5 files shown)


NorthstarDLL/namedpipeclient.cpp
```diff
@@ -0,0 +1,140 @@
...
+			{
+				spdlog::info("New Pipe connection");
+				thisMatchPipe = CreateFile(chBuf, GENERIC_WRITE, 0, nullptr, OPEN_EXISTING, 0, nullptr);
+			} while (thisMatchPipe == INVALID_HANDLE_VALUE);
+		}
+	}
+	return thisMatchPipe;
+}
+
+void SendMessageToPipe(string message)
```
Mauler125, line 49: Pass parameter "message" by const reference instead

NorthstarDLL/namedpipeclient.cpp
```diff
@@ -0,0 +1,140 @@
...
+SQRESULT SQ_SendToNamedPipe(void* sqvm)
+{
+	try
+	{
+		if (Cvar_ns_enable_named_pipe->GetInt() && isConnected)
+		{
+			SendMessageToPipe(ServerSq_getstring(sqvm, 1));
+		}
+	}
+	catch (exception _e)
```
Mauler125, line 64: Pass parameter "_e" by const reference instead, no modification is performed.

NorthstarDLL/namedpipeclient.cpp
Mauler125, line 49: Pass parameter "message" by const reference instead, no modification is performed.

NorthstarDLL/namedpipeclient.cpp
```diff
@@ -0,0 +1,140 @@
...
+				SendMessageToPipe(ServerSq_getstring(sqvm, 1));
+			}
+			else
+			{
+				spdlog::error("INVALID_HANDLE_VALUE while opening named pipe");
+				spdlog::error(GetLastError());
+			}
+		}
+	}
+	catch (exception _e)
```
Mauler125, line 102: Pass parameter "_e" by const reference instead, no modification is performed.

NorthstarDLL/namedpipeclient.cpp
```diff
@@ -0,0 +1,140 @@
...
+	{
+		if (Cvar_ns_enable_named_pipe->GetInt() && isConnected)
+		{
+			SendMessageToPipe(ServerSq_getstring(sqvm, 1));
+
+			CloseHandle(hPipe);
+			isConnected = false;
+		}
+	}
+	catch (exception _e)
```
Mauler125, line 123: Pass parameter "_e" by const reference instead, no modification is performed.

NorthstarDLL/namedpipeclient.cpp
```diff
@@ -0,0 +1,140 @@
+#pragma once
+#include "pch.h"
+#include "namedpipeclient.h"
+#include "squirrel.h" // Squirrel
+
+// Named Pipe Stuff
+#include <windows.h>
```
Mauler125, line 7: Some of these are already included in the PCH file, remove extraneous includes.

NorthstarDLL/namedpipeclient.cpp
```diff
@@ -0,0 +1,140 @@
...
+#include <conio.h>
+#include <tchar.h>
+#include <string>
+#include <atlstr.h>
+
+using namespace std;
```
Mauler125, line 13: All symbols from namespace "std" are imported to the global scope, but only "string" and "exception" are used. Only declare used types, or access types from its namespace.

Mauler125 (commented):
There are a few small things that could be improved.

GeckoEidechse:
Named pipes cause a lot of issues (e.g. with Docker builds) and more so hardcoded names.

I know exactly for which feature this is a prerequisite of so it kinda hurts me to shut down this PR ;-;

Anyway this would probably be more suitable for a plugin, once that API is (eventually) rewritten. For now I don't think this is a good idea to merge this unfortunately :c

pg9182:
Furthermore, named pipes don't integrate well with squirrel compared to alternative systems, and this implementation is one way only, which only solves a fraction of usecases.

Also not a fan of the flexible message lengths either.

Neoministein:
Sad, but understandable.

> Furthermore, named pipes don't integrate well with squirrel compared to alternative systems,

Is there a system that is already implemented?
<br/>

> Also not a fan of the flexible message lengths either.

I am not smart enough to impl message splitting in c++.

PR #282, opened 2022-10-03
Clear token when not needed anymore

emma-miler:
Also blocks migrateme

### DIFF
--- a/NorthstarDLL/clientauthhooks.cpp
+++ b/NorthstarDLL/clientauthhooks.cpp
@@ -15,6 +15,11 @@ const int NOT_DECIDED_TO_SEND_TOKEN = 0;
 const int AGREED_TO_SEND_TOKEN = 1;
 const int DISAGREED_TO_SEND_TOKEN = 2;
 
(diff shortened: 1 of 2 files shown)


Erlite:
based

NorthstarDLL/clientauthhooks.cpp
```diff
@@ -15,6 +15,11 @@ const int NOT_DECIDED_TO_SEND_TOKEN = 0;
 const int AGREED_TO_SEND_TOKEN = 1;
 const int DISAGREED_TO_SEND_TOKEN = 2;
 
+typedef char* (*Auth3PTokenType)();
+Auth3PTokenType Auth3PToken;
+
+char* token_location = 0x0;
```
pg9182, line 21: Would make this static, but it's not urgent.

pg9182 (approved):
LGTM

NorthstarDLL/clientauthhooks.cpp
Erlite: unbased

PR #283, opened 2022-10-07
[COOP] load sp_ rpaks on dedi

ASpoonPlaysGames:
Unsure if this actually fixes coop on dedi, so could use some testing, however, it should do if the issue matches my understanding of it


BobTheBob9 (approved):
i hate that we have to do this but i also don't really see another way of doing it, probably won't be so bad if we pack our own sp_ rpaks with just the anir data though

ASpoonPlaysGames:
> i hate that we have to do this but i also don't really see another way of doing it, probably won't be so bad if we pack our own sp_ rpaks with just the anir data though

I agree, eventually I would think the goal would be to pack our own rpaks, but there doesn't seem to be all that much anir reverse engineering demand tbh

pg9182 (commented):
Doesn't seem to break the dedi (`ghcr.io/pg9182/northstar-dedicated-ci:1-tf2.0.11.0-ns1.9.7-ci.59de24a5.13.3210780084-nslauncher.0fad87eb.3204986072`) with optimized server files, so seems fine as far as I'm concerned.

catornot (approved):
tested this today with a dedicated server and it worked :)

PR #286, opened 2022-10-15
Fix crash that shouldnt happen anyway

emma-miler:


pg9182 (approved):
LGTM

emma-miler:
> The crash wasn't on mainmenupromos btw, it was during auth_with_server, iirc.

You're right, i accidentally fixed the bug for a similar error :P

PR #290, opened 2022-10-17
Fix token clearing hook

emma-miler:
A certain someone forgot to put an `0x` in front of their hex number, meaning the offset is wrong. This would cause token to be leaked again.


GeckoEidechse (approved):
Confirmed to fix the token leak regression in testing.

PR #291, opened 2022-10-17
Fix `sv_alltalk 1` (broken originally by respawn changes)

BobTheBob9:
Restores the behaviour of the convar sv_alltalk to do what it did in normal source (i.e. make voice chat global rather than team-restricted), still uses vanilla behaviour in mp_lobby regardless of convar value


NoCatt:
[nslog2022-11-12_15-50-55.txt](https://github.com/R2Northstar/NorthstarLauncher/files/9995232/nslog2022-11-12_15-50-55.txt)
Here are logs I was told to send by Spoon 
It crashes when 2 (or maybe more) players are in the match and one of them activates vc

ASpoonPlaysGames:
ok so we tested again and were unable to reproduce, idk

BobTheBob9:
NOTE: the ui for displaying talking players isn't properly updated for this! see https://github.com/Syampuuh/Titanfall2/blob/master/scripts/vscripts/client/cl_voice_hud.gnut

ASpoonPlaysGames:
> NOTE: the ui for displaying talking players isn't properly updated for this! see https://github.com/Syampuuh/Titanfall2/blob/master/scripts/vscripts/client/cl_voice_hud.gnut

Should be a fairly easy fix in script, when this gets merged I'll open a mods PR

BobTheBob9:
> > NOTE: the ui for displaying talking players isn't properly updated for this! see https://github.com/Syampuuh/Titanfall2/blob/master/scripts/vscripts/client/cl_voice_hud.gnut
> 
> Should be a fairly easy fix in script, when this gets merged I'll open a mods PR

might not be, depending on how it works and your fix you might show people talking on the other team while alltalk is off

BobTheBob9:
> > > NOTE: the ui for displaying talking players isn't properly updated for this! see https://github.com/Syampuuh/Titanfall2/blob/master/scripts/vscripts/client/cl_voice_hud.gnut
> > 
> > 
> > Should be a fairly easy fix in script, when this gets merged I'll open a mods PR
> 
> might not be, depending on how it works and your fix you might show people talking on the other team while alltalk is off

nvm, should be fine actually

BobTheBob9:
gecko told me to write this here: vanilla has a bug that prevents listen server host in mp_lobby from having their voice chat work, this IS NOT related to this pr!!! but should be fixed separately

GeckoEidechse (approved):
Confirmed working in testing.

We booted up a private match with FFA and `sv_alltalk 1` and using voice chat key caused voice chat to propagate to everyone in lobby.

(this is server side only so nothing breaking. The only thing is that UI won't properly update for old version)

PR #293, opened 2022-10-17
Add support for querying player userinfo cvars in script

BobTheBob9:


p0358 (approved):

BobTheBob9:
not merging this yet as this needs to use the new KeyValues class in launcher rather than just directly calling KeyValues::GetString

GeckoEidechse:
> not merging this yet as this needs to use the new `KeyValues` class in launcher rather than just directly calling `KeyValues::GetString`

Should I convert to draft then maybe? ^^

PR #294, opened 2022-10-21
Fixup more cvars

BobTheBob9:
removes unnecessary devonly flags from a bunch of cvars and a few commands
this was initially made to add re-remove `cl_updaterate_mp`'s devonly flag, the code for which was removed in refactor (oops), but i decided to go through all the devonly cvars and remove devonly from everything unnecessary
you can still fully strip devonly and hidden from cvars/commands with a commandline flag, but this is mainly for normal use

POSSIBLE CHANGES: 
`cl_footstep_event_max_dist` and `cl_footstep_event_max_dist_titan` control the distance at which footsteps from pilots and titans can be heard, I've made these `FCVAR_CHEAT` (i.e. require `sv_cheats 1`), but maybe they should be made `FCVAR_REPLICATED` instead? (i.e. server controls value, which is then sent to client)
only issue with this is that they don't exist on dedicated server atm since they're client vars, so they'd have to be redefined there

`cl_draw_player_model` and `cl_always_draw_3p_player` could maybe be made to require `sv_cheats`? but they don't really offer much in the way of cheaty-ness so idk probably not worth (one toggles whether player should be visible in 3p, other one draws 3p model ontop of 1p model in 1p)


GeckoEidechse (approved):
Didn't crash in testing.

PR #298, opened 2022-10-23
allow kb_act.lst to be compiled from multiple mods

BobTheBob9:
allows a `kb_act.lst` file to be placed in mod root (i.e. next to `mod.json`) to add additional binds to the keyboard binds menu in settings, we previously added binds to this in `Northstar.Client` by directly replacing the file, which isn't great if other mods want to edit it
`kb_act.lst` can be reloaded on ui script vm compile (i.e. `uiscript_reset`)


GeckoEidechse:
Someone wanna make like a 2 line docs PR to https://github.com/R2Northstar/ModdingDocs/ mentioning that this is a thing now? ^^

GeckoEidechse:
Merge conflicts

PR #304, opened 2022-10-26
Update Origin -> Auth Provider

emma-miler:
With the public deprecation of Origin, our messaging should reflect this change. This changes some text strings to better reflect the current state. Also adds the `-noAuthProviderStartup` argument, though this can be removed/replace `-noOriginStartup` argument if wanted.

I'm not sure of the process name of EA App (is it just EADesktop?), so im keeping the old Origin.exe check in for now, though this should probably be removed in the future.

### DIFF
--- a/NorthstarLauncher/main.cpp
+++ b/NorthstarLauncher/main.cpp
@@ -130,7 +130,7 @@ void LibraryLoadError(DWORD dwMessageId, const wchar_t* libName, const wchar_t* 
 	MessageBoxA(GetForegroundWindow(), text, "Northstar Launcher Error", 0);
 }
 
-void AwaitOriginStartup()
+void AwaitAuthProviderStartup()
 {
 	WSADATA wsaData;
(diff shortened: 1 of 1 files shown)


NorthstarLauncher/main.cpp
```diff
@@ -156,7 +156,7 @@ void AwaitOriginStartup()
 			recv(sock, buf, 4096, 0);
 			std::cout << buf << std::endl;
 
-			// honestly really shit, this isn't needed for origin due to being able to check OriginClientService
```
BobTheBob9, line 159: this is inaccurate, we handle ea app and origin differently, this check IS necessary for ea app, but isn't for origin, hence the original comment

BobTheBob9:
this seems kinda unnecessary to me? origin still fully works and our launching code hasn't changed at all because of the deprecation of it, there's not much reason to deprecate support for it until origin can't be used to play the game

BobTheBob9:
that being said we do need better support for launching using ea app, but that's a different problem from what this pr serves

PR #306, opened 2022-10-27
Improve authentication flow

RoyalBlue1:


NorthstarDLL/serverauthentication.cpp
```diff
@@ -267,7 +297,24 @@ void*,, (
 	pNextPlayerToken = serverFilter;
 	iNextPlayerUid = uid;
 
-	return CBaseServer__ConnectClient(server, a2, a3, a4, a5, a6, a7, a8, serverFilter, a10, a11, a12, a13, a14, uid, a16, a17);
+	spdlog::info(
+		"CBaseServer__ClientConnect attempted connection with uid {}, playerName '{}', serverFilter '{}'", uid, playerName, serverFilter);
+
+	CBaseServer__RejectConnection(self, *((int*)self + 3), addr, "Invalid Name.\n");
+	return nullptr;
```
pg9182, line 304: Uhh, this prevents the rest from executing.

pg9182 (changes requested):

NorthstarDLL/serverauthentication.cpp
```diff
@@ -276,26 +320,21 @@ bool,, (R2::CBaseClient* self, char* name, void* netchan_ptr_arg, char b_fake_pl
...
+		R2::CBaseClient__Disconnect(self, 1, "Invalid Name.\n");
 
-	if (strlen(name) >= 64) // fix for name overflow bug
-		R2::CBaseClient__Disconnect(self, 1, "Invalid name");
-	else if (
-		!g_pServerAuthentication->AuthenticatePlayer(self, iNextPlayerUid, pNextPlayerToken) &&
-		g_pServerAuthentication->m_bRequireClientAuth)
-		R2::CBaseClient__Disconnect(self, 1, "Authentication Failed");
+	if (!g_pBanSystem->IsUIDAllowed(iNextPlayerUid))
+		R2::CBaseClient__Disconnect(self, 1, "Banned From server.\n");
```
pg9182, line 335: else?

pg9182 (changes requested):

pg9182 (approved):

NorthstarDLL/serverauthentication.cpp
```diff
@@ -276,26 +326,29 @@ bool,, (R2::CBaseClient* self, char* name, void* netchan_ptr_arg, char b_fake_pl
 // clang-format on
 {
 	// try changing name before all else
-	g_pServerAuthentication->VerifyPlayerName(self, pNextPlayerToken, name);
+	// g_pServerAuthentication->VerifyPlayerName(self, pNextPlayerToken, name);
```
GeckoEidechse, line 329: Why was this just commented out instead of removed? ^^

NorthstarDLL/serverauthentication.cpp
```diff
@@ -146,7 +146,9 @@ bool ServerAuthenticationManager::VerifyPlayerName(const char* authToken, const
 		spdlog::info("Rejecting player with name '{}' because the name contains forbidden characters", name);
 		return false;
 	}
-
+	// TODO: We should really have a better way of doing this for singleplayer
+	// Best way of doing this would be to check if server is actually in singleplayer mode, or just running a SP map in multiplayer
+	// Currently there's not an easy way of checking this, so we just disable this check if mapname starts with `sp_`
```
pg9182, line 150: We should make it more clear that this completely removes player name checking.

pg9182 (changes requested):

NorthstarDLL/serverauthentication.h
```diff
@@ -49,7 +51,7 @@ class ServerAuthenticationManager
 	void RemovePlayer(R2::CBaseClient* player);
 	bool CheckDuplicateAccounts(R2::CBaseClient* player);
 	bool AuthenticatePlayer(R2::CBaseClient* player, uint64_t uid, char* authToken);
-	void VerifyPlayerName(R2::CBaseClient* player, char* authToken, char* name);
+	bool VerifyPlayerName(const char* authToken, const char* name);
```
GeckoEidechse, line 54: Value should be constant hence `const` and it allows for interaction with spdlog (`char *` sometimes give errors with spdlog but `const char *` works according to emma and blue).

NorthstarDLL/serverauthentication.cpp
```diff
@@ -114,23 +115,56 @@ void ServerAuthenticationManager::RemovePlayer(R2::CBaseClient* player)
 		m_PlayerAuthenticationData.erase(player);
 }
 
-void ServerAuthenticationManager::VerifyPlayerName(R2::CBaseClient* player, char* authToken, char* name)
+bool checkIsPlayerNameValid(const char* name)
```
GeckoEidechse, line 118: Should have a doc comment saying what is considered "valid".

NorthstarDLL/serverauthentication.cpp
```diff
@@ -114,23 +115,56 @@ void ServerAuthenticationManager::RemovePlayer(R2::CBaseClient* player)
...
+{
+	int len = strlen(name);
+	if (len >= 64)
```
GeckoEidechse, line 122: For the `64` we probably should have a macro or constant instead of a :sparkles: magic value :sparkles:

pg9182: I disagree (for now, at least); we do this in so many places for so many things, and it's not safe to change it anyways rn.

NorthstarDLL/serverauthentication.h
pg9182: Yes?

NorthstarDLL/serverauthentication.cpp
GeckoEidechse: Moved to #309

pg9182 (approved):

NorthstarDLL/serverauthentication.cpp
GeckoEidechse: Sorta resolved with f6e57dda2dadef19d78b5d6e1379e9faa5f5975f

NorthstarDLL/serverauthentication.cpp
GeckoEidechse: Resolved with abc203b39ccfec555f67857845b2823d97839161

pg9182 (approved):

NorthstarDLL/serverauthentication.cpp
GeckoEidechse: Resolved with 9d626d128a7315b5433abe9070b3ce92151943f9

pg9182:
This might be causing R2Northstar/Atlas#15. I hadn't tested singleplayer when I approved this PR.

@BobTheBob9

PR #280, opened 2022-09-29
Log colours (again)

ASpoonPlaysGames:
it seems that the crashes caused by this was a viper moment, so making a new pr for it

PR #308, opened 2022-10-27
Logging improvements + Colours

ASpoonPlaysGames:
Code is a bit of a mess atm I think, so marking as draft for now :)

Should work tho, test it if you want

Also sorta supercedes #280 i guess?


ASpoonPlaysGames:
the code for finding tags could probably be improved I think? but the rest isn't awful I think

ASpoonPlaysGames:
marking as draft as per conversation on discord

PR #315, opened 2022-10-30
remove dev-only flag from save_enable

H0L0theBard:
Unsure if this is the proper fix for this but it seems as if the client tries to set this to 0 while in a server and cannot due to it being dev-only. 
We previously had it without the dev-only flag and I don't believe there should be any negative effects due to this. 

Should fix R2Northstar/Northstar#359

### DIFF
--- a/NorthstarDLL/misccommands.cpp
+++ b/NorthstarDLL/misccommands.cpp
@@ -265,6 +265,8 @@ void FixupCvarFlags()
 
 		{"ai_ainRebuildOnMapStart", FCVAR_DEVELOPMENTONLY},
 
+		{"save_enable", FCVAR_DEVELOPMENTONLY},
+
 		// cheat commands
 		{"switchclass", FCVAR_DEVELOPMENTONLY},
 		{"set", FCVAR_DEVELOPMENTONLY},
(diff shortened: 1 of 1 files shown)


NorthstarDLL/misccommands.cpp
```diff
@@ -277,6 +277,8 @@ void FixupCvarFlags()
 		{"damagedefs_reparse_client", FCVAR_DEVELOPMENTONLY},
 		{"playerSettings_reparse", FCVAR_DEVELOPMENTONLY},
 		{"_playerSettings_reparse_Server", FCVAR_DEVELOPMENTONLY},
+
+		{"save_enable", FCVAR_DEVELOPMENTONLY},
```
BobTheBob9, line 281: this should probably be moved to be under the sys/engine settings comment/header, other than that looks great

BobTheBob9 (changes requested):

BobTheBob9:
i was planning on pring this anyway but will probably be merging this since you've made a pr for it

NorthstarDLL/misccommands.cpp
H0L0theBard: Fixed :)

BobTheBob9 (approved):

PR #318, opened 2022-11-04
Add cache invalidation to server presence

emma-miler:
This fixes a bug with playercounts being very delayed in the serverlist


GeckoEidechse:
Issue seems to actually be something else, cause dedi server itself reports old player count in console window title.

GeckoEidechse:
Superseded by #322

PR #320, opened 2022-11-05
Logging improvements + Colours (new)

emma-miler:
Supercedes #308.
Adds a list of custom loggers for different applications.
Adds the `NS::log` and `NS::Colors` namespaces
Extends `Color` functionality
Moves log hooks to separate file

Confirmed working with VTOL and Viper


NorthstarDLL/loghooks.cpp
```diff
@@ -0,0 +1,262 @@
...
+	return 0;
+}
+
+// clang-format off
+AUTOHOOK(ConCommand_echo, engine.dll + 0x123680,
+void,, (const CCommand& arg))
+// clang-format on
+{
+	if (arg.ArgC() >= 2)
+		spdlog::info("[echo] {}", arg.ArgS());
```
ASpoonPlaysGames, line 135: Should echo be using it's own logger? It has a colour defined, but it doesn't get used?

ASpoonPlaysGames (changes requested):
Just a couple of things, nothing major

NorthstarDLL/loghooks.cpp
emma-miler: Oh i kinda forgor about this tbf

NorthstarDLL/color.h
```diff
@@ -130,6 +154,44 @@ class Color
 		return newColor;
 	}
 
+	std::string ToANSIColor()
+	{
+		std::string out = "\033[38;2;";
+		out += std::to_string(_color[0]) + ";";
+		out += std::to_string(_color[1]) + ";";
+		out += std::to_string(_color[2]) + ";";
+		out += "49m";
```
ASpoonPlaysGames, line 163: 49 resets the background colour of the text, there shouldn't be any problems with this for now, however if we want to add support for custom background colours (not doable in the source console afaik?) this could cause issues. 

I wouldn't say that this is a problem atm, just something to note, maybe with a comment or something?

NorthstarDLL/logging.cpp
```diff
@@ -12,257 +11,194 @@
...
-	HUD_PRINTCONSOLE,
-	HUD_PRINTTALK,
-	HUD_PRINTCENTER
-};
+		out += str;
+	}
+	// print the string to the console - this is definitely bad i think
+	HANDLE handle = GetStdHandle(STD_OUTPUT_HANDLE);
+	auto ignored = WriteConsoleA(handle, out.c_str(), std::strlen(out.c_str()), nullptr, nullptr);
+	(void)ignored;
```
ASpoonPlaysGames, line 104: I wrote this code, adapted from how spdlog itself does stuff, but I am not confident in it's functionality. If you think that it looks good, you could remove the comment saying that it is bad, or if it is actually bad, i would appreciate if you could improve it

NorthstarDLL/logging.cpp
```diff
@@ -12,257 +11,194 @@
...
-
-	case TextMsgPrintType_t::HUD_PRINTCONSOLE:
-		auto endpos = strlen(text);
-		if (text[endpos - 1] == '\n')
-			text[endpos - 1] = '\0'; // cut off repeated newline
-
-		spdlog::info(text);
-		break;
-	}
+	std::cout << std::flush;
```
ASpoonPlaysGames, line 109: I also am not sure if this is even needed tbh, since we use WriteConsoleA? idk

NorthstarDLL/logging.cpp
```diff
@@ -12,257 +11,194 @@
...
+		dwMode |= ENABLE_PROCESSED_OUTPUT | ENABLE_VIRTUAL_TERMINAL_PROCESSING;
+
+		if (!SetConsoleMode(hOutput, dwMode)) // Some editions of Windows have 'VirtualTerminalLevel' disabled by default.
 		{
-			spdlog::error("Failed creating log file!");
-			MessageBoxA(
-				0, "Failed creating log file! Make sure the profile directory is writable.", "Northstar Warning", MB_ICONWARNING | MB_OK);
+			// If 'VirtualTerminalLevel' can't be set, just disable ANSI color, since it wouldnt work anyway.
+			spdlog::warn("could not set VirtualTerminalLevel. Disabling color output");
+			g_bSpdLog_UseAnsiColor = false;
```
ASpoonPlaysGames, line 143: I believe that this function is called before InitialiseLogging, meaning that g_bSpdLog_UseAnsiColor hasn't been set to true yet, and spdlog::warn might not do anything?

NorthstarDLL/logging.cpp
```diff
@@ -12,257 +11,194 @@
...
+
+	// set whether we should use colour in the logs
+	g_bSpdLog_UseAnsiColor = !strstr(GetCommandLineA(), "-noansiclr");
+
+	// create our console sink
+	auto sink = std::make_shared<ExternalConsoleSink>();
+	// set the pattern
+	if (g_bSpdLog_UseAnsiColor)
+		// dont put the log level in the pattern if we are using colours, as the colour will show the log level
+		sink->set_pattern("[%H:%M:%S] [NAME] [LVL] %v");
```
ASpoonPlaysGames, line 172: Idk if you have looked into this or not, but instead of doing string replacement on [NAME] and [LVL], it might be worth doing custom formatting flags for the sink or something?

NorthstarDLL/logging.h
```diff
@@ -1,4 +1,131 @@
...
+
+	extern std::shared_ptr<ColoredLogger> NORTHSTAR;
+}; // namespace NS::log
+
+void RegisterCustomSink(std::shared_ptr<CustomSink> sink);
+
+inline bool g_bSpdLog_UseAnsiColor = true;
+
+// Could maybe use some different names here, idk
+static const char* level_names[] {"trac", "dbug", "info", "warn", "errr", "crit", "off"};
```
ASpoonPlaysGames, line 110: How much do we care about the level names being the same length? It might be more readable to use single letters for the different levels here (T, D, I, W, E, C)?

NorthstarDLL/color.cpp
```diff
@@ -0,0 +1,26 @@
...
+	Color NATIVE_SV		(150, 50 , 150);
+	Color NATIVE_EN		(252, 133, 153);
+	Color FILESYSTEM	(0  , 150, 150);
+	Color RPAK			(255, 190, 0  );
+	Color NORTHSTAR		(66 , 72 , 128);
+	Color ECHO			(150, 150, 159);
+
+	Color TRACE		(0  , 255, 255);
+	Color DEBUG		(0  , 255, 255);
+	Color INFO		(16 , 160, 16 );
```
ASpoonPlaysGames, line 20: I personally would prefer info to be a more subtle colour as it doesn't signify anything of importance, I don't mind the green though

ASpoonPlaysGames (changes requested):

ASpoonPlaysGames:
Tested with various mod managers etc. seems to work fine (can't speak for other operating systems though)

PR #321, opened 2022-11-05
Restore `-waitfordebugger` flag

emma-miler:


NorthstarLauncher/main.cpp
```diff
@@ -324,6 +324,14 @@ HMODULE LoadDediStub(const char* name)
 int main(int argc, char* argv[])
 {
 
+	if (strstr(GetCommandLineA(), "-waitfordebugger"))
+	{
+		while (!IsDebuggerPresent())
+		{
+			Sleep(100);
```
GeckoEidechse, line 331:
```suggestion
			// Sleep 100ms to give debugger time to attach.
			Sleep(100);
```

GeckoEidechse (approved):
Seems to work in testing (as in it blocks if no debugger present).

NorthstarLauncher/main.cpp
ASpoonPlaysGames: image
yes, github, that is exactly what they did to that code.

In all seriousness, commenting this I would say is pretty important

PR #322, opened 2022-11-06
Fix player count in server presence

emma-miler:
Bob forgot to add this check in refactor


GeckoEidechse (approved):
Confirmed working in testing.

PR #323, opened 2022-11-06
Fix player count server presence (again)

emma-miler:
Oops


GeckoEidechse (approved):
Works in testing, now player count also updates properly when player leaves.

PR #324, opened 2022-11-08
New syntax for reversed structs

emma-miler:
This new file introduces a new syntax for defining structs we have reversed.
This allows us to define fields by their offset, and removes the need for padding fields, making them a lot more maintainable.

### DIFF
--- a/NorthstarDLL/filesystem.h
+++ b/NorthstarDLL/filesystem.h
@@ -19,12 +19,12 @@ OFFSET_STRUCT(VPKFileEntry)
 
 #pragma pack(push, 1)
 // clang-format off
-OFFSET_STRUCT(VPKData)
+struct VPKData
 {
-	STRUCT_SIZE(0x24A)
-	FIELD(0x5, char path[255])
-	FIELD(0x238, int32_t numEntries)
-	FIELD(0x23C, VPKFileEntry* entries)
+	char* directory;
+	char* filename;
+	char* extension;
+	unsigned char unknown[0x38];
 };
 // clang-format on
 #pragma pack(pop)
--- a/NorthstarDLL/filesystem.h
+++ b/NorthstarDLL/filesystem.h
@@ -9,10 +9,12 @@ typedef void* FileHandle_t;
 // clang-format off
 OFFSET_STRUCT(VPKFileEntry)
 {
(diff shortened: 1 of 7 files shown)


NorthstarDLL/filesystem.h
```diff
@@ -5,25 +5,28 @@
...
+// clang-format off
+OFFSET_STRUCT(VPKFileEntry)
 {
-	char* directory;
-	char* filename;
-	char* extension;
-	unsigned char unknown[0x38];
+	STRUCT_SIZE(0x44);
+	FIELD(0x0, char* directory)
+	FIELD(0x4, char* filename)
```
BobTheBob9, line 14: 4 bytes for a ptr?

BobTheBob9: honestly i don't see much of a reason to explicitly define field offsets here, maybe a macro to define multiple fields inline from an offset (0 in this case) could be useful in situations where you don't need to do this? (assuming it isn't possible with the preexisting one

NorthstarDLL/structs.h
```diff
@@ -0,0 +1,62 @@
...
+// To use these macros, define a struct like so:
+/*
+OFFSET_STRUCT(Name)
+{
+	STRUCT_SIZE(0x100)		// Total in-memory struct size
+	FIELD(0x0, int field)	// offset, signature
+}
+*/
+
+#define OFFSET_STRUCT(name) union name
```
BobTheBob9, line 26: debating whether this should be called something like INTERNAL_STRUCT instead, but unsure

BobTheBob9 (changes requested):
looks great mostly

NorthstarDLL/filesystem.h
emma-miler: This is fair actually. I could revert this to its old def

NorthstarDLL/structs.h
emma-miler: Well its not really internal or anything. OFFSET_STRUCT describes what it is: a struct defined with offsets

BobTheBob9: yeah fair

BobTheBob9 (approved):

PR #328, opened 2022-11-13
remove sv_cheats requirement from r_drawviewmodel

BobTheBob9:
allows r_drawviewmodel to be set without cheats, this is fine, since people already do similar things with mods anyway, just makes it easier


emma-miler (approved):

pg9182 (approved):
I don't see an issue with this; there's already other (albeit more convoluted) ways to do the sam e thing anyways.

ASpoonPlaysGames (approved):
Provided that the convar has been tested to actually function, this looks good. Worst case is that the convar crashes or something and people just dont use it ig

emma-miler (approved):
This is also allowed in CSGO, so fine by me

PR #330, opened 2022-11-14
Fix DecodeJSON() not returning SQRESULT_NOTNULL.

Erlite:
This PR simply makes the DecodeJSON() SQ function return `SQRESULT_NOTNULL` as to fix an issue where the Squirrel side would never return the decoded JSON table.


emma-miler (approved):
Did the same thing when fixing warnings. Tested and working

PR #331, opened 2022-11-14
Fix slow keyvalues building

emma-miler:
Turns out that under windows, if you throw a runtime exception and catch it, it still passes through the exception filter. I was not aware of this, and was generating a dump file whenever the exception filter ran, since i assumed we would always be crashing at one point.

### DIFF
--- a/NorthstarDLL/crashhandler.cpp
+++ b/NorthstarDLL/crashhandler.cpp
@@ -31,12 +31,15 @@ void PrintExceptionLog(ExceptionLog& exc)
 	// General crash message
 	spdlog::error("Northstar version: {}", version);
 	spdlog::error("Northstar has crashed! a minidump has been written and exception info is available below:");
-	spdlog::error("Loaded mods: ");
-	for (const auto& mod : g_pModManager->m_LoadedMods)
+	if (g_pModManager)
 	{
-		if (mod.m_bEnabled)
+		spdlog::error("Loaded mods: ");
+		for (const auto& mod : g_pModManager->m_LoadedMods)
 		{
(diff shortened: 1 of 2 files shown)


ASpoonPlaysGames:
> I was not aware of this, and was generating a dump file whenever the exception filter ran

oh my god

NorthstarDLL/crashhandler.cpp
```diff
@@ -263,41 +266,47 @@ long GenerateExceptionLog(EXCEPTION_POINTERS* exceptionInfo)
...
 	CreateMiniDump(exceptionInfo);
-	PrintExceptionLog(*exc);
+	PrintExceptionLog(*storedException);
 	return EXCEPTION_EXECUTE_HANDLER;
 }
 
 long __stdcall ExceptionFilter(EXCEPTION_POINTERS* exceptionInfo)
 {
-	if (!IsDebuggerPresent())
+	if (true)
```
GeckoEidechse, line 294: Uuuhh

emma-miler: i forgor

GeckoEidechse:
74aa17d543d4de8a8c874abbd72ed06ac2fbb563 fixes all crashes being handled being treated as runtime errors.

(@emma-miler couldn't be arsed to type this out so this is me publicly shaming her for it <3)

emma-miler:
shame me harder UwU

GeckoEidechse (approved):
Still holds on newest commit:

> Works in testing in the sense that it doesn't spam log folder with dumps

PR #333, opened 2022-11-15
Fix loggers not sinking to log file

emma-miler:
I forgot to update this piece of code to make it so that the file sink gets pushed to all loggers instead of only the `[NORTHSTAR]` one


GeckoEidechse (approved):
Confirmed working. Previous log files only had `[Northstar]` stuff, new ones seem to have all.

PR #338, opened 2022-11-20
Validate serverlist response using a JSON schema

ASpoonPlaysGames:
Currently this works, but I do not like that I am using a raw string literal for the schema, is there some way to have a json file or something separate that gets packed or whatever at compile time?

Also the schema itself could maybe use some work, namely for regions. Currently they are set to be required, but this might be bad?

### DIFF
--- a/NorthstarDLL/masterserver.cpp
+++ b/NorthstarDLL/masterserver.cpp
@@ -261,7 +261,6 @@ void MasterServerManager::RequestServerList()
 			"lastHeartbeat",
 			"id",
 			"name",
-			"region",
 			"description",
 			"playerCount",
 			"maxPlayers",
--- a/NorthstarDLL/masterserver.cpp
+++ b/NorthstarDLL/masterserver.cpp
@@ -214,7 +214,7 @@ void MasterServerManager::RequestServerList()
 				rapidjson::Document schemaDoc;
 				// this schema validates each entry in the array of servers
 				schemaDoc.Parse(
-R"({
+					R"({
 	"type": "object",
 		"properties": {
 			"lastHeartbeat": {
--- a/NorthstarDLL/masterserver.cpp
+++ b/NorthstarDLL/masterserver.cpp
@@ -13,6 +13,7 @@
 #include "rapidjson/stringbuffer.h"
 #include "rapidjson/writer.h"
 #include "rapidjson/error/en.h"
+#include "rapidjson/schema.h"
 
 #include <cstring>
 #include <regex>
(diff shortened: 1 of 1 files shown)


NorthstarDLL/masterserver.cpp
```diff
@@ -209,6 +210,78 @@ void MasterServerManager::RequestServerList()
...
+			},
+			"id": {
+				"type": "string"
+			},
+			"name": {
+				"type": "string"
+			},
+			"region": {
+				"type": "string"
+			},
```
pg9182, line 231: This field is optional.

NorthstarDLL/masterserver.cpp
```diff
@@ -209,6 +210,78 @@ void MasterServerManager::RequestServerList()
 					goto REQUEST_END_CLEANUP;
 				}
 
+				// rapidjson conforms to JSON Schema draft 4
+				rapidjson::Document schemaDoc;
+				// this schema validates each entry in the array of servers
+				schemaDoc.Parse(
+					R"({
```
pg9182, line 217: Either indent the whole thing properly, or preferably start the object on a new line after the quote and keep the whole thing to the left.

NorthstarDLL/masterserver.cpp
```diff
@@ -221,24 +294,11 @@ void MasterServerManager::RequestServerList()
...
-					if (!serverObj.HasMember("id") || !serverObj["id"].IsString() || !serverObj.HasMember("name") ||
-						!serverObj["name"].IsString() || !serverObj.HasMember("description") || !serverObj["description"].IsString() ||
-						!serverObj.HasMember("map") || !serverObj["map"].IsString() || !serverObj.HasMember("playlist") ||
-						!serverObj["playlist"].IsString() || !serverObj.HasMember("playerCount") || !serverObj["playerCount"].IsNumber() ||
-						!serverObj.HasMember("maxPlayers") || !serverObj["maxPlayers"].IsNumber() || !serverObj.HasMember("hasPassword") ||
-						!serverObj["hasPassword"].IsBool() || !serverObj.HasMember("modInfo") || !serverObj["modInfo"].HasMember("Mods") ||
-						!serverObj["modInfo"]["Mods"].IsArray())
+					if (!serverObj.Accept(validator))
 					{
 						spdlog::error("Failed reading masterserver response: malformed server object");
```
pg9182, line 299: Is it possible to have a more descriptive error here?

pg9182 (changes requested):
Okay, after taking a look, I think schemas are good in general, at least until we don't need them anymore due to squirrel json.

We definitely should keep it in a string literal though rather than some external file. It's much less error-prone and still more readable overall even without syntax highlighting.

NorthstarDLL/masterserver.cpp
ASpoonPlaysGames: This was a clang formatting moment tbh

NorthstarDLL/masterserver.cpp
ASpoonPlaysGames: What sort of thing would you want to be mentioned?

ASpoonPlaysGames:
> Okay, after taking a look, I think schemas are good in general, at least until we don't need them anymore due to squirrel json.

tbh I think it would be possible to implement this on the squirrel side right now anyway, I'm going to investigate a bit and maybe create a new PR which supercedes this

emma-miler:
> We definitely should keep it in a string literal though rather than some external file. It's much less error-prone and still more readable overall even without syntax highlighting.

I disagree with this point. Keeping it in an external file makes it easier to test without having to recompile every time. There's no real downside to loading them from file, and they could even be included in the binary if need be. Working with formatting-sensitive strings while having something like clang-format is bad enough.

pg9182:
@emma-miler 

> Keeping it in an external file makes it easier to test without having to recompile every time.

That is true, however;

> There's no real downside to loading them from file, and they could even be included in the binary if need be.

The downside is the distance from the code which uses it, which directly depends on matching the schema. As such, it should be kept close to reduce the likelihood of them diverging and introducing bugs. Furthermore, this change is likely temporary since it's being replaced with squirrel JSON IIRC, so we should keep it as unintrusive as possible.

> Even if we decide to keep them in code, i think we should have them be imported from a header, and not be defined in-line in business logic.

That, on the other hand, has no real upside. Remember that while JSON schemas aren't typically business logic, we are using it as such in this case (it's being made into an integral part of the response parsing).

NorthstarDLL/masterserver.cpp
pg9182: Fixed in 9865ef3ced549ca5b73fd7ff0c29b44e9299c4ab.

NorthstarDLL/masterserver.cpp
pg9182: That's what I meant by the second part (put the opening brace on a new line, but within the string literal so clang doesn't mangle it -- JSON doesn't care about whitespace).

NorthstarDLL/masterserver.cpp
pg9182: Ugh, I suppose what I was hoping for doesn't really exist: https://github.com/Tencent/rapidjson/issues/551.

PR #341, opened 2022-11-23
Fix dedicated server crash due to null sq context

wolf109909:
hi github

### DIFF
--- a/NorthstarDLL/hoststate.cpp
+++ b/NorthstarDLL/hoststate.cpp
@@ -113,8 +113,8 @@ void, __fastcall, (CHostState* self, double flCurrentTime, float flFrameTime))
 
 		if (g_pSquirrel<ScriptContext::CLIENT>->m_pSQVM != nullptr && g_pSquirrel<ScriptContext::CLIENT>->m_pSQVM->sqvm != nullptr)
 			g_pSquirrel<ScriptContext::CLIENT>->ProcessMessageBuffer();
- 	}
- 
+	}
+	
 	if (g_pSquirrel<ScriptContext::SERVER>->m_pSQVM != nullptr && g_pSquirrel<ScriptContext::SERVER>->m_pSQVM->sqvm != nullptr)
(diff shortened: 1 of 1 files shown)


wolf109909:
i fucked up a bit

wolf109909:
why did i close it

wolf109909:
why format fails (cries

wolf109909:
it took me fucking 
30mins trying to fix clang format
i love myself

NorthstarDLL/hoststate.cpp
```diff
@@ -8,7 +8,7 @@
 #include "r2engine.h"
 #include "limits.h"
 #include "squirrel.h"
-
+#include "dedicated.h"
```
GeckoEidechse, line 11:
```suggestion
#include "dedicated.h"

```

NorthstarDLL/hoststate.cpp
```diff
@@ -105,11 +105,15 @@ void, __fastcall, (CHostState* self, double flCurrentTime, float flFrameTime))
 	}
 
 	// Run Squirrel message buffer
-	if (g_pSquirrel<ScriptContext::UI>->m_pSQVM != nullptr && g_pSquirrel<ScriptContext::UI>->m_pSQVM->sqvm != nullptr)
-		g_pSquirrel<ScriptContext::UI>->ProcessMessageBuffer();
+	// Don't run buffers that belongs to context that is not initialized on dedicated server
```
GeckoEidechse, line 108:
```suggestion
	// Don't run buffers that belong to context that is not initialized on dedicated server
```

GeckoEidechse (changes requested):
Can't say much about the code due to lack of C++ knowledge but I can point out typos and formatting suggestions :>

emma-miler (approved):
While this PR does work, i think the underlying problem is that these contexts apparently do not get created properly on dedicated server, which i would consider to be a bug.

GeckoEidechse:
@pg9182 wanna test this with Docker?

emma-miler:
Superceded by #348

PR #346, opened 2022-11-26
Add X-Northstar-Server-Id header to auth server responses

### LINKED ISSUE #345: Auth server is not uniquely identifiable
We've had multiple people having issues with authentication caused by misconfiguring the server to have overlapping auth ports, which causes only one of them to get the player auth data and return odd authentication errors.

To make it easier to troubleshoot these cases, it should be possible to uniquely identify game servers for each auth port.

To solve this with the current API while maintaining backwards compatibility, I propose adding a server ID header to the /verify response, which will be validated by Atlas if present.

pg9182:
This will allow Atlas to verify the correct auth server for the gameserver is listening on the expected port.

fixes #345


pg9182:
Superseded by #432.

PR #349, opened 2022-11-27
Fix UI VM Destruction

emma-miler:
Fixes a bug with UI VM destruction. The functions in question take a `CSquirrelVM*` instead of an `HSquirrelVM*`, which causes `IsUIVM()` always returns false.

### DIFF
--- a/NorthstarDLL/squirrel.cpp
+++ b/NorthstarDLL/squirrel.cpp
@@ -316,19 +316,20 @@ template <ScriptContext context> CSquirrelVM* __fastcall CreateNewVMHook(void* a
 	spdlog::info("CreateNewVM {} {}", GetContextName(realContext), (void*)sqvm);
 	return sqvm;
 }
-
-template <ScriptContext context> void (*__fastcall DestroyVM)(void* a1, HSquirrelVM* sqvm);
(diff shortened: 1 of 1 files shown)


GeckoEidechse:
Fixes @BobTheBob9 code :>

NorthstarDLL/squirrel.cpp
```diff
@@ -316,19 +316,20 @@ template <ScriptContext context> CSquirrelVM* __fastcall CreateNewVMHook(void* a
 	spdlog::info("CreateNewVM {} {}", GetContextName(realContext), (void*)sqvm);
 	return sqvm;
 }
-
-template <ScriptContext context> void (*__fastcall DestroyVM)(void* a1, HSquirrelVM* sqvm);
-template <ScriptContext context> void __fastcall DestroyVMHook(void* a1, HSquirrelVM* sqvm)
+template <ScriptContext context> void (*__fastcall DestroyVM)(void* a1, CSquirrelVM* sqvm);
```
GeckoEidechse, line 319:
```suggestion

template <ScriptContext context> void (*__fastcall DestroyVM)(void* a1, CSquirrelVM* sqvm);
```

Don't steal my newlines >:(

BobTheBob9 (approved):

PR #350, opened 2022-11-28
Create CMakeLists for building everything

apophis42:
Add CMakeLists to support building through CMake and added more build and IDE configs to gitignore. Currently only using Visual Studio 2017+ as a generator will successfully compile (I only tested with VS2022 but they should be functionally equivalent). Build configuration and very partial compilation works with mingw, which will support #212.

I didn't test that the output build functions correctly since I, frankly, don't know how to do that.

This commit is purely additive so there's no chance of regressing anything.


emma-miler:
While I like the idea behind this PR, I reckon we would need some sort a script to generate these, unless VS has some of generating them automatically from filters. The reasoning behind this is that we currently use VS and msvc, and as long as we are doing that we cannot reasonably expect every PR that includes new files to also manually update the CMakeLists.

emma-miler (changes requested):
While I like the idea behind this PR, I reckon we would need some sort a script to generate these, unless VS has some of generating them automatically from filters. The reasoning behind this is that we currently use VS and msvc, and as long as we are doing that we cannot reasonably expect every PR that includes new files to also manually update the CMakeLists.

apophis42:
> While I like the idea behind this PR, I reckon we would need some sort a script to generate these, unless VS has some of generating them automatically from filters. The reasoning behind this is that we currently use VS and msvc, and as long as we are doing that we cannot reasonably expect every PR that includes new files to also manually update the CMakeLists.

Visual Studio natively supports CMake as of VS 2017, so the solution would be to only use CMakeLists. I doubt there's a good way of scripting a VS solution to CMake since CMake, itself, generates VS solutions, so you'd have a circular build process.

emma-miler:
I see, I'll have a play around with this soon. We are eventually planning to move to using clang as our compiler, and removing our dependency on VS entirely.

apophis42:
> We are eventually planning to move to using clang as our compiler, and removing our dependency on VS entirely.

I've actually got a branch for building with mingw done (which I assume would also allow building with clang), but it builds on top of this branch so I'm waiting until after this to PR it.

GeckoEidechse:
> I've actually got a branch for building with mingw done (which I assume would also allow building with clang) as I understood was one of the steps required for #212, but it builds on top of this branch so I'm waiting until after this to PR it.

Feel free to just draft PR that already.

apophis42:
> > I've actually got a branch for building with mingw done (which I assume would also allow building with clang) as I understood was one of the steps required for #212, but it builds on top of this branch so I'm waiting until after this to PR it.
> 
> Feel free to just draft PR that already.

PR'd as #351.

GeckoEidechse:
Superseded by #467 which was preferred as it also removed the solution project files and updated CI config.

PR #351, opened 2022-11-30
212 build with mingw

apophis42:
Add support for building with mingw to support #212.


NorthstarDLL/squirrel.cpp
```diff
@@ -638,16 +638,37 @@ ON_DLL_LOAD_RELIESON("client.dll", ClientSquirrel, ConCommand, (CModule module))
...
+	MAKEHOOK(
+		module.Offset(0x79A50),
+		reinterpret_cast<LPVOID>(&ScriptCompileErrorHook<ScriptContext::CLIENT>),
+		reinterpret_cast<LPVOID>(&SQCompileError<ScriptContext::CLIENT>));
 
-	MAKEHOOK(module.Offset(0x10190), &CallScriptInitCallbackHook<ScriptContext::CLIENT>, &CallScriptInitCallback<ScriptContext::CLIENT>);
+	MAKEHOOK(
+		module.Offset(0x10190),
+		reinterpret_cast<LPVOID>(&CallScriptInitCallbackHook<ScriptContext::CLIENT>),
+		reinterpret_cast<LPVOID>(&CallScriptInitCallback<ScriptContext::CLIENT>));
```
emma-miler, line 671: Not sure why the formatting changes here

NorthstarDLL/scriptuserinfo.cpp
```diff
@@ -89,7 +89,7 @@ ADD_SQFUNC("bool", GetUserInfoKVBool_Internal, "entity player, string key, bool
 	"Gets the bool value of a given player's userinfo convar by name", ScriptContext::SERVER)
 // clang-format on
 {
-	const R2::CBasePlayer* pPlayer = g_pSquirrel<ScriptContext::SERVER>->getentity<R2::CBasePlayer>(sqvm, 1);
+	const R2::CBasePlayer* pPlayer = g_pSquirrel<ScriptContext::SERVER>->template getentity<R2::CBasePlayer>(sqvm, 1);
```
emma-miler, line 92: Not sure where these are `->template` now? How/why does this even work?

NorthstarDLL/printmaps.cpp
```diff
@@ -109,7 +109,7 @@ void RefreshMapList()
 
 // clang-format off
 AUTOHOOK(_Host_Map_f_CompletionFunc, engine.dll + 0x161AE0,
-int, __fastcall, (const char const* cmdname, const char const* partial, char commands[COMMAND_COMPLETION_MAXITEMS][COMMAND_COMPLETION_ITEM_LENGTH]))
+int, __fastcall, (const char * cmdname, const char * partial, char commands[COMMAND_COMPLETION_MAXITEMS][COMMAND_COMPLETION_ITEM_LENGTH]))
```
emma-miler, line 112: `const char const*` is a valid type hint and this is throwing away pointer-to-const compiler hinting

NorthstarDLL/keyvalues.cpp
```diff
@@ -568,7 +569,7 @@ KeyValues* KeyValues::GetFirstSubKey() const
 //-----------------------------------------------------------------------------
 KeyValues* KeyValues::GetNextKey() const
 {
-	assert(this, "Member function called on NULL KeyValues");
+	assert(this);
```
emma-miler, line 572: Should not be throwing away the error messages here. Either make a new macro or do something else.

NorthstarDLL/hooks.h
```diff
@@ -259,8 +259,8 @@ class __autohook
 	namespace                                                                                                                              \
 	{                                                                                                                                      \
 		type(*callingConvention name) args;                                                                                                \
-		__autohook                                                                                                                         \
-			CONCAT2(__autohook, __LINE__)(&__FILEAUTOHOOK, __STR(name), addr, (LPVOID*)&name, (LPVOID)CONCAT2(__autohookfunc, name));      \
+		__autohook CONCAT2(__autohook, __LINE__)(                                                                                          \
+			&__FILEAUTOHOOK, __STR(name), (LPVOID*)addr, (LPVOID*)&name, (LPVOID)CONCAT2(__autohookfunc, name));                           \
```
emma-miler, line 263: Why?

NorthstarDLL/hooks.h
```diff
@@ -115,7 +115,7 @@ class __autohook
 	{
 		iAddressResolutionMode = ABSOLUTE_ADDR;
 
-		const int iFuncNameStrlen = strlen(funcName) + 1;
+		const auto iFuncNameStrlen = strlen(funcName) + 1;
```
emma-miler, line 118: Should not be `auto`, use the real type. Use `size_t`

NorthstarDLL/filesystem.h
```diff
@@ -24,7 +24,7 @@ OFFSET_STRUCT(VPKFileEntry)
 struct VPKData
 {
 	STRUCT_SIZE(0x50);
-	FIELDS(0x0,
+	STRUCT_FIELD_NOOFFSET(0x0,
```
emma-miler, line 27: Why? I specifically made the macro so that you could use the same name everywhere. Literally have to remove one line from `structs.h` if this doesnt compile with non-msvc compilers

NorthstarDLL/dllmain.cpp
```diff
@@ -78,7 +78,8 @@ bool LoadPlugins()
 			spdlog::info("Failed to load library {}: ", std::system_category().message(GetLastError()));
 			continue;
 		}
-		HRSRC manifestResource = FindResourceW(datafile, MAKEINTRESOURCEW(101), MAKEINTRESOURCEW(RT_RCDATA));
+		static const auto rt_rcdataw = MAKEINTRESOURCEW(10);
+		HRSRC manifestResource = FindResourceW(datafile, MAKEINTRESOURCEW(101), rt_rcdataw);
```
emma-miler, line 82: Why do we cache one of these but not the other? These calls are basically free and there's no reason to do this.

NorthstarDLL/concommand.h
```diff
@@ -83,8 +83,7 @@ typedef void (*FnCommandCallback_t)(const CCommand& command);
 //-----------------------------------------------------------------------------
 // Returns 0 to COMMAND_COMPLETION_MAXITEMS worth of completion strings
 //-----------------------------------------------------------------------------
-typedef int (*__fastcall FnCommandCompletionCallback)(
-	const char* partial, char commands[COMMAND_COMPLETION_MAXITEMS][COMMAND_COMPLETION_ITEM_LENGTH]);
+typedef int (*FnCommandCompletionCallback)(const char* partial, char commands[COMMAND_COMPLETION_MAXITEMS][COMMAND_COMPLETION_ITEM_LENGTH]);
```
emma-miler, line 86: Why?

NorthstarDLL/audio.cpp
```diff
@@ -15,6 +15,15 @@ extern "C"
...
 	extern void* __fastcall Audio_GetParentEvent();
+
+#ifdef __MINGW32__
+	void* __attribute__((noinline)) Audio_GetParentEvent()
+	{
+		void* r12;
+		asm volatile("mov {%%r12, %[reg]|%[reg], r12}" : [reg] "=r"(r12));
+		return r12;
+	}
+#endif
```
emma-miler, line 26: Why is this inlined here?

loader_wsock32_proxy/dllmain.cpp
```diff
@@ -118,7 +118,14 @@ BOOL WINAPI DllMain(HINSTANCE hInst, DWORD reason, LPVOID)
...
 {
 	FARPROC PA = NULL;
+#ifdef _MSC_VER
 	int RunASM();
+#elif defined(__MINGW32__)
+	void __attribute__((naked)) RunASM()
+	{
+		asm volatile ("jmp *%[PA]" : : [PA] "m" (PA));
+	}
+#endif
```
emma-miler, line 128: Why?

NorthstarLauncher/main.cpp
```diff
@@ -223,14 +223,44 @@ void EnsureOriginStarted()
...
+		if (size == 0)
+			return std::nullopt;
+
+		if (size < buffer.size())
+			return std::wstring {buffer.data()};
+
+		buffer.resize(size);
+	}
+#endif
+}
```
emma-miler, line 255: This is not at all portable. We should really be striving for purely c++20 compatible code here, and this is not that.

NorthstarDLL/squirrel.h
```diff
@@ -253,6 +248,10 @@ class SquirrelManagerBase
 #pragma endregion
 };
 
+template <ScriptContext context, typename T> inline void SqRecurseArgs(FunctionVector& v, T& arg);
+
+template <ScriptContext context, typename T, typename... Args> inline void SqRecurseArgs(FunctionVector& v, T& arg, Args... args);
+
```
emma-miler, line 254: Why is this forward declared here?

NorthstarDLL/squirrel.h
```diff
@@ -237,11 +237,6 @@ class SquirrelManagerBase
 		return __sq_setuserdatatypeid(sqvm, stackpos, typeId);
 	}
 
-	template <typename T> inline SQBool getthisentity(HSquirrelVM* sqvm, T* ppEntity)
-	{
-		return __sq_getentity(sqvm, (void**)ppEntity);
-	}
-
```
emma-miler, line 244: Why is this removed?

NorthstarDLL/scriptdatatables.cpp
```diff
@@ -252,7 +252,7 @@ REPLACE_SQFUNC(GetDataTable, (ScriptContext::UI | ScriptContext::CLIENT | Script
 			}
 
 			// add to cache and return
-			CSVData** pUserdata = g_pSquirrel<context>->createuserdata<CSVData*>(sqvm, sizeof(CSVData*));
+			CSVData** pUserdata = g_pSquirrel<context>->template createuserdata<CSVData*>(sqvm, sizeof(CSVData*));
```
emma-miler, line 255: Not sure where these are `->template` now

NorthstarDLL/memory.cpp
```diff
@@ -324,7 +325,7 @@ inline std::pair<std::vector<uint8_t>, std::string> MaskedBytesFromPattern(const
 				}
 				else
 				{
-					assert(false, "Failed to parse invalid pattern string.");
+					assert(false);
```
emma-miler, line 328: Should not be throwing away error message here

emma-miler (changes requested):
While I appreciate the effort put in here, I'm afraid some of the new code does not really match the style the codebase has. A lot of the changes here are out of scope for this PR, and should be done separately. As it stands right now, this PR should not be merged

NorthstarDLL/scriptuserinfo.cpp
apophis42: https://en.cppreference.com/w/cpp/language/dependent_name (The template disambiguator for dependent names) is required for compilation with gcc.

apophis42:
> While I appreciate the effort put in here, I'm afraid some of the new code does not really match the style the codebase has. A lot of the changes here are out of scope for this PR, and should be done separately. As it stands right now, this PR should not be merged

I will point out, besides a couple changes to change from `int` to `auto` (that cause a warning on every file compiled and was flooding the output), every change is in response to a compiler error with mingw which is what the ticket is about.

NorthstarDLL/printmaps.cpp
apophis42: clang-tidy indicates the extra `const` doesn't do anything.

NorthstarDLL/filesystem.h
apophis42: Because using the fields macro with `0x0` causes the preprocessor to try and paste `OFFEST_` and `(` which is invalid under a standards-compliant preprocessor. VS has been warning about this since the standard was changed to 20 (VS' old preprocessor is not standards-compliant) and gcc will refuse to compile this code.

NorthstarDLL/dllmain.cpp
apophis42: I couldn't find a way to get gcc to compile `MAKEINTRESOURCEW` on an `LPSTR` (the type of `RT_RCDATA`) through any amount of casting.

NorthstarDLL/concommand.h
apophis42: There's a warning (in both MSVC and gcc) that `__fastcall` doesn't do anything in a function pointer typedef and because this file is part of the preprocessed header it floods the output, which was making it hard to look for errors.

NorthstarLauncher/main.cpp
apophis42: I assumed that not unnecessarily changing the behaviour of the MSVC code would be more desirable (afaict `_wdupenv_s` is better since you don't have to call it multiple times to get the environment variable). If you'd like to change to `getenv` for both I can do that.

NorthstarDLL/squirrel.h
apophis42: gcc requires it.

NorthstarDLL/squirrel.h
apophis42: `__sq_getentity` is never declared or defined in the project. From what I can tell, MSVC doesn't care because `getthisentity` is never used so it never evaluated the template. gcc is evaluating the template regardless and doesn't compile.

NorthstarDLL/keyvalues.cpp
apophis42: Are you saying to make a new macro to replicate the functionality of the non-standard `assert` or a macro to only use the non-standard `assert` in MSVC and throw away the error message in gcc?

I'm unsure what the non-standard call actually did, because there's no documentation for an assert that takes a second parameter (https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/assert-macro-assert-wassert?view=msvc-170) and the macro appears to be defined as
```
#define assert(expression) (void)(                                                       \
            (!!(expression)) ||                                                              \
            (_wassert(_CRT_WIDE(#expression), _CRT_WIDE(__FILE__), (unsigned)(__LINE__)), 0) \
        )
```
from `C:\Program Files (x86)\Windows Kits\10\Include\10.0.22000.0\ucrt\assert.h`

If you can point to some documentation and clarify what the desired change is I can do that.

NorthstarDLL/audio.cpp
apophis42: The existing assembly can't compile under gcc and since gcc supports inline assembly, it was simpler to write the replacement that way.

loader_wsock32_proxy/dllmain.cpp
apophis42: https://github.com/R2Northstar/NorthstarLauncher/pull/351#discussion_r1038354033

NorthstarDLL/keyvalues.cpp
pg9182: @emma-miler 

https://en.cppreference.com/w/cpp/error/assert

> There is no standardized interface to add an additional message to assert errors.
> ...
> The implementation of assert in [Microsoft CRT](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/assert-macro-assert-wassert?view=msvc-160) does not conform to C++11 and later revisions, because its underlying function (_wassert) takes neither __func__ nor an equivalent replacement.

pg9182:
@apophis42 Have you considered building using clang? I've been leaning towards https://github.com/mstorsjo/llvm-mingw msvc or msys2's mingw-w64 clang-msvc (not ucrt) personally for compatibility, ease of setup, being up-to-date, and binary size.

apophis42:
> @apophis42 Have you considered building using clang? I've been leaning towards https://github.com/mstorsjo/llvm-mingw msvc or msys2's mingw-w64 clang-msvc (not ucrt) personally for compatibility, ease of setup, being up-to-date, and binary size. I've used both in projects before.

I haven't put a lot of thought into clang. At work we used to use clang through mingw and it was a horrible pain and had numerous spurious errors that changing to gcc fixed. This was 5 years ago, so it may have changed since then. clang is also becoming the least standards compliant compiler since Apple and Google started working on their own languages (swift and Carbon). https://en.cppreference.com/w/cpp/compiler_support#cpp20 That said, switching from MSVC to clang would likely require most if not at all the same changes that switching from MSVC to gcc requires since most of the changes were just making things standards-compliant.

pg9182:
On the other hand, GCC has non-insignificant issues and disadvantages when using it to cross-compile for (or natively compile on) Windows, including:

- Extra GCC support libraries.
- More compatibility stuff required.
- More complex setup in most cases.
- Lack of PDB support.
- No good VS integration.
- More complex cross-compilation setup.

Also see https://github.com/R2Northstar/NorthstarLauncher/pull/226. We did discuss this on Discord a while back.

apophis42:
> On the other hand, GCC has non-insignificant issues and disadvantages when using it to cross-compile for (or natively compile on) Windows, including:
> 
>     * Extra GCC support libraries.
> 
>     * More compatibility stuff required.
> 
>     * More complex setup in most cases.
> 
>     * Lack of PDB support.
> 
>     * No good VS integration.
> 
>     * More complex cross-compilation setup.
> 
> 
> Also see #226. We did discuss this on Discord a while back.

Those are all good points. This PR should be compatible with clang as well, so I think that the decision of which compiler to use would happen after this ticket, in the step about changing the CI configuration.

pg9182:
One thing I'd like to request, though, is that we avoid introducing any GCC-isms which are unnecessary with clang. I haven't read the diff yet, though. Also, ensure we aren't depending on undefined behaviour which only works with GCC and/or MSVC. I think you're already keeping this in mind, but I'll double-check when I do a review later.

Meson also builds happily with all 3 compilers and integrates with VS, and adoption of it is increasing rapidly, especially in open-source projects (I do concede that CMake still has the vast majority in professional contexts though). Nevertheless, for the record, I'm okay with making CMake the primary build system for Northstar, but I'll be maintaining Meson in parallel as well since it's a bit less finicky to work with in my experience.

apophis42:
> One thing I'd like to request, though, is that we avoid introducing any GCC-isms which are unnecessary with clang. I haven't read the diff yet, though. Also, ensure we aren't depending on undefined behaviour which only works with GCC and/or MSVC. I think you're already keeping this in mind, but I'll double-check when I do a review later.

Yeah, please point out anything like that.

> Meson also builds happily with all 3 compilers and integrates with VS, and adoption of it is increasing rapidly, especially in open-source projects (I do concede that CMake still has the vast majority in professional contexts though). Nevertheless, for the record, I'm okay with making CMake the primary build system for Northstar, but I'll be maintaining Meson in parallel as well since it's a bit less finicky to work with in my experience.

I was unaware Meson had support in VS. Do you mean that you can use the generated VS project (I assume Meson works similarly to CMake and generates a VS solution that it then builds with) or is their first-party integration? I'll absolutely agree that CMake is a pain to work with, although fortunately this project is a lot simpler than my work ones, so it's not too bad.

> Also, thanks for taking the initiative to get this all working; it probably wouldn't have been done for another half-year at least if you hadn't stepped up to do it.

NP, happy to help.

pg9182:
> I was unaware Meson had support in VS. Do you mean that you can use the generated VS project (I assume Meson works similarly to CMake and generates a VS solution that it then builds with) or is their first-party integration? I'll absolutely agree that CMake is a pain to work with, although fortunately this project is a lot simpler than my work ones, so it's not too bad.

Meson generates VS solutions (although I don't know how well it works in practice; I rarely use Visual Studio -- but I have heard good things about it). However, it isn't first-party integration, which is why I'm not as strongly against it for Northstar.

NorthstarDLL/printmaps.cpp
emma-miler: Ah i see, whoever wrote this specific function got the order of their keywords messed up. What i *assume* to be intended was `const char* const`. In its current form, this is indeed incorrect

apophis42: I agree, that probably was the intention. Would you like me to change it to `const char* const`?

emma-miler: That's probably the best course of action here. I have another PR waiting for some other stuff that fixes a whole bunch of little things like this, but that needs to wait a bit for when the toolchain shift is done

GeckoEidechse:
I'm assuming there's no interest to port newest main which has CMake already merged?

Mingw support would be interesting but we need to still primarily target MSVC cause that's what the game was built with.

The reasong for mingw support was being able to build Windows binaries of Northstar on Linux but that's basically resolved with #501

ASpoonPlaysGames:
Closing as stale and huge merge conflicts

Issue #347, opened 2022-11-27
HEAD of main fails to start on Wine in northstar-dedicated


pg9182:
**Issue 1 (bisected to 9e469ac28b610ecb8bce3e6c279660fae78861cf, #310):**

- Crashes on dedicated startup.
- Why?

**Issue 2 (presumably 2139e0ec6d95ed7c5f673b53dec71f39a013dd93):**

- Crash dumps don't work anymore.
- No output for crashes.
- No exit on crash.

**Issue 3 (blamed to bf09852285941e20a04731443d03c693e3f8ba2d, #320):**

- Title format on startup changed.
- This, combined with the other issues, means that even the watchdog can't catch the startup crash, and the server remains registered to the master server even though it's not actually working.

<details><summary><b>Logs</b></summary>

pg9182/northstar-dedicated@c8ce820ab4c12f182cafa19250fdfe9b7ad9ffec (v1.10.7) + R2Northstar/NorthstarLauncher@9e469ac28b610ecb8bce3e6c279660fae78861cf = [`ghcr.io/pg9182/northstar-dedicated-ci:1-tf2.0.11.0-ns1.10.7-ci.59de24a5.34.3555918725-nslauncher.9e469ac2.3510284696`](https://github.com/pg9182/northstar-dedicated-ci/actions/runs/3555918725)

```
ns-aitdm_dev_local-1  | 2022-11-27T00:06:45.819670097Z Northstar Dedicated Server - Docker (hostname: 719928e02de2)
ns-aitdm_dev_local-1  | 2022-11-27T00:06:45.819701535Z 
ns-aitdm_dev_local-1  | 2022-11-27T00:06:45.819705598Z     https://northstar.tf
ns-aitdm_dev_local-1  | 2022-11-27T00:06:45.819708407Z     https://northstar.tf/discord
ns-aitdm_dev_local-1  | 2022-11-27T00:06:45.819710953Z     https://github.com/R2Northstar/Northstar
ns-aitdm_dev_local-1  | 2022-11-27T00:06:45.819713445Z     https://github.com/pg9182/northstar-dedicated
ns-aitdm_dev_local-1  | 2022-11-27T00:06:45.819715866Z 
ns-aitdm_dev_local-1  | 2022-11-27T00:06:45.819718311Z Merging files...
ns-aitdm_dev_local-1  | 2022-11-27T00:06:45.819720718Z 
[...]

PR #352, opened 2022-12-01
Don't set initial dedicated server title to the profile name

ASpoonPlaysGames:
Should help with #347 apparently?
Could use testing with docker, but it doesn't set the title for normal dedis anymore.
image

### DIFF
--- a/NorthstarDLL/nsprefix.cpp
+++ b/NorthstarDLL/nsprefix.cpp
@@ -1,5 +1,6 @@
 #include "pch.h"
 #include "nsprefix.h"
+#include "dedicated.h"
 #include <string>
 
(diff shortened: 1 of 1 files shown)


pg9182 (approved):
Tested `ghcr.io/pg9182/northstar-dedicated-ci:1-tf2.0.11.0-ns1.10.7-ci.59de24a5.40.3594698427-nslauncher.c2e87b36.3594426638-nsmods.4730389c`.

GeckoEidechse (approved):
Works on Windows. Running `NorthstarLauncher.exe -dedicated -multiple -profile=test` the profile name is not in window title.

PR #354, opened 2022-12-01
Update exit value in hooks to valid code

emma-miler:
Apparently -1 is invalid


GeckoEidechse:
As hinted at by @pg9182 in https://github.com/R2Northstar/NorthstarLauncher/pull/353#discussion_r1037360833

PR #355, opened 2022-12-01
Log function lookup failure on Call

emma-miler:
Logs when failing to look up function on Squirrel manager call.


ASpoonPlaysGames (approved):
code looks good

PR #360, opened 2022-12-06
generally cleanup authentication code and fix client state issues with rejection from local server

BobTheBob9:
generally cleans up auth code, adds proper support for bots to auth, fixes issues with sp and rejects connections rather than accepting them then disconnecting
also patches client to ensure that listen server is properly closed after the local client is rejected from it, which is necessary if we want to reject on auth fail


GeckoEidechse (changes requested):
With this, instead of 
invalid username
I now get
auth failed

In both cases, trying to join multiplayer afterwards just results in an infinite loading screen.

GeckoEidechse (approved):
Confirmed working in testing. No more "_Invalid username_" or authentication failed when launching single player. Multiplayer auth still works fine.

PR #363, opened 2022-12-11
Remove launcher proxy

emma-miler:
Launcher proxy hasn't been shipped in over half a year. The only reason we kept it was for legacy reasons.
It was used to launch the game back when we were using the unpacked .exe, but not it is useless.


p0358:
Well it technically still should work and be a valid alternative, though I'm not sure if there were cases where wsock32 wouldn't work. But think something like WINE which could want to use built-in wsock32 DLL and you'd need to set up overrides to load a proxy. In such case a launcher proxy could become a compelling alternative that doesn't force you to alter wine prefix's settings...

emma-miler:
Ill talk to some people who use the Wine version, but i dont think that should be an issue.

p0358:
Of course in the end it's not that big of a deal to have separate wine prefix itself, but I'm not sure if it's guaranteed changing around DLL overrides settings won't ever cause some issues in some wine version or system configuration (idk). But if anything like that was to happen, this alternate proxy is a pretty easy drop-in replacement that could then help workaround/diagnose such issue.

So I'm personally not in favor of removing it, given it's not a lot of code, it's pretty simple* and a third-ish launch method isn't that bad considering there are already two. Admittedly though its usefulness also depends on the standard NorthstarLauncher.exe approach not working in the first place either, but at the same time with EA App that becomes more prevalent too lately.

\* This simplicity and like 90% shared code could even warrant making it part of the wsock proxy code and have it built as launcher proxy with some defines, but I think that could in the end just make things more complicated and harder to understand/maintain.......

p0358:
In favor of removal is that it'd probably be still in the git history just in case, I guess.

But at least there should be no issues porting this one to be compiled with clang as-is though, because it basically just loads two DLLs...

emma-miler:
We could always put it on the tools repo and mark it as archived, that way we will always have a public backup in case we ever need it again.

BobTheBob9 (approved):
doesn't seem needed in really any case atm

emma-miler:
Archived at https://github.com/R2NorthstarTools/LauncherProxy

PR #364, opened 2022-12-11
Make most mod-setting related funcs global.

Erlite:
This PR makes mod related functions such as `NSGetModVersionByModName()` global, as they previously were only accessible in UI contexts.

Example:
```lua
Spyglass_SayAll(format("Initializing core, version %s.", NSGetModVersionByModName("IMC.Spyglass")));
```
image


BobTheBob9 (approved):
absolutely 0 issues here, these only access globally available northstar stuff so not potential for issues really

PR #366, opened 2022-12-11
Add NSGetModName

emma-miler:
Add several functions that allow scripts and native to get the mod they are in/being called from.


GeckoEidechse:
To add some context, this is useful for proposed keyvalue store so that we can have key value store per mod, where one mod cannot access values of another mod.

emma-miler:
This, but also just generally useful for generic/library-like mods

NorthstarDLL/squirrel/squirrel.cpp
```diff
@@ -549,6 +549,25 @@ template <ScriptContext context> void SquirrelManager<context>::ProcessMessageBu
 	_call(m_pSQVM->sqvm, message.args.size());
 }
 
+ADD_SQFUNC(
+	"string",
+	NSGetModName,
```
BobTheBob9, line 554: should call this NSGetCallingModName or something similar to indicate that it comes from stack rather than like, a parameter

BobTheBob9 (approved):
looks good mostly

BobTheBob9 (approved):

GeckoEidechse:
Was this merged without testing? ._.

PR #343, opened 2022-11-24
Plugin system v2

emma-miler:
Still a little bit work in progress. Feel free to review and comment. Will depend on a few more PR's across multiple repos

PR #375, opened 2022-12-19
Expose Cursor Position to sq vm

uniboi:
UI can do cool things with the cursor position.

```cpp
array<int> NSGetCursorPosition()
```
returns position as [x,y] in pixel relative to the game window origin.

Depends on https://github.com/R2Northstar/NorthstarLauncher/pull/343


Erlite:
Might want to change the branch to main, makes it hard to figure out your changes over the others.

uniboi:
literally click the link https://github.com/R2Northstar/NorthstarLauncher/pull/375/commits/d628d87d91e44c8c18412c5ac32f1dc21ec8fd42

Erlite:
I'm memeing, you did the same to my HTTP PR.

Code looks good to me, can probably be merged once Emma's PR is in.

RoyalBlue1:
why not return a struct with 2 ints? you need to add the struct in the init.nut though

ASpoonPlaysGames:
Whats the interaction with users who are tabbed out of the game? Want to avoid ui clicks "through" other windows

uniboi:
> why not return a struct with 2 ints? you need to add the struct in the init.nut though

because a) I forgor that's possible b) my c++ experience is limited to Hello World and c) GetScreenSize() gives an array back as well

> Whats the interaction with users who are tabbed out of the game? Want to avoid ui clicks "through" other windows

positions are relative to the game window origin, so if you're on your second position your x position might be 2000 for example

ASpoonPlaysGames:
> positions are relative to the game window origin, so if you're on your second position your x position might be 2000 for example

That's not really what i mean. Windows can be overlayed on top of each other so there is a chance that when using a second window, the position could still be being used (when it shouldn't really)

ASpoonPlaysGames:
image
for example, when im tabbed into file explorer, Apex shouldn't be getting my cursor position and returning a value that apex cares about

uniboi:
yeah no I don't check if the game is in focus or whatever, I just get the cursor position

ASpoonPlaysGames:
Probably should return a null value of some kind when the game isn't in focus, some people use virtual keyboards and stuff so tracking mouse positions all the time could be seen as a security vulnerability

emma-miler:
Rebase on current main, cherrypick the changes you need, and make a new PR. Also, return a vector instaed

PR #376, opened 2022-12-21
Add NSDisconnectPlayer() function.

Erlite:
This PR adds the following function:

```lua
bool function NSDisconnectPlayer(entity player, string reason);
```

This uses the native `CBaseClient__Disconnect()` function to ensure the player is booted from the server, while still giving them a disconnect reason much like the `disconnect` command.

Available on server VM.


BobTheBob9 (approved):
might be good to add a player.Disconnect function in squirrel to correspond to this, but that's for another pr

Issue #378, opened 2022-12-22
In-game console does not show when using toggle_console


ASpoonPlaysGames:
Seems to be caused by #365

From my limited attempts to find the source of the issue, `(*g_pSourceGameConsole)->Activate();` does not seem to be working?

ASpoonPlaysGames:
Solved by #379

PR #379, opened 2022-12-22
Fix console not enabling

emma-miler:
Issue was caused by `sourceinterface.cpp` not being in the filters, so wasn't compiled :/


BobTheBob9 (approved):

GeckoEidechse:
(Closes #378)

PR #381, opened 2022-12-25
Process all squirrel messages in the buffer instead of one per frame.

Erlite:
This PR simply makes the squirrel manager process all queued messages from native to Squirrel scripts every frame, instead of the previous behaviour which only processed one.

This is going to be needed for my next PR which adds a callback to Squirrel scripts for when the VM is about to be destroyed.


Erlite:
Tested using HTTP requests, no issue whatsoever, doesn't really do anything that can break anyway.

emma-miler (approved):

PR #382, opened 2022-12-25
Add missing breaks to AsyncCall_External()

Erlite:
Simply fixes a few missing `break`s in `AsyncCall_External()` leading to weird behaviour.


catornot:
lmao
thats why I was getting functions called on 2 vms

emma-miler (approved):

PR #383, opened 2022-12-25
Add Destroy callback for squirrel scripts.

Erlite:
This PR adds a new callback to configure in a mod's `mod.json` file. The `Destroy` callback will be executed right before the VM of the given context is destroyed:

```json
{
    "Path": "spyglass/client_globals.nut",
    "RunOn": "CLIENT && MP",
    "ClientCallback":
    {
        "After": "Spyglass_ClientGlobalsInit",
        "Destroy": "Spyglass_ClientGlobalsDestroy"
    }
}
```
```lua
void function Spyglass_ClientGlobalsDestroy()
{
    printt("hello i get destroyed oh no")
}
```

The above example would give the following result:

image

### DIFF
--- a/NorthstarDLL/squirrel/squirrel.cpp
+++ b/NorthstarDLL/squirrel/squirrel.cpp
@@ -1,5 +1,6 @@
 #include "pch.h"
 #include "squirrel.h"
+#include "logging/logging.h"
 #include "core/convar/concommand.h"
 #include "mods/modmanager.h"
 #include "dedicated/dedicated.h"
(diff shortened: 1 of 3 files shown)


NorthstarDLL/squirrel/squirrel.cpp
```diff
@@ -209,7 +210,47 @@ template <ScriptContext context> void SquirrelManager<context>::VMCreated(CSquir
...
+					{
+						NS::log::squirrel_logger<context>()->error(
+							"VMDestroyed() was unable to find function with name '{}'. Is it global?", callback.DestroyCallback);
+						continue;
+					}
+
+					pushobject(m_pSQVM->sqvm, &functionobj);
+					pushroottable(m_pSQVM->sqvm);
+
+					_call(m_pSQVM->sqvm, 0);
```
emma-miler, line 241: Why do you manually use `_call` here, as opposed to using `Call`?

Erlite: I just copied code from AsyncCall() tbh heh. I'll fix it.

BobTheBob9:
Waiting on changes mentioned in Emma's previous comment

BobTheBob9 (approved):
compiles and works fine in testing

PR #386, opened 2023-01-02
add FCVAR_REPLICATED to ns_server_name

catornot:
Tested, works correctly. Although it's not getting reset after leaving a server (idk how to fix this).

### DIFF
--- a/NorthstarDLL/server/serverpresence.cpp
+++ b/NorthstarDLL/server/serverpresence.cpp
@@ -79,7 +79,7 @@ ServerPresenceManager::ServerPresenceManager()
 	Cvar_ns_server_presence_update_rate = new ConVar(
 		"ns_server_presence_update_rate", "5000", FCVAR_GAMEDLL, "How often we update our server's presence on server lists in ms");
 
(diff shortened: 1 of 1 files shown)


F1F7Y:
> Although it's not getting reset after leaving a server (idk how to fix this).

That's exactly how FCVAR_REPLICATED works though.

catornot:
I am making proximity chat mod which is supposed to 100% server side. to differentiate servers apart I need some id, In this case its the server's name. if `ns_server_name` doesn't have `FCVAR_REPLICATED` flag the clients wouldn't know the name of the server. 

this would be useful for this [pr](https://github.com/R2Northstar/NorthstarMods/pull/536) in Northstar mods

BobTheBob9:
> Can you gimme a quick ELI5 why we want that?
> 
> (My convar knowledge is minimal and giving a reason for the change helps me better understand it ^^)

just lets clients know the name of the server they're currently on

BobTheBob9 (changes requested):
not much reason not to replicate description as well

catornot:
now `ns_server_desc` as FCVAR_REPLICATED too

BobTheBob9 (approved):

PR #387, opened 2023-01-03
Expose Cursor Position to UI vm

uniboi:
exposes a function to retrieve the current cursor position to UI
```js
vector ornull NSGetCursorPosition()
```
returns `null` if the game currently isn't focused, and a vector containing the cursor position otherwise.

All dependencies are resolved.


NorthstarDLL/scripts/client/cursorposition.cpp
```diff
@@ -0,0 +1,18 @@
+#include "pch.h"
+#include "squirrel/squirrel.h"
+#include "util/wininfo.h"
+
+ADD_SQFUNC("vector ornull", NSGetCursorPosition, "", "", ScriptContext::UI)
+{
+	POINT p;
+	if (GetCursorPos(&p) && ScreenToClient(*g_gameHWND, &p))
```
BobTheBob9, line 8: is there much benefit using winapi to get this rather than finding it using a game func of some kind? game probably keeps track of this too

BobTheBob9 (changes requested):
doesn't build atm, is util/wininfo.h not included?

NorthstarDLL/scripts/client/cursorposition.cpp
uniboi: a) idk how to decompile the game
b) idk how to call native functions using offsets
c) I have 0 c++ experience

uniboi:
I included wininfo.h in this pr to maybe speed things up.

BobTheBob9:
don't see any issues with this at a glance, will review once tested

uniboi:
I tested with the workflow artifact on windows and linux with this script and it works as intended:
```
void function cursortest()
{
    while( 1 )
    {
        printt(NSGetCursorPosition())
        wait 1
    }
}
```

BobTheBob9 (changes requested):
Largely fine and works well in testing, only issue is that cursor position isn't clamped to window size when in windowed mode (i.e. cursor position can be either negative, or above window size), unsure if this would be an issue in practice? Marking as request changes for now but open to this not being required

BobTheBob9:
if needed i can pr the code for this, shouldn't be too hard to implement?

BobTheBob9 (approved):

PR #388, opened 2023-01-04
Improve response to UI VM compile errors

ASpoonPlaysGames:
Now should open the console automatically when UI VM hits a compile error, could still use some work I think, but this now works.

Also fixes a bug in `IsUIVM` which was the cause of the above not working properly to begin with, thanks to @RoyalBlue1 

Also improves logging around compile errors a bit, logging the response (or lack thereof) to the compile 

Users should now see something like this when they boot the game with a UI compile error, as opposed to the blank screen
image

### DIFF
--- a/NorthstarDLL/squirrel/squirrel.cpp
+++ b/NorthstarDLL/squirrel/squirrel.cpp
@@ -376,14 +376,14 @@ void __fastcall ScriptCompileErrorHook(HSquirrelVM* sqvm, const char* error, con
 		// kill dedicated server if we hit this
 		if (IsDedicatedServer())
 		{
-			logger->error("Exiting dedicated server due to fatal compile error");
+			logger->error("Exiting dedicated server, compile error is fatal");
 			// flush the logger before we exit so debug things get saved to log file
 			logger->flush();
 			exit(EXIT_FAILURE);
 		}
 		else
 		{
-			logger->error("Disconnecting due to fatal compile error");
+			logger->error("Disconnecting, compile error is fatal");
 
 			R2::Cbuf_AddText(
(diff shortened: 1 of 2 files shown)


NorthstarDLL/squirrel/squirrel.cpp
```diff
@@ -376,12 +375,15 @@ void __fastcall ScriptCompileErrorHook(HSquirrelVM* sqvm, const char* error, con
...
 		if (IsDedicatedServer())
 		{
+			logger->error("Exiting dedicated server, compile error is fatal");
 			// flush the logger before we exit so debug things get saved to log file
 			logger->flush();
 			exit(EXIT_FAILURE);
 		}
 		else
 		{
+			logger->error("Disconnecting, compile error is fatal");
```
BobTheBob9, line 385: possibly remove these? not much point and might hide the actual error in logs

NorthstarDLL/squirrel/squirrel.cpp
```diff
@@ -395,6 +397,10 @@ void __fastcall ScriptCompileErrorHook(HSquirrelVM* sqvm, const char* error, con
 				R2::Cbuf_AddText(R2::Cbuf_GetCurrentPlayer(), "showconsole", R2::cmd_source_t::kCommandSrcCode);
 		}
 	}
+	else
+	{
+		logger->warn("Not disconnecting, compile error is non-fatal");
```
BobTheBob9, line 402: possibly remove these? not much point and might hide the actual error in logs

BobTheBob9 (changes requested):

BobTheBob9 (approved):
works in testing

PR #389, opened 2023-01-05
Restrict chat message charset

pg9182:


GeckoEidechse (approved):
Confirmed working in testing

PR #390, opened 2023-01-06
Revert "add FCVAR_REPLICATED to ns_server_name"

GeckoEidechse:
Reverts R2Northstar/NorthstarLauncher#386

Current code causes this to be a breaking change for joining players. We need to first adjust current code base to make this non-breaking


GeckoEidechse:
@catornot for info

BobTheBob9 (approved):

PR #391, opened 2023-01-06
Fix issues of restriction of chat message charset

p0358:
Fixes regressions introduced with #389 (https://github.com/R2Northstar/NorthstarLauncher/pull/389/commits/4df475857755bbc0fddd72c34c0ecbc10db95e90).

Previous implementation could cause crashes due to null pointer dereference and removed all UTF-8 characters.

Current implementation fixes aforementioned null pointer dereference, preserves valid UTF-8 sequences while removing invalid ones (invalid ones won't crash). Furthermore, duplicated code was moved out to a common file.

Also allows `:` as suggested by cpdt analogous to here: https://github.com/R2Northstar/NorthstarLauncher/blob/main/NorthstarDLL/client/localchatwriter.cpp#L414

### DIFF
--- a/NorthstarDLL/NorthstarDLL.vcxproj
+++ b/NorthstarDLL/NorthstarDLL.vcxproj
@@ -445,6 +445,7 @@
     <ClInclude Include="squirrel\squirrelautobind.h" />
     <ClInclude Include="squirrel\squirrelclasstypes.h" />
     <ClInclude Include="squirrel\squirreldatatypes.h" />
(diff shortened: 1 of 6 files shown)


GeckoEidechse (approved):
Works in testing: Still strips `BEL` but keeps `üäö` which previous version would strip.

NoCatt (approved):
The German "Umlaute" (ä,ö,ü) seem to work again.
Chat colours crashing the client also are fixed.

PR #392, opened 2023-01-06
Pedantic fix for #391 (use if-else instead of if)

p0358:
So that me and @GeckoEidechse can calmly sleep knowing it was fixed xD

These ifs are mutually exclusive hence the fix (we check for byte 1).

image


ASpoonPlaysGames (approved):
I mean not much to review here, but yeah

BobTheBob9 (approved):

PR #393, opened 2023-01-06
Force-include `pch.h` to all files, remove explicit includes

p0358:
Just so nobody needs to remember to include it when creating new files. Forced inclusion makes sense, given it **has** to always be included by default, otherwise compiler error is thrown.

It should be safe to do as even CMake does this by default if you use precompiled headers (search for `/FI` on the page): https://cmake.org/cmake/help/v3.16/command/target_precompile_headers.html

Visual Studio's intellisense should recognize this with zero issues automatically, I am unsure about other IDEs however. VS is still the primary one though.

Third commit fixes an issue where header defined the field as 32 while cpp was in 256. Somehow only my changes above uncovered this issue...


NorthstarDLL/util/version.h
```diff
@@ -1,6 +1,6 @@
 #pragma once
 
 extern char version[16];
-extern char NSUserAgent[32];
+extern char NSUserAgent[256];
```
ASpoonPlaysGames, line 4: What's up with this? Doesn't seem particularly relevant to this PR?

p0358: It's defined at 256 bytes long in the relevant .cpp file, which meant this definition here was wrong. It should have thrown a compile error before, but somehow didn't and nobody noticed. And my PR's changes made it complain that 32 here mismatched 256 in the other place...

ASpoonPlaysGames: Makes sense, fair enough

ASpoonPlaysGames (approved):

BobTheBob9 (approved):
looks good, may need to merge main before merging though

PR #394, opened 2023-01-06
Fix overriding individual localisation strings being weird

BobTheBob9:
Change how we load localisation to fix overriding individual localisation strings being weird.
With the old localisation loading code, our custom value in `Northstar.Custom` for the string `GAMEMODE_FW` wouldn't load, this fixes that by changing where we load our localisation to be more consistent


GeckoEidechse (approved):
Confirmed working in testing. Both `script_client print(Localize("#GAMEMODE_FW"))` and the loading screen for Frontier War correctly say "Frontier War"

image

PR #398, opened 2023-01-13
Execute cfgs before server startup for gamemode being run

BobTheBob9:
Allows mods to create a cfg for the initialisation and cleanup of gamemodes, to be run on server before the server initialises. This is mainly to setup engine vars that need to be set before server init (i.e. the network convars required for FW to work correctly)
to use, create a `setup_gamemode_{gamemode}.cfg` and `cleanup_gamemode_{gamemode}.cfg` file in `mod/cfg/server`, and they will be run automatically on server startup/destruction.

Also contains utility commands primarily for these cfgs, mainly `cvar_reset`, that resets a cvar to its default value

### DIFF
--- a/NorthstarDLL/core/convar/concommand.cpp
+++ b/NorthstarDLL/core/convar/concommand.cpp
@@ -1,8 +1,11 @@
 #include "pch.h"
 #include "concommand.h"
 #include "shared/misccommands.h"
+#include "engine/r2engine.h"
 
 #include <iostream>
+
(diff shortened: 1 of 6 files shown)


GeckoEidechse:
> [...] create a `setup_gamemode_{gamemode}.cfg` and `cleanup_gamemode_{gamemode}.cfg` file in `mod/cfg/server`, and they will be run automatically on server startup/destruction.

sample config file names:
```
setup_gamemode_fw.cfg
cleanup_gamemode_fw.cfg
```

RoyalBlue1:
Code looks good

PR #358, opened 2022-12-02
Retarget to use CMake+Clang/LLVM

emma-miler:
This will need updating of CI before it can build properly.
Will invalidate basically all non-trivial PRs. I'd be happy to help update ones that want to update to the new format.

PR #399, opened 2023-01-16
Build on linux with mingw32-w64 

pYtato-KK:
Changes for GCC, Makefiles for automated building.

Things that don't work:
- Game freezes just before main menu
- Static linking libraries, despite explicitly supplying the linker with `-l:lib*.dll.a`
    - libcurl.dll and its dependencies must be copied to the Titanfall 2 install directory
    - DLLs have to be copied from /usr/x86_64-w64-mingw32/bin

Info on some changes:
- MinHook included as a git submodule, as I had to recompile it with MinGW
    - MinHook has had MinGW support since v1.3.2
    - Linking failed with included MinHook, possibly because included MinHook wasn't built with MinGW support
- Minidumps are disabled for mingw, as I couldn't find a proper replacement

Improvements to do:
- Use CMake (merge #358)
- Get past game freeze
- Minidump replacement
- Change code so that preprocessor checks for mingw aren't needed anymore
- Static linking of curl and winpthreads

# Build host information
Build requirements:
- mingw32-w64 with posix threads
- libcurl (or use libcurl from include dir, which I didn't bother trying)

I built on Gentoo with mingw32-w64.
Gentoo can be used on any UNIX system with [Gentoo prefix](https://wiki.gentoo.org/wiki/Prefix), though it could take some time to compile.

MinGW was emerged with crossdev and has to be built with the posix thread model. Instructions for this can be found at  [Gentoo wiki](https://wiki.gentoo.org/wiki/Mingw#x86_64_toolchain). After MinGW is installed curl has to be emerged with `USE="static-libs" x86_64-w64-mingw32-emerge curl`. Curl won't build without the `static-libs` USE flag.

# Debugging
As the current linux build freezes just before main menu, some debugging is required. 

Proton can dump scripts for debugging with winedbg with `PROTON_DUMP_DEBUG_COMMANDS=1`. Scrips are dumped to `/tmp/proton_$USER`.
The run scripts fail to launch Titanfall, and using the `gdb_attach` script causes gdb to error with `Invalid hex digit 59`

Any info on debugging with gdb and winedbg would be appreciated, @ me on discord `something funny#0916`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,7 @@ bld/
 [Oo]bj/
 [Ll]og/
 [Ll]ogs/
+linuxbuild/
 
 # Visual Studio 2015/2017 cache/options directory
 .vs/
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "NorthstarDLL/minhook"]
+	path = NorthstarDLL/minhook
+	url = https://github.com/TsudaKageyu/minhook.git
--- a/Makefile
+++ b/Makefile
@@ -2,14 +2,16 @@
 BUILD_DIR := linuxbuild/
 MAKEOPTS :=
 
-debug: MAKEOPTS += debug
-debug: all
-
 
 all: $(BUILD_DIR)/NorthstarLauncher.exe $(BUILD_DIR)/Northstar.dll
 	mkdir -p $(BUILD_DIR)
 	cp NorthstarDLL/$(BUILD_DIR)/Northstar.dll $(BUILD_DIR)
+	cp NorthstarDLL/$(BUILD_DIR)/linuxbuild/MinHook.dll $(BUILD_DIR)
 	cp NorthstarLauncher/$(BUILD_DIR)/NorthstarLauncher.exe $(BUILD_DIR)
+
+
+debug: MAKEOPTS += debug
+debug: all
 
 
 $(BUILD_DIR)/NorthstarLauncher.exe:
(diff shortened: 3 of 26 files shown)


pYtato-KK:
Northstar.dll built by GitHub CI does not freeze, the game and reaches main menu.

pYtato-KK:
DLL loading errors also happens when compiled with `-static`. More info on static linking with MinGW needed.

pYtato-KK:
Attaching gdb to Titanfall2.exe:
1. Install a MinGW compatible gdb (emerge x86_64-w64-mingw-gdb)
2. Make proton dump debug scripts with `PROTON_DUMP_DEBUG_COMMANDS=1`
3. Change line 5 in the `gdb_attach` script to `EXE_NAME=${1:-"Titanfall2.exe"}`
4. `export WINE_GDB=x86_64-w64-mingw-gdb` (or whatever the name of your gdb is)
5. Run Titanfall 2
6. Run the `gdb_attach` script
I think that should be all that is required to start debugging

pg9182:
@pYtato-KK, we'll be using https://github.com/Jake-Shadle/xwin MSVCRT rather than mingw as the sysroot, and almost definitely clang-cl as the compiler.

GeckoEidechse:
The main reason we were looking into building with compilers other than MSVC was to also be able to compile on Linux but that has largely been solved with #531 

While being able to compile Northstar with compilers other than MSVC is still something we could look into in order to rely less on undefined behaviour from MSVC, we should probably still primarily use MSVC for actual releases as it's the same compiler Titanfall2 was compiled with.

I'm closing this PR as stale but feel free to re-open if you want to work on this again <3

Issue #396, opened 2023-01-09
Mods system rework


Alystrasz:
Notes from Bob:
* mods need to pull from 2 folders, local (user-installed) and remote (auto-installed)
* local should be down to user preference (though, will need to be disabled on server join if not compatible with server we're joining), remote should be disabled until required server join, then enabled, then disabled again on leave
* we need to move more of the server required mod loading/unloading to native as well, loading is currently in script and unloading doesn't quite exist yet, but both should be more native
* a couple more assets also can't currently be reloaded at all at runtime well, not super required for autodownloading to work at all, but specifically being able to properly unload/reload weapon/playersettings/etc scripts would be very nice, and potentially models too though this seems very unlikely since it doesn't like valve source can do it

https://discord.com/channels/920776187884732556/1025040519560167514/1060355733981503609

---

###### Directories architecture

- [x] Use mods from local and remote directories (#403)
- [x] Put auto-downloaded mods into remote mods directory (#545)
- [x] Use thunderstore mod string as folder name so that we can have different versions at the same time (#535)

###### Dynamic mod activation

- [x] Add new mods entries to `enabledmods.json` if needed (#410)
- [ ] Enable mod only when joining server
- [ ] Disable mod when leaving server
- [ ] Dynamically reload mods' assets (weapons/models/playersettings)

###### Miscellaneous

- [ ] Display a warning message (on UI) with malformed mods (#415)

Alystrasz:
Work on mod system has been split between #746 and https://github.com/R2Northstar/NorthstarLauncher/pulls?q=is%3Aopen+is%3Apr+label%3AMAD, closing this.

PR #403, opened 2023-01-26
feat: Remote mods directory

Alystrasz:
This pull request introduces remote mods folder, located at `/runtime/remote/mods`, that will store future auto-installed mods.

In short, it's the *"mods need to pull from 2 folders, local (user-installed) and remote (auto-installed)"* mentioned by @BobTheBob9 in https://github.com/R2Northstar/NorthstarLauncher/issues/396.

##### Changes

* Creates remote mod directory if it does not exist;
* When loading mods, check local folder AND remote mod folder

*(note: there was already a `REMOTE_MOD_FOLDER_SUFFIX` variable in code, I simply added some stuff around to use it)*


BobTheBob9 (approved):
looks great and shouldn't create any different behaviour between the different mod dirs due to how we store mods at runtime

PR #404, opened 2023-01-28
Safe I/O 2 - Electric Boogaloo!

### LINKED ISSUE #11: Allow Mods to use Northstar to read/write to files within their respective directories
This would allow mods to save data, allowing for better banlists, and other functions. Since there doesn't seem to be a way to do this due to Respawn's modification of Squirrel, the easiest option is probably to add the functions here similarly to how functions like NSSetModEnabled work

EladNLG:
Holy shit! #223 but actually working??? what the fuck???? 

Must be merged with
- https://github.com/R2Northstar/NorthstarMods/pull/595
- https://github.com/R2Northstar/NorthstarLauncher/pull/443

# Safe I/O
Allows mods to write and save files (save location: `<profile>/saves/<mod directory name>`).

Adds many safety precautions to prevent malicious use, even if a malicious actor gains complete access to the Squirrel VM, to minimize damage (e.g. ~~a 50,000 character limit per file, a forced JSON file extension,~~ and disallowing access to files above the mod's save directory.)

* The character limit and forced JSON file extension have been removed. 

"like you *could* fill an HDD with files but like<br>ehh<br>it seems like a niche issue that you'd like<br>notice" -BobTheBob

(In addition, you'd immediately know which mod is problematic, since the saveData is separated into the mod folders)

Also, since the data is no longer necessarily in JSON (and strings cannot contain null characters anyways, so it's not like you can write e.g. an executable), the JSON extension is no longer enforced.

## Documentation

Squirrel Functions:
```js
void function NSSaveFile( string file, string data ) // writes the table as JSON to the specified file. File has to be defined in the mod.json of your mod for this to work.
string function NSLoadFile( string file ) // reads the specified JSON file and returns it as a table. File has to be defined in the mod.json of your mod for this to work.
```

Use `NSSaveFile` and `NSLoadFile` to save and load text files. The data persists between restarts, so this can be useful for saving stuff you want the client/server to save **locally,** like how many kills the player has gained over time.

### Example:

```js
global function Mod_Init
global function MutePlayer

table mutedPlayersData

void function Mod_Init()
{
    mutedPlayersData = DecodeJSON( NSLoadFile( "EladNLG.MuteMod", "eladNLG_muteMod_mutedPlayers" ) )
    AddClientCommandCallback( "mute", MutePlayer )
}

void function MutePlayer( entity player, array<string> args )
{
    if (!PlayerAllowedToMute(player))
        return false;
    if (args.len() < 1)
        return true;
    entity playerToMute = GetPlayerByName(args[0])
    
    // mute the player
    if (!("playerList" in mutedPlayersData))
        mutedPlayersData.playerList <- {}
        
    mutedPlayersData.playerList[playerToMute.GetUID()] <- true
    
    // save the file
    NSSaveFile( "muted_players", EncodeJSON( mutedPlayersData ) )
}
```

## Use cases

- Making a client-side stats mod that saves it's data locally, with stuff like match history.
- TF/Roguelike, namely saving armor instances, achievements, etc.

It's basically better as a "medium-size database(?)" Stuff which doesn't fit in a ConVar, but doesn't warrant a full database.

## Reasons

Using ConVars for saving stuff, *especially* settings, just doesn't work. 

It resets each time you boot up vanilla, the length of the ConVar contents is very limited, and it does not scale well AT ALL if you want to save data on multiple players (for example, a mod that mutes players. After muting about ~40 players, the ConVar will not save all of the muted players, and some will be discarded - and though the same problem would exist with this - this would only happen after ~100,000 players, which is way more than you'd practically need. AND if we wish, we can raise the limit even higher.)

## TODO

- Re-add JSON saving as separate functions (Namely NSSaveJSONFile & NSLoadJSONFile)
- Code cleanup, may be messy I think

***

Closes #11


ASpoonPlaysGames:
Diffs seem kinda fucked? like the project filters are weird and stuff. Is this branch based off of another branch?

EladNLG:
> Diffs seem kinda fucked? like the project filters are weird and stuff. Is this branch based off of another branch?

no. visual studio sucks. working on fixing this rn.

GeckoEidechse:
I don't think I'm fully up-to-date, did Emma's thing about detecting which mod a Squirrel script is from get merged yet? ^^

Cause last time we had the discussion about safe I/O I remember it boiling down to that safe I/O should be a key-value store where each NS mod can only access their own store.

GeckoEidechse:
Oh yeah, turns out it was merged #366

EladNLG:
> I don't think I'm fully up-to-date, did Emma's thing about detecting which mod a Squirrel script is from get merged yet? ^^
> 
> Cause last time we had the discussion about safe I/O I remember it boiling down to that safe I/O should be a key-value store where each NS mod can only access their own store.

Yeah.

It already auto-detects the current mod and stores accordingly (the mod param is gone).

GeckoEidechse:
> It already auto-detects the current mod and stores accordingly (the mod param is gone).

Awesome :D

Is there a reason we still allow for the mod naming the store file? At least the way I see it is that if we limit it to a single file, we could get rid of all the logic around ensuring correct file path and specifying allowed files `mod.json`. ^^

EladNLG:
> > It already auto-detects the current mod and stores accordingly (the mod param is gone).
> 
> Awesome :D
> 
> Is there a reason we still allow for the mod naming the store file? At least the way I see it is that if we limit it to a single file, we could get rid of all the logic around ensuring correct file path and specifying allowed files `mod.json`. ^^

Shit to use. Imagine having to parse a giant file, it would be a nightmare to program. Additionally, if a mod saves a lot of data, each time it would load/save that data would hitch the game for a long while (since the ENTIRE FILE has to be rewritten, and doing that in a separate thread is something beyond my skill level, not to mention it would, in my opinion, over-complicate the process of using Safe I/O). Allowing mods to separate files means that they can spread these load times over a longer period, since instead of having 1 50MB file, you can load 10 5MB files separately, as needed

NorthstarLauncher/NorthstarLauncher.vcxproj
```diff
@@ -46,6 +46,7 @@
   <PropertyGroup Label="UserMacros" />
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
     <LinkIncremental>true</LinkIncremental>
+    <OutDir>..\..\..\..\..\..\Program Files %28x86%29\Steam\steamapps\common\Titanfall2</OutDir>
```
GeckoEidechse, line 49: Uhm

EladNLG: Will fix tomorrow

BobTheBob9:
do you happen to have a simple mod that could be used to test this?

GeckoEidechse:
There was a discussion on Discord about this PR the other day but given that it was in `#general` it's kinda difficult to find it. To recap, my stance on safe I/O is that it should just be a simple key value store API, where modders can call like `persistent_write(key, value)` and `persistent_read(key)` or something along the lines.

The key-value store would be isolated per mod so that mod authors won't have to deal with accidentally using same key as another mod.

Key value store in turn would be able to hold various data types including tables which would allow for nesting datastructures.


The main argument that was made for allowing mod author to specify destination file was that it would be difficult to parse a single large file, however (at least to me) it seems like it shouldn't really make a difference whether you use

```js
void function NSSaveFile( string file, string data )
string function NSLoadFile( string file )
```
with
```json
achievements.json
{
  "achievement1": true,
  "achievement2": false
}

armor.json
[
  {
    "chestplate": {
      "durability": 5,
      "title": "your mom"
    },
  },
  {
    "boots": {
      "durability": 3,
      "extras": [
        "triple jump",
        "idk"
      ]
    }
  }
]
```

or e.g.
```js
void function NSSaveKey( string key, string data )
string function NSLoadKey( string key )
```
with
```json
Author_Mod.json
{
  "achievements": {
    "achievement1": true,
    "achievement2": false
  },
  "armor": [
    {
      "chestplate": {
        "durability": 5,
        "title": "your mom"
      }
    },
    {
      "boots": {
        "durability": 3,
        "extras": [
          "triple jump",
          "idk"
        ]
      }
    }
  ]
}
```
shouldn't matter to the code at least...

BobTheBob9:
> There was a discussion on Discord about this PR the other day but given that it was in `#general` it's kinda difficult to find it. To recap, my stance on safe I/O is that it should just be a simple key value store API, where modders can call like `persistent_write(key, value)` and `persistent_read(key)` or something along the lines.
> 
> The key-value store would be isolated per mod so that mod authors won't have to deal with accidentally using same key as another mod.
> 
> Key value store in turn would be able to hold various data types including tables which would allow for nesting datastructures.
> 
> The main argument that was made for allowing mod author to specify destination file was that it would be difficult to parse a single large file, however (at least to me) it seems like it shouldn't really make a difference whether you use
> 
> ```js
> void function NSSaveFile( string file, string data )
> string function NSLoadFile( string file )
> ```
> 
> with
> 
> ```json
> achievements.json
> {
>   "achievement1": true,
>   "achievement2": false
> }
> 
> armor.json
> [
>   {
>     "chestplate": {
>       "durability": 5,
>       "title": "your mom"
>     },
>   },
>   {
>     "boots": {
>       "durability": 3,
>       "extras": [
>         "triple jump",
>         "idk"
>       ]
>     }
>   }
> ]
> ```
> 
> or e.g.
> 
> ```js
> void function NSSaveKey( string key, string data )
> string function NSLoadKey( string key )
> ```
> 
> with
> 
> ```json
> Author_Mod.json
> {
>   "achievements": {
>     "achievement1": true,
>     "achievement2": false
>   },
>   "armor": [
>     {
>       "chestplate": {
>         "durability": 5,
>         "title": "your mom"
>       }
>     },
>     {
>       "boots": {
>         "durability": 3,
>         "extras": [
>           "triple jump",
>           "idk"
>         ]
>       }
>     }
>   ]
> }
> ```
> 
> shouldn't matter to the code at least...

definitely disagree imo, this kills any potential for modders to output to preexisting file formats (non-json), directory structures, importing files from other things etc, and would be a pain for things like per-player information stores for instance, where having to save and load everything into one file would be terrible for performance, readability etc

BobTheBob9:
if modders want to store everything in one json file they can *choose* to do that, that doesn't mean restricting them to json only or single file only or whatever, limiting things like that just seems like a bad idea overall when one of the core purposes of this project is to make the game super moddable, imo

GeckoEidechse:
> definitely disagree imo, this kills any potential for modders to output to preexisting file formats (non-json), directory structures, importing files from other things etc,

I haven't seen a use-case for non-JSON yet. Writing a custom parser in Squirrel seems like effort to me ngl. And like most people will use Squirrel Tables <-> JSON, so a lot of mods will just have the same duplicated code of

```
load string from file
convert to Squirrel table
use table
...
use table
convert to JSON
save string to file
```

so might as well combine loading and (de)serializing to Squirrel object into a single step :P

Importing files from other mods is probably the biggest argument against single per mod key value store, though the way I see it is that in terms of backwards compatibility on API change, it's always easier to ease restrictions than it is to tighten them.

So should we find that restrictions are to tight, we can always reduce them later.

> [...] and would be a pain for things like per-player information stores for instance, where having to save and load everything into one file would be terrible for performance, readability etc

This is also a good point and something I discussed on Discord but forgot to mention here. Basically my train of thought was to have single simple API for modders. How stuff is actually persisted is up to NortharLauncher to decide (this also helps with "upgrading mods" from within Northstar should our API change, e.g. we change location where we store files).

So in terms of performance, it would then be up to Northstar to support efficient implementations by e.g. sharding the JSON over multiple files etc. In that regard we are probably limited by whatever RapidJSON provides. As for readability, in general persistent store of a mod shouldn't really be accessed by a human user anyway...

&nbsp;

So basically to summarise why I'm more in favour of a per-mod key-value store is
1. It's always easier to lift restrictions in future PR then it is to tighten them
2. It makes it harder for mod-author to fuck up (no dealing with defining files in `mod.json`, no additional conversion step)
3. Lifts all the burden of needing to create an efficient implementation from mod author and puts it on NorthstarLauncher instead
4. Make life for modder easy by giving them dead simple API and handle the rest ourselves.

But again, that's just my take ^^

BobTheBob9:
> so might as well combine loading and (de)serializing to Squirrel object into a single step :P 
we can have a helper/wrapper function for this but like, loading and deserializing is still a one liner, it's not like we're taking off much boilerplate by making everything json

BobTheBob9:
> So in terms of performance, it would then be up to Northstar to support efficient implementations by e.g. sharding the JSON over multiple files etc. In that regard we are probably limited by whatever RapidJSON provides. As for readability, in general persistent store of a mod shouldn't really be accessed by a human user anyway...

at this point you're making your internal implementation overcomplex, modders are programmers, they can handle writing to different files i'm sure and in most cases performance i'm sure *won't* be a concern for modders since they won't be saving much in files anyway, but in the cases where modders are saving alot of data, it's much easier for us to take a hands off approach and let them deal with it, and they'll probably be able to get better performance than a one-size-fits all solution anyway

BobTheBob9:
tbh i'd be in favour of removing the restriction of having mods have to define all their files explicitly and just giving them a directory they can create whatever files they want in, would cover the most usecases (especially in the case of playerdata, which i imagine this will definitely be used for and should be a usecase we try to design around)

EladNLG:
I'm just gonna post my general opinion on forced JSON:

The problem with forced JSON is that

1) the code for converting to a string from a table is a confusing mess that cannot be easily understood without explicit explanation, at least in my opinion
2) it forces people into a certain format (e.g. by forcing JSON, we would be unable to import Icepick gauntlets easily, since they use a custom parser and format) 
3) the same problem happens with forcing mods to use a single file - people would have to *manually* make a valid JSON with all of their gauntlets.

Even worse, let's say we make a Forge-style editor - where save sharing is fundamental - that just became a whole lot harder.

Also, let's say someone's save got corrupted - it'll be much harder for a modder to look through and find the issue when there's one single file - since it's harder to read one large file than multiple smaller ones.

As of right now (I forgot to update the doumentation), there are 2 versions of each function:

`NSSaveFile` / `NSLoadFile` (takes string for data)
& 
`NSSaveJSONFile` / `NSLoadJSONFile` (takes a table for data, auto-converts to/from JSON)

These should cover *all possible use cases.*

I think that the JSON versions could be converted into squirrel helper functions if the code is too confusing - though string memory could be a problem then maybe - so I'd rather to keep the JSON version in C++.

GeckoEidechse:
I think what we are seeing here is two different usecases that would best be covered by two different implementations.

I.e. basic persistent store using keyvalues. This is your stuff like, saving list of admin UIDs in FSU or your custom HUD from HUD revamp (maybe even editable via modsettings? :eyes:)

and then there's your stuff like you mentioned with Forge-clone and gauntlet speedrun stuff which certainly needs individual file read/write.

EladNLG:
> I think what we are seeing here is two different usecases that would best be covered by two different implementations.
> 
> I.e. basic persistent store using keyvalues. This is your stuff like, saving list of admin UIDs in FSU or your custom HUD from HUD revamp (maybe even editable via modsettings? 👀)
> 
> and then there's your stuff like you mentioned with Forge-clone and gauntlet speedrun stuff which certainly needs individual file read/write.

which is why we can (and tbh, *should*, @GeckoEidechse ) make both functions available. Everyone has their own use case - we should be able to support it.

NorthstarDLL/mods/modmanager.cpp
```diff
@@ -264,6 +265,96 @@ Mod::Mod(fs::path modDir, char* jsonBuf)
...
+		for (int i = 0; i < modJson["SaveFiles"].Size(); i++)
+		{
+			// MAX CHARACTERS PER MOD CHECK
+			// likely could be removed?
+			// hardcoded and not a startup arg because this limit is kinda ridiculous anyways.
+			// done to prevent a mod from pasting hundreds of file names to fill up the hard drive.
+			// this also means mods have to actually tell how many characters they're going to use instead of just typing in "9999999999".
+			// we check every time to prevent overflow from decieving us into thinking the character flow is low when it's actually just
+			// overflowed.
+			if (TotalCharLimit > 2000000)
```
BobTheBob9, line 286: > The character limit and forced JSON file extension have been removed.  

isn't it still here?

EladNLG: I forgot to push. Will do later today :P

EladNLG: Done.

NorthstarDLL/mods/modsavefiles.cpp
```diff
@@ -0,0 +1,232 @@
...
+	rapidjson_document doc;
+	doc.Parse(content);
+
+	// basic parse checking shit
+	if (doc.HasParseError())
+	{
+		g_pSquirrel<context>->newtable(sqvm);
+
+		std::string sErrorString = fmt::format(
+			"Failed parsing json file: encountered parse error \"{}\" at offset {}",
```
EladNLG, line 219: Should include file name probably.

EladNLG (commented):
TODO stuff

NorthstarDLL/mods/modsavefiles.cpp
```diff
@@ -0,0 +1,232 @@
...
+	if (ContainsNonASCIIChars(content))
+	{
+		g_pSquirrel<context>->raiseerror(
+			sqvm, fmt::format("File contents may not contain non-ASCII characters! Make sure your strings are valid!", mod->Name).c_str());
+		return SQRESULT_ERROR;
+	}
+
+	fs::create_directories(fs::path(GetNorthstarPrefix()) / "saveData" / fs::path(mod->m_ModDirectory).filename());
+	std::ofstream fileStr(fs::path(GetNorthstarPrefix()) / "saveData" / fs::path(mod->m_ModDirectory).filename() / (fileName));
+	if (fileStr.fail())
```
ASpoonPlaysGames, line 104: This will need proper testing on different Operating Systems, to ensure reserved names cant be used

NorthstarDLL/mods/modsavefiles.cpp
```diff
@@ -0,0 +1,232 @@
...
+	std::string fileName = g_pSquirrel<context>->getstring(sqvm, 1);
+	if (CheckFileName(fileName))
+	{
+		g_pSquirrel<context>->raiseerror(
+			sqvm,
+			fmt::format("File name invalid ({})! Make sure it has no '\\', '/' or non-ASCII charcters!", fileName, mod->Name).c_str());
+		return SQRESULT_ERROR;
+	}
+
+	std::ifstream fileStr(fs::path(GetNorthstarPrefix()) / "saveData" / fs::path(mod->m_ModDirectory).filename() / (fileName));
```
ASpoonPlaysGames, line 166: Given how often the path `fs::path(GetNorthstarPrefix()) / "saveData" / fs::path(mod->m_ModDirectory)` is used, it might be worth making it into a header variable or something so that we don't have to deal with re-evaluating it all the time?

Not a big deal though, this shouldn't get hit very often

NorthstarDLL/mods/modsavefiles.cpp
```diff
@@ -0,0 +1,232 @@
...
```
ASpoonPlaysGames, line 133: I personally would prefer the directory to be called `savedata` or something instead of `saveData`

NorthstarDLL/mods/modsavefiles.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+	std::stringstream jsonStringStream;
+	while (fileStr.peek() != EOF)
+		jsonStringStream << (char)fileStr.get();
+	g_pSquirrel<context>->pushstring(sqvm, jsonStringStream.str().c_str());
+
+	return SQRESULT_NOTNULL;
+}
+
+ADD_SQFUNC("table", NSLoadJSONFile, "string file", "", ScriptContext::CLIENT | ScriptContext::UI | ScriptContext::SERVER)
```
ASpoonPlaysGames, line 148: Ill mention this here just as a note.

Json files do not have to be an object (`table`).
They can be just an array, which we don't (and can't) handle atm because we return a `table` instead of a `var`

This usecase of JSON is perfectly valid, and Juicy (who I cant seem to @ ) has already ran into this problem grabbing data from speedrun.com

If anyone else is looking for the solution to this in the future, (in script) you can use string concatenation to encapsulate the JSON array in an object and then you can parse it into a table with the normal JSON functions
```
string jsonArray = "[\"foo\", \"bar\"]" // result from elsewhere, currently cannot be parsed into a table
jsonArray = "{ \"data\": " + jsonArray + "}" // encapsulate the array in an object
// the array data can now be parsed and used
```

This should be fixed, but should probably be done in a separate PR

EladNLG: Might be able to return arrays, but encoding them will be harder to do.

For the sake of consistency, I'm keeping this table-only and resolving this, though if anyone has any idea on how to approach this. I'm down.

NorthstarDLL/mods/modsavefiles.cpp
EladNLG: i only use windows, linux gamers pls

NorthstarDLL/mods/modsavefiles.cpp
ASpoonPlaysGames: Made into a new issue, can be resolved here just fine.

NorthstarDLL/mods/modsavefiles.cpp
```diff
@@ -0,0 +1,235 @@
...
+	if (CheckFileName(fileName))
+	{
+		g_pSquirrel<context>->raiseerror(
+			sqvm,
+			fmt::format("File name invalid ({})! Make sure it has no '\\', '/' or non-ASCII charcters!", fileName, mod->Name).c_str());
+		return SQRESULT_ERROR;
+	}
+
+	std::string content = g_pSquirrel<context>->getstring(sqvm, 2);
+	if (ContainsNonASCIIChars(content))
```
Alystrasz, line 56: Any particular reason as to disallow using UTF8 characters, for instance?

Alystrasz, line 56: Any particular reason as to disallow using UTF8 characters, for instance?

EladNLG: Looking further into this, UTF-8 should work. 

I'd rather err on the side of caution, though (can't really be sure if writing an executable/potentially malicious file with UTF-8 is possible)

This is the same reason NUL characters are banned.

Alystrasz: I was looking for use-cases requiring this: I thought about storing player names, however I don't think EA accounts allow using special characters ([maybe not in all countries?](https://northstar.cool/assets/NorthstarPromoPoster.png))
But in the end you're right, better be safe than sorry! Authorizing non-ASCII chars can easily be done in a PR later.

EladNLG: > I don't think EA accounts allow using special characters

Regardless, you should be storing UIDs in that case, not usernames.

NorthstarDLL/mods/modsavefiles.cpp
EladNLG: By the looks of this - this does not write to reserved file names.

For linux - afaik - there are no reserved file names.

EladNLG:
## ***Updated Documentation:***


```js
void function NSSaveFile( string file, string data )
```
Saves a file. `file` must be a valid ***file name***, and not access any directories.
Overwrites if the file exists already - otherwise creates it.
**Throws an error if:**
- The file could not be created/written to.
- The file name is invalid.
- The function was called not from a file belonging to a mod.
- The file contents contained non-ASCII characters.

```js
string function NSLoadFile( string file )
```
Loads a file and returns it's contents. `file` must be a valid ***file name***, and not access any directories.
**Throws an error if:**
- The file could not be read from.
- The file name is invalid.
- The function was called not from a file belonging to a mod.

```js
void function NSSaveJSONFile( string file, table data )
```
Converts a table into JSON and then saves it. `file` must be a valid ***file name***, and not access any directories.
Overwrites if the file exists already - otherwise creates it.

```js
table function NSLoadJSONFile( string file )
```
Loads a JSON file, and converts it into a table. `file` must be a valid ***file name***, and not access any directories.
**Throws an error if:**
- The file could not be read from.
- The file name is invalid.
- The file could not be converted into a table (The contents are not valid JSON).
- The function was called not from a file belonging to a mod.

```js
bool function NSDeleteFile( string file )
```
Deletes a file. `file` must be a valid ***file name***, and not access any directories.
Returns `true` if the file was deleted successfully, `false` otherwise.
**Throws an error if:**
- The file name is invalid.

```js
bool function NSDoesFileExist( string file )
```
Loads a file.
Returns `true` if the file loaded successfully, `false` otherwise.
**Throws an error if:**
- The file name is invalid.

EladNLG:
Idk how large to make the size cap

(Note: size cap is not hard, we really don't mind if mods go like a few MB over, they'll just not be able to write anything until they go back below the cap.

EladNLG:
I think that we need to decide on a size limit for mods, now that that's implemented

I'm thinking somewhere between the 10MB to 250MB range.

any ideas?

F1F7Y:
> I think that we need to decide on a size limit for mods, now that that's implemented

Size limit on the entire mod or just safe i/o ?

Asking because adding a size limit on the entire mod will heavily limit map mods with custom textures and models.

EladNLG:
> > I think that we need to decide on a size limit for mods, now that that's implemented
> 
> Size limit on the entire mod or just safe i/o ?
> 
> Asking because adding a size limit on the entire mod will heavily limit map mods with custom textures and models.

Just safe i/o data.

Limiting mods would be imo stupid.

EladNLG:
(Man I need to get better at wording shit)

F1F7Y:
> any ideas?

I'd say somewhere around 10MB. If you ever go above that with a json file you're probably rockstar and should seek help.

EladNLG:
After internal discussion:

## A folder limit of 50MB was decided. 

### This limit is per ***mod***, not per file.

Though might be unnecessary (said by BobTheBob), I think it wouldn't hurt (and he doesn't mind).

## The PR, is now, functionality-wise - done. 

Unless there's some big refactor, there shouldn't be a need for changes.

EladNLG:
Must be merged with [NorthstarMods #595](https://github.com/R2Northstar/NorthstarMods/pull/595).

## File saving is now asynchronous.

No more game hitching :D

EladNLG:
# Testing!

For your convenience - a mod to test Safe I/O quickly and easily.

Do note, for squirrel-really-sucks reasons, that the game will hitch for a while during testing.

This should only take ~15s tho.

## how 🤔 
1. compile the PR
2. Paste [EladNLG.SafeIOTest.zip](https://github.com/R2Northstar/NorthstarLauncher/files/10831732/EladNLG.SafeIOTest.zip) into your mods folder
4. Boot up the game, and run: `script_ui thread TestSafeIO()`

NorthstarDLL/mods/modsavefiles.cpp
GeckoEidechse: Quickly unresolving for now cause I wanna get @wolf109909's opinion in regards to NorthstarCN ^^

GeckoEidechse (commented):
Initial test following https://github.com/R2Northstar/NorthstarLauncher/pull/404#issuecomment-1445156296 and it saves the data to `Titanfall2/R2Northstar/save_data/EladNLG.SafeIOTest/` as expected.

This was just on local client. No testing done on server or similar yet.

NorthstarDLL/mods/modsavefiles.cpp
```diff
@@ -0,0 +1,410 @@
...
+{
+	Mod* mod = g_pSquirrel<context>->getcallingmod(sqvm);
+	if (mod == nullptr)
+	{
+		g_pSquirrel<context>->raiseerror(sqvm, "Has to be called from a mod function!");
+		return SQRESULT_ERROR;
+	}
+
+	std::string fileName = g_pSquirrel<context>->getstring(sqvm, 1);
+	if (CheckFileName(fileName))
```
Alystrasz, line 189: You should check here that the path does not contain any `..`:

image

GeckoEidechse (changes requested):
@Alystrasz managed to break folder restriction



by basically just doing `../`

(ping him for more info, I'm just relaying what was said in Discord channel here)

NorthstarDLL/mods/modsavefiles.cpp
EladNLG: Wait, but how did you do it without using / ?

Alystrasz (changes requested):
Security issue there, mods shouldn't be able to create files outside mod data directory.

NorthstarDLL/mods/modsavefiles.cpp
GeckoEidechse: (I was a second too late with making new review xD)

EladNLG:
Sorry for the technical difficulties, I am bad at git.

Thank god this is getting squashed anyways,

NorthstarDLL/mods/modsavefiles.cpp
EladNLG: Quiiiiiiiiiiiickly resolving because we have decided there is no real security benefit for only allowing ASCII characters.

EladNLG:
# Testing!
(update)

For your convenience - a mod to test Safe I/O quickly and easily.

Do note, for squirrel-really-sucks reasons, that the game will hitch for a while during testing.

This should only take ~15s tho.

## how 🤔 

1. Compile the PR or [download build from CI](https://github.com/R2Northstar/NorthstarLauncher/suites/11211552363/artifacts/573528554)
2. Paste [EladNLG.SafeIOTest.zip](https://github.com/R2Northstar/NorthstarLauncher/files/10837029/EladNLG.SafeIOTest.2.zip) into your mods folder
4. Boot up the game, and run: `script_ui thread TestSafeIO()`

EladNLG:
```diff

-		g_pSquirrel<context>->pushinteger(
-			sqvm,
-			(int)(fs::file_size(dir / fileName) / 1024)); // throws if file does not exist
+		g_pSquirrel<context>->pushinteger(sqvm,
+										  (int)(fs::file_size(dir / fileName) / 1024)); // throws if file does not exist
```

I'm not doing this @GeckoEidechse

wolf109909:
> ```diff
> -		g_pSquirrel<context>->pushinteger(
> -			sqvm,
> -			(int)(fs::file_size(dir / fileName) / 1024)); // throws if file does not exist
> +		g_pSquirrel<context>->pushinteger(sqvm,
> +										  (int)(fs::file_size(dir / fileName) / 1024)); // throws if file does not exist
> ```
> 
> I'm not doing this @GeckoEidechse

no way ...

GeckoEidechse:
> ```diff
> -		g_pSquirrel<context>->pushinteger(
> -			sqvm,
> -			(int)(fs::file_size(dir / fileName) / 1024)); // throws if file does not exist
> +		g_pSquirrel<context>->pushinteger(sqvm,
> +										  (int)(fs::file_size(dir / fileName) / 1024)); // throws if file does not exist
> ```
> 
> I'm not doing this @GeckoEidechse

Just do

```
// clang-format off
THE LINES WITH ISSUE HERE
// clang-format on
```
then

EladNLG:
already fixed

I am just complaining about clang-format

GeckoEidechse:
skipping code review for this PR cause I don't feel qualified to do proper code review on this + I still don't agree with the design decision (which is totally fine btw, enough other contributors seem to be fine with it ^^)

EladNLG:
> I still don't agree with the design decision (which is totally fine btw, enough other contributors seem to be fine with it ^^)

Which design decision?

GeckoEidechse:
> > I still don't agree with the design decision (which is totally fine btw, enough other contributors seem to be fine with it ^^)
> 
> Which design decision?

Full file access vs key-value store. Again I'm just one person's opinion and others are fine and/or prefer the current implementation, so PR will be merged upon approval accordingly regardless of my opinion :P

EladNLG:
Another small thing

thinking about removing the native SaveJSON func and replacing it with a wrapper instead for consistency in code.

It's faster to keep it in native, i guess, but since we can't decode tables asynchronously (accessing SQTables async is a bad idea), it means the code could stay inconsistent

would love opinions about this

EladNLG:
Additionally, I wonder if letting mods access subfolders is a good idea, since after the incident Safe I/O no longer checks if the file being accessed is in a subfolder.

This could be something we support, meaning:
- Small modifications to allow checking subfolders (namely `NSGetAllFiles`)
- Potentially an issue with @BobTheBob9's plans to put convar cfgs in /save_data/Author.Mod/cfg/
- no real use for this(?)

I don't see any real disadvantage to this otherwise?

ASpoonPlaysGames:
> * Potentially an issue with @BobTheBob9's plans to put convar cfgs in /save_data/Author.Mod/cfg/
Nah, if a mod wants to write a cfg to their folder, who are we to stop them? There isnt rly much point in them doing it, but I see no reason to stop them.
If a mod is writing to other mod's cfgs and breaking stuff then its simply a malicious mod and should just not be installed.

As far as I'm concerned the "safe" bit of safe IO is protecting the system, not other mods.

> * no real use for this(?)
Access to subfolders is always a nice thing to have, for organisational reasons

EladNLG:
> If a mod is writing to other mod's cfgs and breaking stuff then its simply a malicious mod and should just not be installed.

this should be impossible - every mod is contained within it's own folder :)

ASpoonPlaysGames:
> this should be impossible - every mod is contained within it's own folder :)

Sounds good then

NorthstarDLL/mods/modsavefiles.cpp
```diff
@@ -0,0 +1,464 @@
...
+	deleteThread.detach();
+}
+
+bool ContainsInvalidChars(std::string str)
+{
+	// we don't allow null characters either, even if they're ASCII characters because idk if people can
+	// use it to circumvent the file extension suffix.
+	return std::any_of(str.begin(), str.end(), [](char c) { return c == '\0'; });
+}
+bool CheckFileName(fs::path str, fs::path dir)
```
Alystrasz, line 163: This returns `true` if file name is NOT correct, right?
I find this a little counter-intuitive, I wouldn't mind some documentation describing this behaviour.

NorthstarDLL/mods/modsavefiles.cpp
```diff
@@ -0,0 +1,464 @@
...
+		return SQRESULT_ERROR;
+	}
+
+	fs::path dir = savePath / fs::path(mod->m_ModDirectory).filename();
+	std::string fileName = g_pSquirrel<context>->getstring(sqvm, 1);
+	if (CheckFileName(dir / fileName, dir))
+	{
+		g_pSquirrel<context>->raiseerror(
+			sqvm,
+			fmt::format("File name invalid ({})! Make sure it has no '\\', '/' or non-ASCII charcters!", fileName, mod->Name).c_str());
```
Alystrasz, line 189: Error message is incorrect, this is raised when file path is not correct (outside mods folder), and using non-ASCII characters isn't a valid reason for raising an error anymore.

NorthstarDLL/mods/modsavefiles.cpp
```diff
@@ -0,0 +1,464 @@
...
+		return SQRESULT_ERROR;
+	}
+
+	std::string content = g_pSquirrel<context>->getstring(sqvm, 2);
+	if (ContainsInvalidChars(content))
+	{
+		g_pSquirrel<context>->raiseerror(
+			sqvm, fmt::format("File contents may not contain non-ASCII characters! Make sure your strings are valid!", mod->Name).c_str());
```
Alystrasz, line 197: Same as above remark (wrong error message).

NorthstarDLL/mods/modsavefiles.cpp
```diff
@@ -0,0 +1,464 @@
...
+		return SQRESULT_ERROR;
+	}
+
+	fs::path dir = savePath / fs::path(mod->m_ModDirectory).filename();
+	std::string fileName = g_pSquirrel<context>->getstring(sqvm, 1);
+	if (CheckFileName(dir / fileName, dir))
+	{
+		g_pSquirrel<context>->raiseerror(
+			sqvm,
+			fmt::format("File name invalid ({})! Make sure it has no '\\', '/' or non-ASCII charcters!", fileName, mod->Name).c_str());
```
Alystrasz, line 241: Same as above remark (wrong error message).

NorthstarDLL/mods/modsavefiles.cpp
```diff
@@ -0,0 +1,464 @@
...
+		return SQRESULT_ERROR;
+	}
+
+	std::string content = EncodeJSON<context>(sqvm);
+	if (ContainsInvalidChars(content))
+	{
+		g_pSquirrel<context>->raiseerror(
+			sqvm, fmt::format("File contents may not contain non-ASCII characters! Make sure your strings are valid!", mod->Name).c_str());
```
Alystrasz, line 249: Same as above remark (wrong error message).

NorthstarDLL/mods/modsavefiles.cpp
```diff
@@ -0,0 +1,464 @@
...
+		return SQRESULT_ERROR;
+	}
+
+	fs::path dir = savePath / fs::path(mod->m_ModDirectory).filename();
+	std::string fileName = g_pSquirrel<context>->getstring(sqvm, 1);
+	if (CheckFileName(dir / fileName, dir))
+	{
+		g_pSquirrel<context>->raiseerror(
+			sqvm,
+			fmt::format("File name invalid ({})! Make sure it has no '\\', '/' or non-ASCII charcters!", fileName, mod->Name).c_str());
```
Alystrasz, line 291: Same as above remark (wrong error message).

NorthstarDLL/mods/modsavefiles.cpp
```diff
@@ -0,0 +1,464 @@
...
+{
+	Mod* mod = g_pSquirrel<context>->getcallingmod(sqvm);
+
+	fs::path dir = savePath / fs::path(mod->m_ModDirectory).filename();
+	std::string fileName = g_pSquirrel<context>->getstring(sqvm, 1);
+	if (CheckFileName(dir / fileName, dir))
+	{
+		g_pSquirrel<context>->raiseerror(
+			sqvm,
+			fmt::format("File name invalid ({})! Make sure it has no '\\', '/' or non-ASCII charcters!", fileName, mod->Name).c_str());
```
Alystrasz, line 311: Same as above remark (wrong error message).

NorthstarDLL/mods/modsavefiles.cpp
```diff
@@ -0,0 +1,464 @@
...
```
Alystrasz, line 330: Same as above remark (wrong error message).

NorthstarDLL/mods/modsavefiles.cpp
```diff
@@ -0,0 +1,464 @@
...
```
Alystrasz, line 359: Same as above remark (wrong error message).

NorthstarDLL/mods/modsavefiles.cpp
```diff
@@ -0,0 +1,464 @@
...
+	Mod* mod = g_pSquirrel<context>->getcallingmod(sqvm, 1);
+	fs::path dir = savePath / fs::path(mod->m_ModDirectory).filename();
+	std::string pathStr = g_pSquirrel<context>->getstring(sqvm, 1);
+	fs::path path = dir;
+	if (pathStr != "")
+		path = dir / pathStr;
+	if (CheckFileName(path, dir))
+	{
+		g_pSquirrel<context>->raiseerror(
+			sqvm, fmt::format("File name invalid ({})! Make sure it has no '\\', '/' or non-ASCII charcters!", pathStr, mod->Name).c_str());
```
Alystrasz, line 380: Same as above remark (wrong error message).

NorthstarDLL/mods/modsavefiles.cpp
```diff
@@ -0,0 +1,464 @@
...
+	Mod* mod = g_pSquirrel<context>->getcallingmod(sqvm);
+	fs::path dir = savePath / fs::path(mod->m_ModDirectory).filename();
+	std::string pathStr = g_pSquirrel<context>->getstring(sqvm, 1);
+	fs::path path = dir;
+	if (pathStr != "")
+		path = dir / pathStr;
+	if (CheckFileName(path, dir))
+	{
+		g_pSquirrel<context>->raiseerror(
+			sqvm, fmt::format("File name invalid ({})! Make sure it has no '\\', '/' or non-ASCII charcters!", pathStr, mod->Name).c_str());
```
Alystrasz, line 412: Same as above remark (wrong error message).

NorthstarDLL/mods/modsavefiles.cpp
```diff
@@ -0,0 +1,464 @@
...
+				// this ain't a cloud service.
+				if (GetSizeOfFolderContentsMinusFile(dir, file.filename().string()) + contents.length() > MAX_FOLDER_SIZE)
+				{
+					// tbh, you're either trying to fill the hard drive or use so much data, you SHOULD be congratulated.
+					spdlog::error(fmt::format("Mod spamming save requests? Folder limit bypassed despite previous checks. Not saving."));
+					mutex.get().unlock();
+					return;
+				}
+
+				std::ofstream fileStr(file);
```
Alystrasz, line 54: I don't know where to put this, but file name encoding seems to fail with non-ASCII characters, as you can see in this screenshot: 

image
Related squirrel code:
```javascript
NSSaveFile( "héllo.json", "héy" )
```

Note that loading the file (*i.e. `NSLoadFile( "héllo.json", [...])`) works correctly with non-ASCII characters.

EladNLG: I dunno how to fix this

If anyone knows you're welcome to unresolve this

EladNLG: Looks like a wrong encoding issue, but I don't think you can exactly change the encoding for the path?

Alystrasz: Why did you resolve this if the issue isn't fixed?
Maybe there are some streams that handle non-ASCII file names.

EladNLG: Because I've no plans to fix it.

The file name may be wrong, but it decodes correctly, so I see no priority in fixing it.

Additionally, any research I have done on the issue was fruitless, so I don't see any point in trying to fix this at all ¯\_(ツ)_/¯

EladNLG: I could make file names ASCII only to resolve this as well, but there's no real reason to do that outside of this bug.

EladNLG:
# UPDATE:

If you receive a `Has to be called from a mod function!` error whilst calling the functions from your own mod, `callingmodfunc` seems to be broken, and that seems to **not** be a fault of this PR :(

ASpoonPlaysGames (commented):
I would say that this is close to being ready for a merge.

The peepeepoopoo man has been testing this while developing TitanFramework, and they have not had any issues so far (that are not simply skill issues)

NorthstarDLL/mods/modsavefiles.cpp
```diff
@@ -0,0 +1,474 @@
...
+			try
+			{
+				mutex.get().lock();
+
+				std::ifstream fileStr(file);
+				if (fileStr.fail())
+				{
+					spdlog::error("A file was supposed to be loaded but we can't access it?!");
+					// we should throw a script error here. But idk how lmao
+					g_pSquirrel<context>->AsyncCall("NSHandleLoadResult", handle, "");
```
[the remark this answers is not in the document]
EladNLG: added optional onfailure callback if the read fails.

calls onsuccess with empty string/table if not used and read fails otherwise (for those who'd rather that)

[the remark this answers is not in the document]
ASpoonPlaysGames: just making a note that this is handled in script now

ASpoonPlaysGames (approved):
code looks alright, and works in testing according to peepeepoopooman

NorthstarDLL/mods/modsavefiles.cpp
```diff
@@ -0,0 +1,483 @@
...
+		return SQRESULT_ERROR;
+	}
+
+	fs::path dir = savePath / fs::path(mod->m_ModDirectory).filename();
+	std::string fileName = g_pSquirrel<context>->getstring(sqvm, 1);
+	if (!IsPathSafe(fileName, dir))
+	{
+		g_pSquirrel<context>->raiseerror(
+			sqvm,
+			fmt::format("File name invalid ({})! Make sure it results in a path inside your mod's save folder.", fileName, mod->Name)
```
Alystrasz, line 302: Another reason for this to be raised is the usage of non-ASCII characters, I think it should be mentioned here:
```suggestion
			fmt::format("File name invalid ({})! Make sure it does not contain any non-ASCII character, and results in a path inside your mod's save folder.", fileName, mod->Name)
```

NorthstarDLL/mods/modsavefiles.cpp
```diff
@@ -0,0 +1,483 @@
...
+	fs::path dir = savePath / fs::path(mod->m_ModDirectory).filename();
+	std::string pathStr = g_pSquirrel<context>->getstring(sqvm, 1);
+	fs::path path = dir;
+	if (pathStr != "")
+		path = dir / pathStr;
+	if (!IsPathSafe(pathStr, dir))
+	{
+		g_pSquirrel<context>->raiseerror(
+			sqvm,
+			fmt::format("File name invalid ({})! Make sure it results in a path inside your mod's save folder.", pathStr, mod->Name)
```
Alystrasz, line 430: Another reason for this to be raised is the usage of non-ASCII characters, I think it should be mentioned here:
```suggestion
			fmt::format("File name invalid ({})! Make sure it does not contain any non-ASCII character, and results in a path inside your mod's save folder.", pathStr, mod->Name)
```

Alystrasz (changes requested):
Error messages need to be reworked since ASCII characters aren't allowed in file names anymore.

NorthstarDLL/mods/modsavefiles.cpp
```diff
@@ -0,0 +1,483 @@
...
+				.c_str());
+		return SQRESULT_ERROR;
+	}
+
+	g_saveFileManager->SaveFileAsync<context>(dir / fileName, content);
+
+	return SQRESULT_NULL;
+}
+
+#include "scripts/scriptjson.h"
```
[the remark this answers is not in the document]
EladNLG: Fixed. Previously was here because it caused an error otherwise, but it no longer does. (ugh)

NorthstarDLL/mods/modsavefiles.cpp
```diff
@@ -0,0 +1,483 @@
...
+// I don't want to know why I have to pass the string path as a separate parameter
+// but if I don't ASCII check fails. So here we are.
+bool IsPathSafe(const std::string param, fs::path dir)
+{
+	auto const normRoot = fs::weakly_canonical(dir);
+	auto const normChild = fs::weakly_canonical(dir / param);
+
+	auto itr = std::search(normChild.begin(), normChild.end(), normRoot.begin(), normRoot.end());
+	// we return if the file is safe (inside the directory) and uses only ASCII chars in the path.
+	// clang-format off
```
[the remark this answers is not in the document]
EladNLG: Fixed.

Alystrasz (approved):
Files cannot be created outside mod folder or using non-ASCII characters anymore.
Code looks good to me.

EladNLG:
we do be waiting tho 🕔

F1F7Y (commented):
Will test later this weekend. Dont really want to sign off on the code as my cpp isnt as good as bobbers or pgs

NorthstarDLL/mods/modsavefiles.cpp
```diff
@@ -0,0 +1,515 @@
...
+	doc.Accept(writer);
+
+	// return the converted string
+	return buffer.GetString();
+}
+
+ON_DLL_LOAD("engine.dll", ModSaveFFiles_Init, (CModule module))
+{
+	savePath = fs::path(GetNorthstarPrefix()) / "save_data";
+	g_saveFileManager = new SaveFileManager;
```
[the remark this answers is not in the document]
EladNLG: seeing as in other instances where the instance is never deleted they still use a regular pointer, I think I'll change the name, but keep it as is to keep consistency with the rest of the codebase.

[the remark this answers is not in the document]
EladNLG: also, since we only assign to `g_pSaveFileManager` _once_, when engine.dll gets loaded - unique_ptr provides no benefit.

F1F7Y (approved):
Works in testing, needs #443 and [#595](https://github.com/R2Northstar/NorthstarMods/pull/595)

EladNLG:
i sure hope this doesn't fuck up :)

GeckoEidechse:
Kinda want someone with decent C++ experience like @BobTheBob9 or @pg9182 or similar to look at it

Alystrasz:
> Kinda want someone with decent C++ experience like @BobTheBob9 or @pg9182 or similar to look at it ^^

image

\>:(

GeckoEidechse (commented):
Skimming over it in 2 seconds I didn't notice a mention of where Safe I/O files are stored. Maybe mention it in the PR description, so that whoever does ModdingDocs PR can just copy/paste it ^^

EladNLG:
> Skimming over it in 2 seconds I didn't notice a mention of where Safe I/O files are stored. Maybe mention it in the PR description, so that whoever does ModdingDocs PR can just copy/paste it ^^

was already talked about - they're saved in `<profile>/saves/<mod directory>`.

NorthstarDLL/mods/modsavefiles.cpp
```diff
@@ -0,0 +1,515 @@
...
+				mod->Name)
+				.c_str());
+		return SQRESULT_ERROR;
+	}
+
+	g_pSaveFileManager->DeleteFileAsync<context>(dir / fileName);
+	return SQRESULT_NOTNULL;
+}
+
+// The param is not optional because that causes issues :)
```
GeckoEidechse, line 406: What issues?

NorthstarDLL/mods/modsavefiles.cpp
```diff
@@ -0,0 +1,515 @@
...
+		return SQRESULT_ERROR;
+	}
+}
+
+// ok, I'm just gonna explain what the fuck is going on here because this
+// is the pinnacle of my stupidity and I do not want to touch this ever
+// again & someone will eventually have to maintain this.
+
+// P.S. if you don't want me to do the entire table -> JSON -> string thing...
+// Fix it in scriptjson first k thx bye
```
GeckoEidechse, line 485: Wasn't that fix merged already?

NorthstarDLL/mods/modsavefiles.cpp
```diff
@@ -0,0 +1,515 @@
...
+	{
+		g_pSquirrel<context>->raiseerror(
+			sqvm, fmt::format("File contents may not contain NUL/\\0 characters! Make sure your strings are valid!", mod->Name).c_str());
+		return SQRESULT_ERROR;
+	}
+
+	fs::create_directories(dir);
+	// this actually allows mods to go over the limit, but not by much
+	// the limit is to prevent mods from taking gigabytes of space,
+	// this ain't a cloud service.
```
GeckoEidechse, line 275: Ditto

NorthstarDLL/mods/modsavefiles.cpp
```diff
@@ -0,0 +1,515 @@
...
+		return SQRESULT_ERROR;
+	}
+
+	fs::create_directories((dir / fileName).parent_path());
+	// this actually allows mods to go over the limit, but not by much
+	// the limit is to prevent mods from taking gigabytes of space,
+	// this ain't a cloud service.
+	if (GetSizeOfFolderContentsMinusFile(dir, fileName) + content.length() > MAX_FOLDER_SIZE)
+	{
+		// tbh, you're either trying to fill the hard drive or use so much data, you SHOULD be congratulated.
```
GeckoEidechse, line 223: I'd swap sarcastic comments with what's actually happening / **why** an error is thrown.

The **what** can usually be inferred from code (although additional comments can help, keep in mind that a lot of first time contributors lack coding knowledge so a few additional comments don't hurt), so comments should be used to cover the **why** part.

GeckoEidechse (commented):
Just a bunch of things I noticed while skimming through the code.

Given that I lack C++ knowledge this is mostly bikeshedding/general stuff.

NorthstarDLL/mods/modsavefiles.cpp
EladNLG: that fix was never made. it's a completely different thing.

NorthstarDLL/mods/modsavefiles.cpp
EladNLG: ????

NorthstarDLL/mods/modsavefiles.cpp
```diff
@@ -0,0 +1,515 @@
...
+		spdlog::error("DIR READ FAILED! Is the path valid?");
+		spdlog::info(path.string());
+		g_pSquirrel<context>->raiseerror(sqvm, ex.what());
+		return SQRESULT_ERROR;
+	}
+}
+
+// ok, I'm just gonna explain what the fuck is going on here because this
+// is the pinnacle of my stupidity and I do not want to touch this ever
+// again & someone will eventually have to maintain this.
```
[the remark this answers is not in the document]
EladNLG: Explanation provided in the function through it's comments.

NorthstarDLL/mods/modsavefiles.cpp
EladNLG: random crashes.

JMM889901 (approved):
Been making stuff on this for a while now, and no issues yet that aren't just a skill issue

ASpoonPlaysGames:
I think it would be really cool if modsettings and safe IO were to be released in the same update (they both are too large for a patch release and delaying either one much longer is gonna really suck honestly)

ASpoonPlaysGames:
Also @pg9182 just a reminder that you requested your own review on this 3 weeks ago, in case you forgot.

Your review on this PR would be really nice to have

GeckoEidechse:
> I think it would be really cool if modsettings and safe IO were to be released in the same update (they both are too large for a patch release and delaying either one much longer is gonna really suck honestly)

We can always just have two minor (in the semver sense) releases back-to-back, i.e `v1.15` one week and `v1.16` the week after. Based on past experiences smaller releases are better as I cannot rely on contributors to help me debug an issue on a broken release so smaller release means less commits to try between to figure out which one introduced a regression...

NorthstarDLL/mods/modsavefiles.cpp
```diff
@@ -0,0 +1,532 @@
...
+					spdlog::error("A file was supposed to be loaded but we can't access it?!");
+
+					g_pSquirrel<context>->AsyncCall("NSHandleLoadResult", handle, false, "");
+					mutex.get().unlock();
+					return;
+				}
+
+				std::stringstream stringStream;
+				while (fileStr.peek() != EOF)
+					stringStream << (char)fileStr.get();
```
pg9182, line 123: This is unnecessarily inefficient; just do `stringStream << fileStr.rdbuf()`.

NorthstarDLL/mods/modsavefiles.cpp
```diff
@@ -0,0 +1,532 @@
...
+
+				g_pSquirrel<context>->AsyncCall("NSHandleLoadResult", handle, true, stringStream.str());
+
+				fileStr.close();
+				mutex.get().unlock();
+				// side-note: this causes a leak?
+				// when a file is added to the map, it's never removed.
+				// no idea how to fix this - because we have no way to check if there are other threads waiting to use this file(?)
+				// tried to use m.try_lock(), but it's unreliable, it seems.
```
pg9182, line 132: If we make concurrent accesses fail instead of waiting, we can reduce the size of the critical section and make a global map of strings to bools protected by a single mutex instead. This will prevent buggy mods from queuing infinitely, while also resolving the leak.

NorthstarDLL/mods/modsavefiles.cpp
```diff
@@ -0,0 +1,532 @@
...
+			}
+			catch (std::exception ex)
+			{
+				spdlog::error("LOAD FAILED!");
+				mutex.get().unlock();
+				spdlog::error(ex.what());
```
pg9182, line 138: No `NSHandleLoadResult`?

NorthstarDLL/mods/modsavefiles.cpp
```diff
@@ -0,0 +1,532 @@
...
+template <ScriptContext context> void SaveFileManager::DeleteFileAsync(fs::path file)
+{
+	// P.S. I don't like how we have to async delete calls but we do.
+	auto m = std::ref(mutexMap[file]);
+	std::thread deleteThread(
+		[m, file]()
+		{
+			try
+			{
+				m.get().lock();
```
pg9182, line 156: Same here.

NorthstarDLL/mods/modsavefiles.cpp
```diff
@@ -0,0 +1,532 @@
...
+	// use it to circumvent the file extension suffix.
+	return std::any_of(str.begin(), str.end(), [](char c) { return c == '\0'; });
+}
+
+// Checks if the relative path (param) remains inside the mod directory (dir).
+// Paths are restricted to ASCII because encoding is fucked and we decided we won't bother.
+bool IsPathSafe(const std::string param, fs::path dir)
+{
+	auto const normRoot = fs::weakly_canonical(dir);
+	auto const normChild = fs::weakly_canonical(dir / param);
```
pg9182, line 190: Note that weakly_canonical can throw; this should be wrapped in a try/catch returning false if it catches.

NorthstarDLL/mods/modsavefiles.cpp
```diff
@@ -0,0 +1,532 @@
...
+
+	auto itr = std::search(normChild.begin(), normChild.end(), normRoot.begin(), normRoot.end());
+	// we return if the file is safe (inside the directory) and uses only ASCII chars in the path.
+	return itr == normChild.begin() && std::none_of(
+										   param.begin(),
+										   param.end(),
+										   [](char c)
+										   {
+											   unsigned char unsignedC = static_cast<unsigned char>(c);
+											   return unsignedC > 127 || unsignedC < 0;
```
pg9182, line 200: Why not just limit it to alnum specifically instead of ASCII in general? We should also restrict the special files on Windows (see https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file) (as an alternative, we could prepend a common suffix like an underscore to the filenames).

NorthstarDLL/mods/modsavefiles.cpp
```diff
@@ -0,0 +1,532 @@
...
+		return SQRESULT_ERROR;
+	}
+
+	g_pSquirrel<context>->pushinteger(sqvm, g_pSaveFileManager->LoadFileAsync<context>(dir / fileName));
+
+	return SQRESULT_NOTNULL;
+}
+
+// bool NSDoesFileExist(string file)
+ADD_SQFUNC("bool", NSDoesFileExist, "string file", "", ScriptContext::SERVER | ScriptContext::CLIENT | ScriptContext::UI)
```
pg9182, line 345: Is this necessary? It's less racy to just handle a missing file while attempting to read it instead of having a specific function for it.

NorthstarDLL/mods/modsavefiles.cpp
```diff
@@ -0,0 +1,532 @@
...
+				.c_str());
+		return SQRESULT_ERROR;
+	}
+
+	g_pSquirrel<context>->pushbool(sqvm, fs::exists(dir / (fileName)));
+	return SQRESULT_NOTNULL;
+}
+
+// int NSGetFileSize(string file)
+ADD_SQFUNC("int", NSGetFileSize, "string file", "", ScriptContext::SERVER | ScriptContext::CLIENT | ScriptContext::UI)
```
pg9182, line 369: Same thing.

pg9182 (changes requested):

NorthstarDLL/mods/modsavefiles.cpp
ASpoonPlaysGames: fun fact, we do this in lots of places (namely, mod loading)

pg9182: ugh, feel free to submit a PR to change that, then

NorthstarDLL/mods/modsavefiles.cpp
ASpoonPlaysGames: Modders might want to make use of it in some way, one case i can think of is checking if the user already has saved data for the mod (different from empty saved data) to do some first-time setup or defaults or something.

If it doesn't see use though, it doesn't see use. I personally could go either way with this tbh

NorthstarDLL/mods/modsavefiles.cpp
pg9182: oh, and if we do this for mod loading too, it'd probably be even better to do this:

```c++
std::ifstream st("whatever");
std::string str;
st.seekg(0, std::ios::end);   
str.reserve(st.tellg());
st.seekg(0, std::ios::beg);
str.assign((std::istreambuf_iterator<char>(st)), std::istreambuf_iterator<char>());
```

That'll prevent the string from being continuously re-allocated, which should be significantly faster and reduce the spikes in memory usage.

Could even put this in a helper function somewhere (and if we do that, we should also add proper error handling too).

NorthstarDLL/mods/modsavefiles.cpp
uniboi: Fair point, what would possible usecases for this be?

ASpoonPlaysGames: Checking how much space the mod has left to store things?

NorthstarDLL/mods/modsavefiles.cpp
ASpoonPlaysGames: sometimes, I really do hate c++

NorthstarDLL/mods/modsavefiles.cpp
uniboi: fair

ASpoonPlaysGames: Tbh I really don't foresee any mods actually going over the limit, except if someone does some like logging system in script or something weird

NorthstarDLL/mods/modsavefiles.cpp
EladNLG: That seems unnecessarily complex - we could just use a simple mutex for all file actions to avoid the leak completely (since an SSD/hard drive can only read one file at once anyways)

I don't think concurrent reads from multiple mods should result in a fail, as that would just make file reads on startup fail 99% of the time, and would complicate the file reading process too much.

NorthstarDLL/mods/modsavefiles.cpp
EladNLG: So, wait, do I change it or just leave it as-is and resolve this?

NorthstarDLL/mods/modsavefiles.cpp
pg9182: > we could just use a simple mutex for all file actions to avoid the leak completely (since an SSD/hard drive can only read one file at once anyways)

That may be suitable, but we'd have to test it. I'd be happy with that solution for now, though, and to deal with it if it causes performance issues later.

> concurrent reads from multiple mods should result in a fail

I mean concurrent accesses to the same file, and specifically, concurrent writes/deletes while reading. Multiple reads are fine.

NorthstarDLL/mods/modsavefiles.cpp
pg9182: > checking if the user already has saved data for the mod

Attempt to read it, then if you get a result saying it doesn't exist, do the setup logic. Anything else is just asking for hard-to-debug issues at some point.

NorthstarDLL/mods/modsavefiles.cpp
ASpoonPlaysGames: > So, wait, do I change it or just leave it as-is and resolve this?

I'd say definitely at least do `stringStream << fileStr.rdbuf()` for now

idk about the other (longer) suggestion, not sure if @pg9182 was suggesting to do that for mod loading, here, or both?

NorthstarDLL/mods/modsavefiles.cpp
pg9182: > Checking how much space the mod has left to store things?

Could maybe add a function for that specifically instead?

NorthstarDLL/mods/modsavefiles.cpp
pg9182: Yeah, it has the entire kitchen sink except for one of the knobs xd.

NorthstarDLL/mods/modsavefiles.cpp
ASpoonPlaysGames: Idk how I feel about the competing concepts here (prior checking vs handling error cases)

Generally I prefer prior checks to ensure that I won't do weird things, but for squirrel i tend to lean the other way (scripting moment)

I have no preference here overall tbh

NorthstarDLL/mods/modsavefiles.cpp
```diff
@@ -0,0 +1,532 @@
...
+
+	std::string content = g_pSquirrel<context>->getstring(sqvm, 2);
+	if (ContainsInvalidChars(content))
+	{
+		g_pSquirrel<context>->raiseerror(
+			sqvm, fmt::format("File contents may not contain NUL/\\0 characters! Make sure your strings are valid!", mod->Name).c_str());
+		return SQRESULT_ERROR;
+	}
+
+	fs::create_directories((dir / fileName).parent_path());
```
EladNLG, line 237: uh this is stupid i think?

NorthstarDLL/mods/modsavefiles.cpp
ASpoonPlaysGames: > Could maybe add a function for that specifically instead?

I think that this is probably a good idea, The only use case (that is sane) that I can think of is to check how much space is remaining, but that would just mean the modder has to do `MAX_FILE_SIZE_CONST_THING - NSGetFileSize()` to actually get the value that they want.

Speaking of having to use a const, the size limitation const should imo be exposed to squirrel so that modders can ensure that they arent trying to write excessively sized files or something

NorthstarDLL/mods/modsavefiles.cpp
pg9182: It's usually better to handle the error cases, unless you don't even plan to open the file to begin with (in which case checking if it exists only is faster and less error-prone). For first-time initialization, error handling is much better in general.

ASpoonPlaysGames: I don't see a valid case where a modder will want to check that a file exists without then reading it (except maybe to avoid overriding a file? But if that's the case why weren't they reading it?), so with that being said I'm going to lean on the side of error handling

NorthstarDLL/mods/modsavefiles.cpp
EladNLG: This was introduced to avert the encoding problems, not to check if a file name is actually "valid".

There are many, _many_ prerequisites for that and I think implementing all of them would be neigh impossible; The best way to check if a file is valid is to try and write to it. I do think that if a save fails, it shouldn't be silent, however, since we have to write to the file asynchronously, that's what is currently done. Maybe we should add a callback for that?

NorthstarDLL/mods/modsavefiles.cpp
JMM889901: I currently do this but thats mostly because i didn't want to use a try-catch and the failcallback wasn't a thing when i wrote the code that uses that
```
void function cl_titanFrameworkPersistenceInitThread()
{
    if(!NSDoesFileExist("frameworkCFG.json"))
    {
        print("Creating new frameworkCFG.json")
        NSSaveJSONFile("frameworkCFG.json", {version = FRAMEWORK_VERSION});
    }
    if(!NSDoesFileExist("frameworkXP.json"))
    {
        print("Creating new frameworkXP.json")
        NSSaveJSONFile("frameworkXP.json", {version = FRAMEWORK_VERSION});
    }
    NSLoadJSONFile( "frameworkXP.json", loadPersistentXP)
    NSLoadJSONFile( "frameworkCFG.json", loadPersistentData)
}
```

EladNLG (commented):
poopoo

NorthstarDLL/mods/modsavefiles.cpp
EladNLG: This was introduced to avert the encoding problems, not to check if a file name is actually "valid".

There are many, _many_ prerequisites for that and I think implementing all of them would be neigh impossible; The best way to check if a file is valid is to try and write to it. I do think that if a save fails, it shouldn't be silent, however, since we have to write to the file asynchronously, that's what is currently done. Maybe we should add a callback for that?

NorthstarDLL/mods/modsavefiles.cpp
EladNLG: > I don't see a valid case where a modder will want to check that a file exists without then reading it (except maybe to avoid overriding a file? But if that's the case why weren't they reading it?), so with that being said I'm going to lean on the side of error handling

1) To confirm overriding a file
2) If a read fails - it may be due to multiple reasons, not just the file not existing.

NorthstarDLL/mods/modsavefiles.cpp
EladNLG: Will add in next commit `NSGetModSizeLimit` to get the size limit of a save folder, and `NSGetTotalSpaceRemaining` which returns the space remaining for a save folder until it reaches the cap.

ASpoonPlaysGames: > Will add in next commit `NSGetModSizeLimit` to get the size limit of a save folder

Would it be better to push the const used here to the squirrel VM directly instead of making a function for it?

NorthstarDLL/mods/modsavefiles.cpp
EladNLG: No need for a `NSHandleLoadResult` call here..? This is DeleteFileAsync.

NorthstarDLL/mods/modsavefiles.cpp
EladNLG: True, will call it `MAX_FOLDER_SIZE`.

NorthstarDLL/mods/modsavefiles.cpp
EladNLG: Resolved in df0554a

NorthstarDLL/mods/modsavefiles.cpp
EladNLG: Resolved in https://github.com/R2Northstar/NorthstarLauncher/commit/df0554a0e7cd8995eb8ecfaf045f659f99785b34

NorthstarDLL/mods/modsavefiles.cpp
EladNLG: Resolved in https://github.com/R2Northstar/NorthstarLauncher/commit/df0554a0e7cd8995eb8ecfaf045f659f99785b34

F1F7Y:
@catornot made a safe io plugin, although not released as far as i know.

My personal view on this is to have new features like this in plugins and reimplemented features ( like bots ) in launcher.

ASpoonPlaysGames:
> @catornot made a safe io plugin, although not released as far as i know.

They did, yeah. However it is an inferior version, afaik you have to like, poll the plugin every frame to know if a task is finished and stuff due to plugins limitations. (No hate to catornot, they made it in a couple of hours it's just a limitation of plugins v2)

> My personal view on this is to have new features like this in plugins and reimplemented features ( like bots ) in launcher.

I would agree, but plugins is still a bit half-baked atm, both in functionality and in mod manager support. Also in this case I would strongly recommend shipping a plugin like this with northstar or implementing in launcher, because of the number of mods that will use it. 

Imo this falls under the same sort of category as ModSettings and squirrel http, where they probably should be a plugin/mod, but because of how much they are used it's easier to ship them with northstar in some capacity and not have tickets caused by missing dependencies.

uniboi:
> Imo this falls under the same sort of category as ModSettings and squirrel http, where they probably should be a plugin/mod, but because of how much they are used it's easier to ship them with northstar in some capacity and not have tickets caused by missing dependencies.

We can still include plugins in a northstar install like discord rpc.

I think sensitive additions like HTTP and I/O should be plugins because that allows easy disabling without launch args & custom implementations.
Also more modularity is preferable imo

NorthstarDLL/mods/modsavefiles.cpp
uniboi: > I'd be happy with that solution for now, though, and to deal with it if it causes performance issues later.

So this is resolved?

GeckoEidechse: bump

EladNLG: this is resolved. we now use a single mutex for all file operations.

NorthstarDLL/mods/modsavefiles.cpp
EladNLG: Yeah, I know. That's why I didn't resolve it yet.

NorthstarDLL/mods/modsavefiles.cpp
EladNLG: Can this be resolved? I think this function can stay (as it may have usecases)

ASpoonPlaysGames: Yeah sure

NorthstarDLL/mods/modsavefiles.cpp
EladNLG: fixed in https://github.com/R2Northstar/NorthstarLauncher/pull/404/commits/a820107d116e1fb882369ad18703eab20c547f09

PR #406, opened 2023-01-29
Format content of `enabledmods.json`

Glacir:
Before:

```json
{
    "Northstar.Client": true, "Northstar.CustomServers": true,  "Northstar.Custom": true
}
```

After:
```json
{
    "Northstar.Client": true,
    "Northstar.CustomServers": true,
    "Northstar.Custom": true
}
```


ASpoonPlaysGames:
I see no downside really, and it could help when debugging broken mods sometimes (could edit the json easier instead of deleting enabledmods.json)

ASpoonPlaysGames (approved):

BobTheBob9 (approved):

PR #407, opened 2023-01-29
fix broken filters

BobTheBob9:
oops, filters are broken again due to a missing </ClInclude>, this fixes them


RoyalBlue1 (approved):
Fixes the filter

GeckoEidechse (approved):
(Mandatory approval for GH)

PR #409, opened 2023-01-29
Allow `launchplaylist` to be used to start dedicated servers

BobTheBob9:
`launchplaylist` selects a random map and mode from a given playlist and starts a server on that playlist, this PR allows it to be used as a startup command to set startup map on a dedicated server

### DIFF
--- a/NorthstarDLL/dedicated/dedicated.cpp
+++ b/NorthstarDLL/dedicated/dedicated.cpp
@@ -48,9 +48,9 @@ void RunServer(CDedicatedExports* dedicated)
 	// initialise engine
 	g_pEngine->Frame();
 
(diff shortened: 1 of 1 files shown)


GeckoEidechse (approved):
Behaved as expected when testing dedicated server with each

- `+launchplaylist fw` which loaded Frontier War on Forwardbase Kodai
- no extra args which just gave private match lobby
- `+map mp_thaw +playlist fw +mp_gamemode fw` which loaded Frontier War on Exoplanet
- `+playlist fw +mp_gamemode fw` which loaded into private match lobby with _Skirmish_ selected (same behaviour as `v1.12.0` already has)

PR #410, opened 2023-01-30
Create mod entry in `enabledmods.json` if it doesn't exist

Alystrasz:
Currently, when you add a new mod to your `mods/` directory, when there's no associated entry in the `enabledmods.json` file, it is considered enabled by default; mod entries are only written in `enabledmods.json` when toggling them via the mods interface.

In this pull request, I propose to create mod entries in `enabledmods.json` when detecting they don't exist.


BobTheBob9:
will this write downloaded mods to the json as well? while we don't have autodownloading yet we do read the mod directory for downloaded mods atm, and i don't think we should ever write downloaded mods to the json

Alystrasz:
> will this write downloaded mods to the json as well? while we don't have autodownloading yet we do read the mod directory for downloaded mods atm, and i don't think we should ever write downloaded mods to the json

Fixed in 5de25eba2f45b3ed8e1ece2e9160ca042493c6d9 and 21975fad173a2f3f16c28abe918368dc20a6d3c5

F1F7Y:
This would be pretty pog ngl

F1F7Y (approved):
Works in testing

Alystrasz:
I won't edit the labels for this PR, since I made it, but I'd say this is ready to be merged. :)

Jan200101 (dismissed):

Jan200101 (approved):
re-doing the review
code still looks good and it matches its surrounding code

NorthstarDLL/mods/modmanager.cpp
```diff
@@ -703,11 +706,21 @@ void ModManager::LoadMods()
...
+	// This is used to check if some mods have a folder but no entry in enabledmods.json
+	bool newModsDetected = false;
+
 	for (Mod& mod : m_LoadedMods)
 	{
 		if (!mod.m_bEnabled)
 			continue;
 
+		// Add mod entry to enabledmods.json if it doesn't exist
+		if (!mod.m_bIsRemote && !m_EnabledModsCfg.HasMember(mod.Name.c_str()))
```
GeckoEidechse, line 718: I assume that means we ignore remote mods when adding to `enabledmods.json`? ^^

Alystrasz: Exactly yeah, as discussed in https://github.com/R2Northstar/NorthstarLauncher/pull/410#issuecomment-1426284180

GeckoEidechse (approved):
Code looks fine as far as I can tell. Primarily looked at control flow.

PR #411, opened 2023-02-02
Add `CGlobals` class and `g_pGlobals`, and update code to support

BobTheBob9:
adds the CGlobals class and global pointer g_pGlobals, contains a few useful engine vars that we can use in preexisting code (code has been updated to do so)


RoyalBlue1:
Code looks good and it seems to run fine

RoyalBlue1 (approved):
Code looks good and it seems to run fine

GeckoEidechse (approved):
Approving without proper review per request from @BobTheBob9 to merge. @RoyalBlue1 already looked at it anyway.

PR #415, opened 2023-02-04
feat: Bad mod formatting warnings

Alystrasz:
While loading mods, this checks if a `mod.json` file is present in mods directories; if not, this will display a warning message, telling user mod hasn't been installed properly.

This does not analyze hidden directories (with a filename beginning with a `.`).

image

### DIFF
--- a/NorthstarDLL/mods/modmanager.cpp
+++ b/NorthstarDLL/mods/modmanager.cpp
@@ -381,19 +381,23 @@ void ModManager::LoadMods()
 
 	for (std::filesystem::directory_iterator modIterator : {classicModsDir, remoteModsDir})
 		for (fs::directory_entry dir : modIterator)
+		{
 			if (fs::exists(dir.path() / "mod.json"))
 				modDirs.push_back(dir.path());
 			else
 			{
+				std::string filename = dir.path().filename().generic_string().c_str();
+				// Don't display an error for hidden directories
+				if (filename.at(0) == '.')
+					continue;
+
 				spdlog::warn("Directory {} has no mod.json file.", dir.path().generic_string().c_str());
 				std::string errorMessage = std::format(
 					"The directory {} does not contain a mod.json file.\nMake sure you correctly installed it.",
 					dir.path().generic_string().c_str());
-				MessageBoxA(
-					GetForegroundWindow(), errorMessage.c_str(),
-					"Badly formatted mod",
-					0);
+				MessageBoxA(GetForegroundWindow(), errorMessage.c_str(), "Badly formatted mod", 0);
 			}
+		}
 
 	for (fs::path modDir : modDirs)
 	{
--- a/NorthstarDLL/mods/modmanager.cpp
+++ b/NorthstarDLL/mods/modmanager.cpp
@@ -17,6 +17,7 @@
 #include <string>
 #include <sstream>
 #include <vector>
+#include <server/serverchathooks.h>
 
 ModManager* g_pModManager;
 
@@ -382,6 +383,17 @@ void ModManager::LoadMods()
 		for (fs::directory_entry dir : modIterator)
 			if (fs::exists(dir.path() / "mod.json"))
 				modDirs.push_back(dir.path());
+			else
+			{
+				spdlog::warn("Directory {} has no mod.json file.", dir.path().generic_string().c_str());
+				std::string errorMessage = std::format(
+					"The directory {} does not contain a mod.json file.\nMake sure you correctly installed it.",
+					dir.path().generic_string().c_str());
+				MessageBoxA(
+					GetForegroundWindow(), errorMessage.c_str(),
+					"Badly formatted mod",
+					0);
+			}
 
 	for (fs::path modDir : modDirs)
 	{
--- a/NorthstarDLL/mods/modmanager.cpp
+++ b/NorthstarDLL/mods/modmanager.cpp
@@ -17,7 +17,6 @@
 #include <string>
 #include <sstream>
 #include <vector>
-#include <server/serverchathooks.h>
 
 ModManager* g_pModManager;
 


Alystrasz:
@GeckoEidechse if you want to keep doing [your shenanigans](https://github.com/R2Northstar/NorthstarLauncher/pull/405#issuecomment-1412805019), this will ignore your git folders 😁

GeckoEidechse:
> @GeckoEidechse if you want to keep doing [your shenanigans](https://github.com/R2Northstar/NorthstarLauncher/pull/405#issuecomment-1412805019), this will ignore your git folders grin

Depending on how fast VTOL switches to using `enabledmods.json` we might also have to ignore the `disabled` folder it uses to hide the `mod.json`.
@BigSpice is there an ETA on the fix in VTOL? ^^

GeckoEidechse:
Actually I just realised that this might also cause issues for players where there is a leftover `disabled` folder in the mods dir. Not sure if Viper ever cleaned that up after switching to `enabledmods.json`.

It might make more sense to check folder recursively for `mod.json` and only warn if there is actually one in a subdirectory.

Alystrasz:
> It might make more sense to check folder recursively for mod.json and only warn if there is actually one in a subdirectory.

What about mod folders with no `mod.json` file? (quite unlikely, I know)

GeckoEidechse:
> What about mod folders with no `mod.json` file? (quite unlikely, I know)

For that to occur, the player must have manually deleted that file, given that a mod author wouldn't release a mod with missing `mod.json` cause that mod would have never loaded on their end in the first place. I'd argue that this would be a rarer occurrence then a player accidentally creating an empty folder. ^^

Alystrasz:
> It might make more sense to check folder recursively for mod.json and only warn if there is actually one in a subdirectory.

Addressed in c0e44c9.

image

BobTheBob9:
concept is good, but would be nicer if this was exposed to script and shown in ui, rather than than shown as a messagebox? messagebox would be annoying in cases where mod formatting isn't an issue since it'd block execution

BobTheBob9:
> What method would you use to show error message in UI? All cool UI methods I know are callable from Squirrel VM.


something on the title screen to display warnings would work, would need to expose all messages to squirrel on reload and do some ui stuff in script to display them

> Yeah that's also cool, because this way we're sure user knows some of their mods are fucked (in comparison with a notification that closes automatically without user noticing it)

it'd be very annoying for developers in situations where they know they'll get the warning and don't want to wait for a blocking textbox every startup, is what i'm thinking

NorthstarDLL/mods/modmanager.cpp
```diff
@@ -380,8 +380,33 @@ void ModManager::LoadMods()
...
+							"mod.json file for directory {} is located at the wrong location ({}).",
+							dir.path().generic_string().c_str(),
+							subdir.path().generic_string().c_str());
+						std::string errorMessage = std::format(
+							"The folder {} does contain a mod.json file, but not at the correct location.\n\nExpected: {}\nFound: "
+							"{}\n\nMake sure you correctly installed it.",
+							dir.path().filename().generic_string(),
+							dir.path().generic_string().append("/mod.json"),
+							subdir.path().generic_string().append("/mod.json"));
+						MessageBoxA(GetForegroundWindow(), errorMessage.c_str(), "Badly formatted mod", 0);
```
pg9182, line 406: Please don't do this for dedicated servers.

Alystrasz:
Marking this as draft while I refactor to address comments.

uniboi:
> What method would you use to show error message in UI? All cool UI methods I know are callable from Squirrel VM.

Define a global function in the UI that displays a dialog / menu / submenu that explains what's going on and call that function from native after the UI VM is initialized and mods that are formatted wrong are detected

Alystrasz:
Reviews have been addressed, @uniboi implemented the UI part in https://github.com/R2Northstar/NorthstarMods/pull/597.
I'm marking this as ready for reviews.

uniboi:
Do note that this changes the function signatures of some script methods as they are currently incorrect. https://github.com/R2Northstar/NorthstarMods/pull/597 fixes those issues as well.

NorthstarDLL/mods/modmanager.cpp
Alystrasz: Faulty mods' paths are not displayed in a dialog pop-up anymore, but stored in a variable instead.

GeckoEidechse:
Wait, how does this depend on https://github.com/R2Northstar/NorthstarMods/pull/597 ?

Isn't it the other way round or do both depend on each other?

As in can this launcher PR not be added alone without breaking something in mods?

Alystrasz:
> As in can this launcher PR not be added alone without breaking something in mods?

It definitely can, but you won't see any displayed errors, which is the point of both PR 😃

GeckoEidechse:
> > As in can this launcher PR not be added alone without breaking something in mods?
> 
> It definitely can, but you won't see any displayed errors, which is the point of both PRs 😃

Oh yeah, that's totally fine and expected :P
The _"Depends on"_ just threw me off, updating PR description accordingly ^^

uniboi:
> Wait, how does this depend on [R2Northstar/NorthstarMods#597](https://github.com/R2Northstar/NorthstarMods/pull/597) ?
> 
> Isn't it the other way round or do both depend on each other?
> 
> As in can this launcher PR not be added alone without breaking something in mods?

No, this introduces compile errors (without the mods pr) because I corrected some signatures of functions exposed to squirrel.

Alystrasz:
Mb, I forgot about the `ornull` part.
Off to bed!

GeckoEidechse:
Aight, updating PR description again xD

NorthstarDLL/scripts/client/scriptmodmenu.cpp
```diff
@@ -1,165 +1,108 @@
...
-	// manual lookup, not super performant but eh not a big deal
-	for (Mod& mod : g_pModManager->m_LoadedMods)
+	if (index < 0 || index >= g_pModManager->m_LoadedMods.size())
 	{
-		if (!mod.Name.compare(modName))
-		{
-			g_pSquirrel<context>->pushstring(sqvm, mod.Version.c_str());
-			return SQRESULT_NOTNULL;
-		}
+		spdlog::warn("Tried disabling mod at {} but only {} mods are loaded", index, g_pModManager->m_LoadedMods.size());
```
ASpoonPlaysGames, line 63: Do we know how to push a proper script error? (with the message)

If so we should be doing that here

NorthstarDLL/mods/modmanager.cpp
```diff
@@ -398,10 +398,44 @@ void ModManager::LoadMods()
...
 	for (std::filesystem::directory_iterator modIterator : {classicModsDir, remoteModsDir})
 		for (fs::directory_entry dir : modIterator)
+		{
 			if (fs::exists(dir.path() / "mod.json"))
 				modDirs.push_back(dir.path());
+			else if (fs::is_directory(dir.path()))
+			{
+				std::string filename = dir.path().filename().generic_string().c_str();
+				// Don't display an error for hidden directories
+				if (filename.at(0) == '.')
```
ASpoonPlaysGames, line 412: Technically speaking you can have non-hidden directories that start with `.`

Example would be the `.github` folder

Not rly sure that it's a big deal in any way, if someone makes a mod/package that starts with `.` then thats their problem

NorthstarDLL/mods/modmanager.cpp
```diff
@@ -398,10 +398,44 @@ void ModManager::LoadMods()
...
+						spdlog::warn(
+							"mod.json file for directory {} is located at the wrong location ({}).",
+							dir.path().generic_string().c_str(),
+							subdir.path().generic_string().c_str());
+						// read mod json file
+						std::ifstream jsonStream(modPath);
+						std::stringstream jsonStringStream;
+
+						while (jsonStream.peek() != EOF)
+							jsonStringStream << (char)jsonStream.get();
```
ASpoonPlaysGames, line 429: Is there not a better way to just read the entire file to a string?

ASpoonPlaysGames (changes requested):
Most of the code looks good, just a couple of questions/comments

NorthstarDLL/scripts/client/scriptmodmenu.cpp
uniboi: I think I've heard someone say that the function doesn't work but in principle we know.
This should be done in a future PR for all exposed functions imo

NorthstarDLL/mods/modmanager.cpp
Alystrasz: Yeah, a good example is indeed people directly cloning `NorthstarMods` into their game folder (just like https://github.com/R2Northstar/NorthstarLauncher/pull/405#issuecomment-1412805019); I think directories starting with a `.` are configuration-related folders, and thus mods shouldn't be named that way.

NorthstarDLL/mods/modmanager.cpp
Alystrasz: I stole this one from [NorthstarLauncher existing code](https://github.com/R2Northstar/NorthstarLauncher/blob/6857b93f623dffedffac23fdd3983c74621d2b64/NorthstarDLL/mods/modmanager.cpp#L419); if you know a better way to do this, please let me know :)

ASpoonPlaysGames, line 429:
```suggestion
						jsonStringStream << jsonStream.rdbuf();
```
this should work a bit better, adding the entire file buffer in one go instead of byte by byte

(original conversation was based on only the second line, so I couldnt suggest a change on both lines)

NorthstarDLL/scripts/client/scriptmodmenu.cpp
Alystrasz: @uniboi do we know?

uniboi: AFAIK we do not know how to get it working right now. I can try some things but would prefer it in a future PR as said above

NorthstarDLL/mods/modmanager.cpp
uniboi: @Alystrasz can you apply this change?

Alystrasz: On it atm

Alystrasz: Done in https://github.com/R2Northstar/NorthstarLauncher/pull/415/commits/25847b44dedc639c53fa1d64c7037da16ab78573

Alystrasz:
Feature is still working as expected after newer changes.

image

image

F1F7Y (approved):
Works in testing

Unsure how we feel about [smart pointers](https://github.com/R2Northstar/NorthstarLauncher/pull/415/files#diff-5cc5bc0761e894d10f01d04e3292a405376f5f3c6007b7d19c10bcb8a0b07161R152). Could also use a new class instead of `Mod` as we don't need most of `Mod` for invalid mods.

uniboi:
What's your issue with smart pointers?

F1F7Y:
> What's your issue with smart pointers?

The fact we don't use them anywhere else ( apart from logging, but that dont count ). Also the fact that a few lines up there is a `std::vector<Mod>`.

My personal opinion is that this can be merged as is since ill probably get to cleaning this up ( and possibly including some of bobs changes from mod loading refactor ) sooner or later.

GeckoEidechse:
@Alystrasz just a heads-up that this PR sadly has merge conflicts now ^^"

Alystrasz:
> @Alystrasz just a heads-up that this PR sadly has merge conflicts now ^^"

Done :)

ASpoonPlaysGames:
@Alystrasz would this need a rework to check `packages` as well? We already have basic format requirements for `packages` already so maybe not but want your opinion

Alystrasz:
> @Alystrasz would this need a rework to check `packages` as well? We already have basic format requirements for `packages` already so maybe not but want your opinion

About `packages`, currently we only check if directory names match the `AUTHOR-MOD-VERSION` pattern, not checking if the mod manifest is at the right place; this has been added in e398b5cb5a40ed08b0fff20ed78df9dddf51170e and 0d6b62491326bbf764bbb5ad32fad5dc8772cb9d.

I removed the "READY TO MERGE" label since the diff shows this is deleting functions such as `NSGetModNames` (probably some merge operation failed), I need to address that.

Alystrasz:
Waiting for @uniboi to re-add the `NSGetInvalidMods` function with new syntax since they've done it in the first place :)

Alystrasz:
Moving the PR to draft while waiting for changes from authors.

PR #416, opened 2023-02-07
Add `FUNCTION_AT` and `VAR_AT` macros

BobTheBob9:
adds macros for referencing variables and functions at given locations in memory, TODO: convert all code to use these macros where they should be


BobTheBob9:
making this ready to review as gecko said it would be chill atm

GeckoEidechse (approved):
Printed the thing Bob said it should print when running the command Bob said I should run so I guess it works?

PR #419, opened 2023-02-15
Use Semver in Dependency Constants

uniboi:
If a mod for a dependency constant is loaded, this will create a sqvm constant that's a SQTable holding the dependency's `major`, `minor` and `patch` version as integer. All other subversions are omitted.
If a dependency is not loaded, the constant will be `0` instead.

This does not break compatability with old versions since a table is a truthy value like `1`, as it is now.

This allows modders to compile code for ranges of mod versions.

For example:
```
#if HAS_UEBERBLICK
// This compiles only if the dependency is loaded. No matter the version of the dependency
#endif

#if HAS_UEBERBLICK && HAS_UEBERBLICK.major == 1
// This compiles ONLY if the dependency version is 1.x.x
#endif

#if HAS_UEBEBLICK && HAS_UEBERBLICK.major == 1 && HAS_UEBERBLICK.minor >= 2 && HAS_UEBERBLICK.minor <= 5
// This compiles only if the dependency's version is between 1.2.x - 1.5.x
#endif
```

Invalid version numbers are represented as `-1`.

- "1.2.test"
  major: 1
  minor: 2
  patch: -1

- "1."
  major: 1
  minor: -1
  patch: -1

- "heheheha"
  major: -1
  minor: -1
  patch: -1

etc.


NorthstarDLL/squirrel/squirrel.cpp
```diff
@@ -200,12 +200,55 @@ template <ScriptContext context> void SquirrelManager<context>::VMCreated(CSquir
...
+					if (std::all_of(sub.begin(), sub.end(), ::isdigit))
+						semver.push_back(stoi(sub));
+					else
+						semver.push_back(-1);
+				}
+				else
+					semver.push_back(-1);
+
+				last = 0;
+				next = 0;
```
uniboi, line 229: idk if this is needed but it looked like without this the launcher crashes sometimes when reloading mods

GeckoEidechse:
I assume `v1.12.1-rc` would then be:

major: 1
minor: 12
patch: 1

?

uniboi:
v1.12.1-rc would be

mj: -1
mn: 12
pt: -1

since I did not bother writing an semver parser because
a) I have never seen a mod with a leading v in the version
b) I don't think anyone is playing on release candidates

GeckoEidechse:
> v1.12.1-rc would be
> 
> mj: -1
> mn: 12
> pt: -1
> 
> since I did not bother writing an semver parser because
> a) I have never seen a mod with a leading v in the version

the leading `v` was a force of habit, forgot that it's not in the mods version

> b) I don't think anyone is playing on release candidates

sadly cause we really need more playtesters x_x

uniboi:
I could just terminate on a `-`, I guess. parsing rc subversion doesn't make sense imo

GeckoEidechse:
> I could just terminate on a `-`, I guess. parsing rc subversion doesn't make sense imo

Yeah that's fine. Was just wondering whether it would be a `-1` or whether the `-rcX` is stripped and turned into just the `1` for patch ^^

uniboi:
Ignores now anything after a `-`

BobTheBob9:
genuinely surprised any kind of table stuff works in #ifs?

uniboi:
Conditions for #if are normal squirrel expressions, they just need to be able to be evaluated at compile time (ie be constant).

Even stuff like this works `#if { k = 1 }.k == int(CLIENT)`, even if unnecessary

uniboi:
fuck

uniboi:
This is broken atm actually idk what I did

uniboi:
fixed again.
[dep_test.zip](https://github.com/R2Northstar/NorthstarLauncher/files/10947024/dep_test.zip)
To test: Install both mods in the zip. In the mp lobby, the globals `a = 1`, `b = 2` and `c = 3` should be defined while `d = 4` isn't.

EladNLG:
Small request:

Make the version available as a string as well :)

uniboi:
> Make the version available as a string as well :)

you can now do checks like `#if CONSTANT.version == "1.2.3" #elseif CONSTANT.version == "1.2.3-rc1" #endif`

ScureX (approved):
very good i like so hot and sexy mmhmmm nice code yes

NorthstarDLL/squirrel/squirrel.cpp
```diff
@@ -200,13 +202,53 @@ template <ScriptContext context> void SquirrelManager<context>::VMCreated(CSquir
...
+				std::string v = dependency.Version;
+				std::vector<SQInteger> semver;
+				size_t last = 0;
+				size_t next = 0;
+
+				size_t rc = v.find('-');
+				if (rc != std::string::npos)
+				{
+					v = v.substr(0, rc);
+				}
```
pg9182, line 214: If we want to be truly semver-compliant, the prerelease logic is a lot more complicated, both for splitting, and for comparison.

https://cs.opensource.google/go/x/mod/+/refs/tags/v0.9.0:semver/semver.go
https://semver.org/spec/v2.0.0.html#spec-item-9

uniboi: I can include something like https://github.com/zmarko/semver to replace my very simple parsing logic

GeckoEidechse: I think that's going a bit overkill. I'm personally fine with just limiting it to `MAJOR.MINOR.PATCH` and only extending from there if need be ^^

pg9182: Ehh, I'm okay with either (as long as we support whatever thunderstore supports, at a minimum), but we can't call it semver if it isn't.

GeckoEidechse: Thunderstore currently only supports `MAJOR.MINOR.PATCH`

https://github.com/thunderstore-io/Thunderstore/issues/634

ASpoonPlaysGames: We should support whatever the `version` in the mod.json supports, we shouldn't limit ourselves to just what thunderstore supports, as a thunderstore package != a northstar mod

pg9182: However, they say that "`in any case, we've not ruled out extended semver support`"...

To ensure it doesn't break in the future, we could check the format to ensure it only contains `MAJOR.MINOR.PATCH[-rc#]`, and always count it as a non-match if it doesn't.

ASpoonPlaysGames: > However, they say that "`in any case, we've not ruled out extended semver support`"...

This PR should have nothing to do with thunderstore in any capacity, therefore making this irrelevant.

uniboi: I might be misremembering things but iirc the `Version` in the mod.json has nothing to do with the `version` in the thunderstore manifest.

ASpoonPlaysGames: > I might be misremembering things but iirc the `Version` in the mod.json has nothing to do with the `version` in the thunderstore manifest.

correct

uniboi: > Ehh, I'm okay with either (as long as we support whatever thunderstore supports, at a minimum), but we can't call it semver if it isn't.

Well let's call it "Dependency Constants that support version comparisons" or whatever, doesn't really matter

NorthstarDLL/squirrel/squirrel.h
```diff
@@ -282,6 +295,51 @@ class SquirrelManagerBase
 		// there are entity constants for other types, but seemingly CBaseEntity's is the only one needed
 		return (T*)__sq_getentityfrominstance(m_pSQVM, &obj, __sq_GetEntityConstant_CBaseEntity());
 	}
+
+	inline SQRESULT createslot(HSquirrelVM* sqvm, const SQChar* key, const SQChar* value, bool isAsset = false, SQInteger idx = -3)
```
uniboi, line 299: I just remembered squirrel tables support any type as keys not just tables, if anyone knows a way to make this generic I'd appreciate it

uniboi: I updated `createslot` to be more generic, it should now be possible to insert anything as keys and values, except assets since there isn't a way rn to tell apart sq strings and assets

NorthstarDLL/squirrel/squirrel.cpp
pg9182: Ah, that's good to know. I'm not terribly familiar with the modding side of NS, and didn't realize that we version them independently from thunderstore.

catornot (commented):
tested the pr; works well except it crashed once ( used the last ci build btw )

used a script command to test this : `script printt(HAS_DEP_TEST_2.version, "major:",HAS_DEP_TEST_2.major,"minor:",HAS_DEP_TEST_2.major,"patch:",HAS_DEP_TEST_2.patch)`

with dep_test_2
```bash
1.2.3 major: 1 minor: 1 patch: 3
```
without dep_test_2
```bash
[NORTHSTAR] [info] sq_compilebuffer returned SQRESULT_NULL
[NORTHSTAR] [info] sq_call returned SQRESULT_ERROR
```

with rc for patch
```bash
1.2.3-rc major: 1 minor: 1 patch: 3
```

another note it also works by just using `#if DEP` :)

[nslog2023-03-20 19-28-52.txt](https://github.com/R2Northstar/NorthstarLauncher/files/11023883/nslog2023-03-20.19-28-52.txt)

uniboi:
It looks like the game crashes when accessing a dependency constant table and then recompiling scripts

ASpoonPlaysGames:
@uniboi is this still going to be worked on? Or has the crashing discouraged you?

uniboi:
I think it might just be a refcount issue, I haven't tested any further though.
Might try soon:tm:

GeckoEidechse:
@uniboi heads-up that there are merge conflicts here ^^

uniboi:
fixed in #685

PR #310, opened 2022-10-28
Squirrel bridge v3

emma-miler:
Adds `call` and `schedule_call` with completely automatic variadic argument pushing and threadsafe asynchronous calls
Depends on https://github.com/R2Northstar/NorthstarMods/pull/525

PR #428, opened 2023-03-03
Bring back sq_getobject and sq_getthisentity for the client and ui vm

### LINKED ISSUE #427: __sq_getobject and __sq_getthisentity are null on client and ui
it seams like they were removed at some point since `__sq_getobject` was added in #310. I don't know when it was removed or why. 

they are getting set for ui as null so they were implemented at some point.
 https://github.com/R2Northstar/NorthstarLauncher/blob/32165afe41c4cf4d3b261ee7a9c69c369253cc13/NorthstarDLL/squirrel/squirrel.cpp#L690 https://github.com/R2Northstar/NorthstarLauncher/blob/32165afe41c4cf4d3b261ee7a9c69c369253cc13/NorthstarDLL/squirrel/squirrel.cpp#L691

`sq_getobject` would be useful for some client plugins

catornot:
for some reason they were removed in pr #310.
I need `sq_getobject` for some plugins on the ui vm and since they are null, I can't use them.

would close #427 :)


uniboi (approved):
looks like the functions have been removed for client on accident

GeckoEidechse (approved):
Ok no clue why they were removed in the original PR.

(didn't test PR btw)

PR #432, opened 2023-03-05
Use connectionless packets for authentication

### LINKED ISSUE #429: Perform server auth over game port
Reuse the game port for server auth and verification instead of running a separate HTTP server.

- easier server setup
- only a single port to forward
- easier p2p and tunneling if we ever decide to do so
- nicer architecture
- auth working is more closely tied to the specific server, eliminating a very common class of hard-to-troubleshoot config problems

The implementation in Atlas will be backwards-compatible with old clients/servers.

pg9182:
This replaces the built-in HTTP server with connectionless packets for authentication.

A new `T` packet type is defined for JSON requests from Atlas (replies are sent over HTTP). This type does not conflict with existing connectionless packet types.

This change requires an Atlas update, but does not require a version gate since compatibility with old gameservers is retained and clients are not affected.

- closes #429
- includes #434 (merge it first)
- requires R2Northstar/Atlas#31 (or server registration will fail)


pg9182:
Testing the connectionless packet encoding and handling:

```go
package main

import (
	"crypto/aes"
	"crypto/cipher"
	"crypto/hmac"
	"crypto/rand"
	"crypto/sha256"
	"encoding/json"
	"errors"
	"fmt"
	"net"
	"net/netip"
	"os"
	"time"
)

var ErrTimeout = errors.New("connection timed out")

func main() {
	fmt.Println(r2send(netip.MustParseAddrPort("10.33.0.189:47120"), time.Second, r2encodeAtlasSigjson1("dummy", map[string]any{
		"test": true,
	})))
}

func r2encodeAtlasSigjson1(key string, data any) []byte {
	j, err := json.Marshal(data)
	if err != nil {
		panic(err)
	}

	sig := hmac.New(sha256.New, []byte(key))
	sig.Write(j)

	var b []byte
	b = append(b, '\xFF', '\xFF', '\xFF', '\xFF') // connectionless: int32(-1)
	b = append(b, 'T')                            // packet kind
	b = append(b, "sigjson1\x00"...)              // packet type
	b = sig.Sum(b)                                // sigjson1 - signature
	b = append(b, j...)                           // sigjson1 - data
	return b
}

const (
	r2cryptoNonceSize = 12
	r2cryptoTagSize   = 16
	r2cryptoKey       = "X3V.bXCfe3EhN'wb"
	r2cryptoAAD       = "\x01\x02\x03\x04\x05\x06\x07\x08\t\n\x0b\x0c\r\x0e\x0f\x10"
)

func r2send(addr netip.AddrPort, timeout time.Duration, pkt []byte) error {
	conn, err := net.DialUDP("udp", nil, net.UDPAddrFromAddrPort(addr))
	if err != nil {
		return fmt.Errorf("connect to server: %w", err)
	}
	defer conn.Close()

	t := time.Now().Add(timeout)
	conn.SetWriteDeadline(t)
	conn.SetReadDeadline(t)

	pkt, err = r2cryptoEncrypt(pkt)
	if err != nil {
		return fmt.Errorf("encrypt connection packet: %w", err)
	}
	if _, err := conn.Write(pkt); err != nil {
		if errors.Is(err, os.ErrDeadlineExceeded) {
			err = fmt.Errorf("%w: %v", ErrTimeout, err)
		}
		return fmt.Errorf("send connection packet: %w", err)
	}
	return nil
}

func r2cryptoEncrypt(b []byte) ([]byte, error) {
	pkt := make([]byte, r2cryptoNonceSize+r2cryptoTagSize+len(b))

	nonce := pkt[:r2cryptoNonceSize]
	if _, err := rand.Read(nonce); err != nil {
		return nil, err
	}

	c, err := aes.NewCipher([]byte(r2cryptoKey))
	if err != nil {
		panic(err)
	}

	a, err := cipher.NewGCMWithTagSize(c, r2cryptoTagSize)
	if err != nil {
		panic(err)
	}

	// Go appends the ciphertext, then the tag to the dest (nonce)
	tmp := a.Seal(nil, nonce, b, []byte(r2cryptoAAD))
	copy(pkt[r2cryptoNonceSize:], tmp[len(b):])
	copy(pkt[r2cryptoNonceSize+r2cryptoTagSize:], tmp)
	return pkt, nil
}
```

pg9182:
For reference, `A`, `H`, `I`, and `N` are currently already taken as connectionless packet types.

image

pg9182 (commented):
Tested:
- Game server connection success.
- Game server connection rejection.
- Pdata fetch/update.
- Error handling when master server returns an error for `/server/connect`.
- Ignoring connection packets with invalid signatures.
- Connecting from v1.12.1 and #433.
- Fatal error during startup if bcrypt fails to initialize.
- Wine dedicated server startup.
- Artificially delayed response to connect packet, connect packet deduplication.

pg9182:
Since I've deployed R2Northstar/Atlas#31, this can be tested on the main masterserver.

pg9182:
Currently debugging an issue where `curl_easy_perform` hangs when running in Docker.

pg9182 (commented):
Tested:
- Game server connection success.
- Game server connection rejection.
- Pdata fetch/update.
- Error handling when master server returns an error for `/server/connect`.
- Ignoring Atlas connectionless packets with invalid signatures.
- Connecting from v1.12.1 and #433.
- Fatal error during startup if bcrypt fails to initialize.
- Artificially delayed response to connect packet, connect packet deduplication.
- Multiple servers on the same IP.
- Everything above, in docker (`docker run --rm -it -p 37017:37017/udp -e NS_PORT=37017 -e NS_SERVER_NAME="auth-udp {{hostname}}" -v /titanfall2:/mnt/titanfall:ro ghcr.io/pg9182/northstar-dedicated-ci:1-tf2.0.11.0-ns1.12.1-ci.c14bdc62.72.4340013708-nslauncher.7d72570a.4339994660`).

pg9182:
Updated to split pdata and connection response into two requests, improve logging.

pg9182 (commented):
Tested:
- Game server connection success.
- Game server connection rejection.
- Pdata fetch/update.
- Error handling when master server returns an error for `/server/connect`.
- Ignoring Atlas connectionless packets with invalid signatures.
- Connecting from v1.12.1 and #433.
- Fatal error during startup if bcrypt fails to initialize.
- Artificially delayed response to connect packet, connect packet deduplication.
- Multiple servers on the same IP.
- Everything above, in docker (`docker run --rm -it -p 37017:37017/udp -e NS_PORT=37017 -e NS_SERVER_NAME="auth-udp {{hostname}}" -v /titanfall2:/mnt/titanfall:ro ghcr.io/pg9182/northstar-dedicated-ci:1-tf2.0.11.0-ns1.12.1-ci.c14bdc62.75.4340581292-nslauncher.eef946ac.4340561319`).

pg9182:
This is now completely ready for review. Note that unless @BobTheBob9 has objections, I'm planning to merge this with rebase to preserve the history (each commit compiles and works).

NorthstarDLL/server/servernethooks.cpp
```diff
@@ -0,0 +1,208 @@
...
+	}
+
+	if (dbg)
+		spdlog::warn("ignoring Atlas connectionless packet (size={} type={}): unknown type", packet->size, pType);
+	return;
+}
+
+AUTOHOOK(ProcessConnectionlessPacket, engine.dll + 0x117800, bool, , (void* a1, R2::netpacket_t* packet))
+{
+	if (4 < packet->size && packet->data[4] == 'T')
```
pg9182, line 169: @BobTheBob9, is `T` acceptable?

BobTheBob9: don't see why not? might be better to have these as named consts in code rather than just magic chars though

BobTheBob9 (changes requested):
not actually tested yet, may be more things but for now looks pretty great

NorthstarDLL/server/servernethooks.cpp
pg9182: It is only used in one place, so I do think it's a bit more explicit to keep it here. Will add a comment though.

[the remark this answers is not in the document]
pg9182: Will add a comment as mentioned above.

NorthstarDLL/server/servernethooks.cpp
```diff
@@ -0,0 +1,208 @@
...
+	{
+		ProcessAtlasConnectionlessPacket(packet);
+		return false;
+	}
```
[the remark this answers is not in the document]
pg9182: Okay.

NorthstarDLL/server/servernethooks.cpp
```diff
@@ -0,0 +1,208 @@
...
+		"Whether to log detailed debugging information for Atlas connectionless packets (warning: this allows unlimited amounts of "
+		"arbitrary data to be logged)");
+
+	Cvar_net_debug_atlas_packet_insecure = new ConVar(
+		"net_debug_atlas_packet_insecure",
+		"0",
+		FCVAR_NONE,
+		"Whether to disable signature verification for Atlas connectionless packets (DANGEROUS: this allows anyone to impersonate Atlas)");
+
+	if (Cvar_net_debug_atlas_packet_insecure->GetBool())
```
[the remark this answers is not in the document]
pg9182: Good catch; I haven't checked if that message actually shows yet. I'll probably just remove it since it'd get buried anyways, and we warn when receiving the packet too (plus nobody will ever know this exists unless they dig through the code).

pg9182:
Latest image: `ghcr.io/pg9182/northstar-dedicated-ci:1-tf2.0.11.0-ns1.12.1-ci.c14bdc62.77.4350815031-nslauncher.b7b0c4c9.4346821849`.

pg9182:
Rebased.

pg9182:
For testing on v1.11.x, see #441.

pg9182 (commented):
Works well after some public testing.

- 703/704 successful auth requests.
- Have been running my own servers for a few days.
- Other people (usually ones having trouble with the current method) have tried it for theirs with success.
- 97th percentile auth time is the same or lower than the HTTP auth.

GeckoEidechse (approved):
Given that it was tested in public for a while now, I'm happy to sign off the testing part.

pg9182:
Rebased.

NorthstarDLL/server/servernethooks.cpp
```diff
@@ -0,0 +1,218 @@
...
+}
+
+AUTOHOOK(ProcessConnectionlessPacket, engine.dll + 0x117800, bool, , (void* a1, R2::netpacket_t* packet))
+{
+	// packet->data consists of 0xFFFFFFFF (int32 -1) to indicate packets aren't split, followed by a header consisting of a single
+	// character, which is used to uniquely identify the packet kind. Most kinds follow this with a null-terminated string payload
+	// then an arbitrary amoount of data.
+
+	// T (no rate limits since we authenticate packets before doing anything expensive)
+	if (4 < packet->size && packet->data[4] == 'T')
```
BobTheBob9, line 178: a constant rather than just 'T' here would be nice, but not required

BobTheBob9 (approved):

NorthstarDLL/server/servernethooks.cpp
pg9182: figured it was more clear to be able to see all the types in one function

PR #433, opened 2023-03-05
Use detailed master server error messages when available

pg9182:
The error.msg field has always been around, but we never used it.

recommends R2Northstar/NorthstarMods#598 (to show it for server auth as well as self auth)


pg9182:
Rebased.

BobTheBob9 (approved):
looks good

PR #434, opened 2023-03-05
Reject banned players during masterserver auth

### LINKED ISSUE #431: Check banlist during masterserver auth

pg9182:
This is backwards-compatible with old masterserver versions (it will be treated as a general authentication failure).

closes #431 
recommends #433 (client support)
recommends R2Northstar/Atlas#30 (atlas support)
recommends R2Northstar/NorthstarMods#598 (client support)

### DIFF
--- a/NorthstarDLL/server/auth/serverauthentication.cpp
+++ b/NorthstarDLL/server/auth/serverauthentication.cpp
@@ -74,7 +74,7 @@ void ServerAuthenticationManager::StartPlayerAuthServer()
 					}
 					if (!g_pBanSystem->IsUIDAllowed(uid))
 					{
-						response.set_content("{\"success\":false,\"reject\":\"Banned from server.\"}", "application/json");
+						response.set_content("{\"success\":false,\"reject\":\"Banned from this server.\"}", "application/json");
 						return;
 					}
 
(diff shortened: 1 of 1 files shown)


pg9182 (commented):
**If client doesn't have #433 or masterserver doesn't have R2Northstar/Atlas#30 or mods doesn't have TODO:**
image

**If masterserver has R2Northstar/Atlas#30:**
image

**If client has #433 and masterserver has R2Northstar/Atlas#30 and mods has TODO:**
image

NorthstarDLL/server/auth/serverauthentication.cpp
```diff
@@ -62,6 +62,22 @@ void ServerAuthenticationManager::StartPlayerAuthServer()
...
+						uid = std::strtoull(request.get_param_value("id").c_str(), nullptr, 10);
+					}
+					catch (std::exception const& ex)
+					{
+						response.set_content("{\"success\":false}", "application/json");
+						return;
+					}
+					if (!g_pBanSystem->IsUIDAllowed(uid))
+					{
+						response.set_content("{\"success\":false,\"reject\":\"Banned from server.\"}", "application/json");
```
GeckoEidechse, line 77:
```suggestion
						response.set_content("{\"success\":false,\"reject\":\"Banned from this server.\"}", "application/json");
```

Would slightly change the phrasing to ensure that we don't have players thinking they received a global Northstar ban (which doesn't exists but they may not know that). Proper localisation is separate PR anyway as discussed.

pg9182: Changed in both places.

NorthstarDLL/server/auth/serverauthentication.cpp
```diff
@@ -62,6 +62,22 @@ void ServerAuthenticationManager::StartPlayerAuthServer()
...
+					try
+					{
+						uid = std::strtoull(request.get_param_value("id").c_str(), nullptr, 10);
+					}
+					catch (std::exception const& ex)
+					{
+						response.set_content("{\"success\":false}", "application/json");
+						return;
+					}
+					if (!g_pBanSystem->IsUIDAllowed(uid))
```
BobTheBob9, line 75: why still call this if the strtoull call from earlier fails?

BobTheBob9 (changes requested):
pretty much great, consider doing the IsUIDAllowedCall in the same scope as strtoull i think?

NorthstarDLL/server/auth/serverauthentication.cpp
pg9182: We don't; there's a return in the try/catch.

BobTheBob9 (approved):

PR #437, opened 2023-03-07
me when case inconsistency

EladNLG:
one line fix baby!

`callingmodfunc` fails when a file path uses uppercase letters, because `ModManager::NormaliseModFilePath` is not used to convert uppercase letters to lowercase ones.

This is the fix.

das it :)


pg9182 (approved):
LGTM, matches other places where `g_pModManager->m_ModFiles.find` is called.

PR #443, opened 2023-04-01
Fix DecodeJSON fatalParseErrors param

EladNLG:
Fixes `DecodeJSON` failing to raise a script error if the document has parsing errors.

### DIFF
--- a/NorthstarDLL/scripts/scriptjson.cpp
+++ b/NorthstarDLL/scripts/scriptjson.cpp
@@ -212,10 +212,12 @@ ADD_SQFUNC(
 			doc.GetErrorOffset());
 
 		if (bFatalParseErrors)
+		{
 			g_pSquirrel<context>->raiseerror(sqvm, sErrorString.c_str());
-		else
-			spdlog::warn(sErrorString);
+			return SQRESULT_ERROR;
+		}
 
+		spdlog::warn(sErrorString);
 		return SQRESULT_NOTNULL;
 	}
 


EladNLG:
Surely this 4 line fix will not take 2 months to merge

F1F7Y (approved):
Works in testing, code looks good

NorthstarDLL/scripts/scriptjson.cpp
```diff
@@ -212,10 +212,12 @@ ADD_SQFUNC(
 			doc.GetErrorOffset());
 
 		if (bFatalParseErrors)
+		{
 			g_pSquirrel<context>->raiseerror(sqvm, sErrorString.c_str());
```
uniboi, line 216:
```suggestion
			return g_pSquirrel<context>->raiseerror(sqvm, sErrorString.c_str());
```

NorthstarDLL/scripts/scriptjson.cpp
```diff
@@ -212,10 +212,12 @@ ADD_SQFUNC(
...
-		else
-			spdlog::warn(sErrorString);
+			return SQRESULT_ERROR;
```
uniboi, line 217:
```suggestion
```

uniboi (approved):
code looks good just a visual improvement. This will generate a compiler warning because `raiseerror` is typed to return `SQInteger`, despite returning always `-1`, which is a valid `SQResult`.

NorthstarDLL/scripts/scriptjson.cpp
EladNLG: Not adding this because it isn't consistent with the rest of the repository.

NorthstarDLL/scripts/scriptjson.cpp
EladNLG: Not adding this because it isn't consistent with the rest of the repository.

NorthstarDLL/scripts/scriptjson.cpp
uniboi: L

GeckoEidechse:
Requesting review from @RoyalBlue1 cause she wrote the original code IIRC ^^

RoyalBlue1 (approved):
I didn't @BobTheBob9  did but it looks good

pg9182 (approved):
Looks good, seems to work fine.

Note that this probably shouldn't be included as a 1.12 patch (@GeckoEidechse) since it's technically a behaviour change (even though the old behaviour was broken).

PR #362, opened 2022-12-09
feat: Verified mods auto-downloading

Alystrasz:
This pull request introduces automatic mods fetching.

`NorthstarMods` associated pull request: https://github.com/R2Northstar/NorthstarMods/pull/542

---

This allows for the client to download and install mods that are required by a server (client+server-side mods, like new game mods or maps for instance). List of verified mods is local (compiled in the dll), but it can be fetched from master server in the future, to ease the addition of new verified mods.

With this feature, if users want to join a server requiring a gamemode mod or a map mod, their game can automatically download it for them, instead of forcing them to go through manual mod installation process (which can be hard for some people). While downloading a mod, progress download is displayed and live-updated on a dialog pop-up.

## Verified mod entries

A verified mod entry looks like this:

```json
{
    "Fifty.mp_frostbite": {
        "DependencyPrefix": "Fifty-Frostbite",
        "Versions": [
            { "Version": "0.0.1", "Checksum": "99fc4f0b3363e44985f20525a2bf079..." },
            { "Version": "0.0.2", "Checksum": "40780308e0e4a70f7f35f50f2ca188a..." }
        ]
    }
}
```

Versions are checked individually to ensure no malicious code has been added during mod updates.

Please note that it is up to the community to decide what's needed for a mod to be "verified", and how such mod is "verified".

Discussion here: https://github.com/R2Northstar/VerifiedMods

## Mod installation process

1. Check if mod is verified
    a. Check if it is listed
    b. Check if version is listed
2. Download mod archive from Thunderstore
    a. Fetch zip from API
    b. Compare checksum with local checksum
3. Unzip mod archive content directly into game folder
4. Reload mods list
5. Remove downloaded archive

Please note that downloaded archive is checksumed, to verify it is the very archive that has been verified.

#### `Northstar.Client`

* `OnServerSelected` method becomes async, to allow waiting for mod verification;
* `_OnServerSelected` method now displays some dialogs:
  * One error message when one attemps to join a server requiring a mod which cannot be auto-downloaded;
  * One loading message, displaying while a verified mod is being downloaded;
  * One error message when mod downloading/extraction failed;
* If a needed mod is already present but disabled, this activates it.

#### `Launcher`

* Includes a new library, `libzip`
* Includes the already-included but not linked `openssl` lib;
* All verified-mods-related code is stored in `verifiedmods.cpp`
* Squirrel-exposed methods:
  * `bool NSIsModVerified(string modName, string modVersion)`\: tells if a mod is verified or not;
[...]

PR #445, opened 2023-04-07
Improve logging for mods

ASpoonPlaysGames:
- Removes the log spam of "changing mod search path"
- Logs mods as they are loaded, alongside their ConVars, Scripts, ConCommands, etc.
- Warns in the console about errors in the `mod.json` instead of silently skipping things


GeckoEidechse:
Regarding:

> - Removes the log spam of "changing mod search path"

Do we still keep an indicator from where the mod is loaded from? Not saying we need to keep that message but given that with auto-download ( #362  ) we'll have a second location where we are loading mods from it would probably be useful to know still where the mod is loaded from ^^

ASpoonPlaysGames:
> Do we still keep an indicator from where the mod is loaded from?

Unsure, but I can always add it when the mod is loaded

ASpoonPlaysGames:
> Do we still keep an indicator from where the mod is loaded from?

This has now been done as of https://github.com/R2Northstar/NorthstarLauncher/pull/445/commits/41fa93d0620da5fb3a01354510169dfc06765f86

ASpoonPlaysGames:
> 1. This implies the conavrs were registered even though the mod is disabled

Yh i need to reword that a bit, they were registered from a modloading point of view, but not as far as the game is concerned

> 2. Don't use `goto`

`goto` is better to read than excessive amounts of nested conditionals and indentation. Normally I would split the different bits of modloading into different functions, and then simply `return` in the guard clauses, but splitting it up into different functions seemed unnecessary for something that is only referenced once

ASpoonPlaysGames:
@F1F7Y 's PR to this branch moved things over to their own functions

I have now fixed merge conflicts (which meant parsing InitScript) so I would consider this complete now I think

NorthstarDLL/core/filesystem/filesystem.cpp
```diff
@@ -73,8 +73,6 @@ void SetNewModSearchPaths(Mod* mod)
 	{
 		if ((fs::absolute(mod->m_ModDirectory) / MOD_OVERRIDE_DIR).string().compare(sCurrentModPath))
 		{
-			NS::log::fs->info("Changing mod search path from {} to {}", sCurrentModPath, mod->m_ModDirectory.string());
```
GeckoEidechse, line 76: Unrelated to PR, can someone explain me the difference between `NS::log::fs` and `spdlog::info`?

I know what the later does, just kinda confused why we also had our custom logger... ^^"

ASpoonPlaysGames: We have multiple different loggers for spdlog, all under `NS::log`.
`NS::log::fs` is for filesystem logs, and has the prefix `[FILESYSTM]` in logs
`NS::log::rpak` is for rpak filesystem logs, and has the prefix `[RPAK_FSYS]` in logs
etc. etc.

Basically its so we can get better context about where a line of logging came from. It was part of the log colours PR

GeckoEidechse: Aaah, makes sense. So, we use `spdlog` as generic log then? Or why the switch? ^^"

ASpoonPlaysGames: Yeah basically. `spdlog` I believe just uses the default logger (which is setup to prefix with `[NORTHSTAR]`)
There may be an `NS::log` variant for the same job, but meh

Alystrasz (approved):
Confirmed compiling and working in testing.
Meaningful logs will help a lot in tickets.

ASpoonPlaysGames:
image
fifty contributed to this PR, so probably not best if they review

Alystrasz:
Good catch!

F1F7Y:
I think i can improve on this a bit more, but think its good to be merged as is for now

GeckoEidechse:
Imma try to update branch to CMake, hopefully PR survives the change ^^

GeckoEidechse (approved):
Tested and confirmed:

- [x] Removes the log spam of "changing mod search path"
- [sorta] Logs mods as they are loaded, alongside their ConVars, Scripts, ~~ConCommands~~ (didn't have any example, someone gimme one please <3), etc.
- [x] Warns in the console about errors in the mod.json instead of silently skipping things

Honestly, if other contributors/modders want this fine by me. Some stuff is nice, kinda indifferent about the rest.

uniboi:
just merge it istg

PR #373, opened 2022-12-17
Add script concommands

emma-miler:
Example: 
```
"ConCommands": [
	{
	  "Name": "concommand_test",
	  "Function": "ConCommandTest",
	  "Context": "UI",
	  "Flags": 0,
	  "HelpString": "example script concommand"
	}
  ],
  ```

PR #446, opened 2023-04-08
Fix incorrect script ConCommand argc

Glacir:
Calling a ConCommand added in `mod.json` (introduced in #373) with 0 args raises an error

```
> concommand_test
> SCRIPT ERROR: [UI] wrong number of parameters (calling ConCommandTest): found 2, expected 1
```
where `ConCommandTest()` is a Squrriel function with **0** argument. `command.ArgC()` is 1 when no arg is inputted to console so wrong args are passed to squirrel.

Here's a example of my mod.json.
```json
"ConCommands": [
	{
	  "Name": "concommand_test",
	  "Function": "ConCommandTest",
	  "Context": "UI",
	  "Flags": 0,
	  "HelpString": "example script concommand"
	}
  ],
```

Tested:
- ConCommand with 0 and 1 arg works fine.

BTW: args passed to squirrel is always a array of string so `command.ArgC()` is always 2 if more than 0 arg is inputted in console. Not the point of this pr but should we change it?

### DIFF
--- a/NorthstarDLL/mods/modmanager.cpp
+++ b/NorthstarDLL/mods/modmanager.cpp
@@ -298,11 +298,18 @@ template <ScriptContext context> auto ModConCommandCallback_Internal(std::string
 {
 	if (g_pSquirrel<context>->m_pSQVM && g_pSquirrel<context>->m_pSQVM)
 	{
-		std::vector<std::string> args;
-		args.reserve(command.ArgC());
-		for (int i = 1; i < command.ArgC(); i++)
-			args.push_back(command.Arg(i));
-		g_pSquirrel<context>->AsyncCall(name, args);
+		if (command.ArgC() == 1)
+		{
+			g_pSquirrel<context>->AsyncCall(name);
+		}
+		else
+		{
+			std::vector<std::string> args;
+			args.reserve(command.ArgC());
+			for (int i = 1; i < command.ArgC(); i++)
+				args.push_back(command.Arg(i));
+			g_pSquirrel<context>->AsyncCall(name, args);
+		}
 	}
 	else
 	{


emma-miler (approved):
Cannot test, but looks fine to me

NorthstarDLL/mods/modmanager.cpp
```diff
@@ -298,11 +298,18 @@ template <ScriptContext context> auto ModConCommandCallback_Internal(std::string
...
 	if (g_pSquirrel<context>->m_pSQVM && g_pSquirrel<context>->m_pSQVM)
 	{
-		std::vector<std::string> args;
-		args.reserve(command.ArgC());
-		for (int i = 1; i < command.ArgC(); i++)
-			args.push_back(command.Arg(i));
-		g_pSquirrel<context>->AsyncCall(name, args);
+		if (command.ArgC() == 1)
+		{
+			g_pSquirrel<context>->AsyncCall(name);
```
pg9182, line 303: Will need to ensure this doesn't break existing functions.

NorthstarDLL/mods/modmanager.cpp
```diff
@@ -298,11 +298,18 @@ template <ScriptContext context> auto ModConCommandCallback_Internal(std::string
...
+		}
+		else
+		{
+			std::vector<std::string> args;
+			args.reserve(command.ArgC());
+			for (int i = 1; i < command.ArgC(); i++)
```
pg9182, line 309: > @Glacir: BTW: args passed to squirrel is always a array of string so command.ArgC() is always 2 if more than 0 arg is inputted in console. Not the point of this pr but should we change it?

If it means that this doesn't actually work, then yes.

Glacir: It works but not intuitive. For modders who register concommand, what they expect is register a concommand like
```
command [string a] [string b]
```
and link it to a squirrel function

```
void function foo(string a, string b)
```

But the current logic in ns launcher pass the args as an `array<string>` so modders need to write their squirrel function like this:

```
void function foo(array<string> args)
```

NorthstarDLL/mods/modmanager.cpp
pg9182: What happens if the squirrel function takes an argument, but argc is 0?

NorthstarDLL/mods/modmanager.cpp
pg9182: Ah, I suppose I misunderstood what you meant. That makes sense.

NorthstarDLL/mods/modmanager.cpp
Glacir: `argc` is always >= 1. I suppose what you mean is if no arg is inputted for a concommand that requires an argument? If so, it raises an error like the situation I posted.

pg9182: In that case, it would probably be a good idea to see if we can check the squirrel function arg count so we don't break existing users.

pg9182: Also, why not just pass an empty array?

pg9182 (changes requested):
I took another look at this, and I don't think this change is necessary. It's not really a bugfix, it's a behaviour change which could break existing mods, and will cause errors if you pass zero arguments to a concommand expecting some, or some arguments to a concommand expecting none.

I'm not sure exactly what would be "broken" about the existing code except for the fact that it reserves one more array element than strictly necessary. The fact that we pass an array of arguments (like the way `main` functions work) isn't a bug IMO.

NorthstarDLL/mods/modmanager.cpp
pg9182: Took another look at this; it's okay.

pg9182: It's okay, nobody really uses this yet, so it shouldn't have a huge impact.

pg9182 (approved):
Will technically break mods which accept 0 args and had a squirrel function which took an empty array as a workaround, but this shouldn't impact most (if any) existing mods based on asking on Discord.

PR #448, opened 2023-04-16
Fix getvector returning garbage data

uniboi:
`sq_getvector+0x6120` returns a pointer to a vector:
```c
longlong sq_getvector(longlong sqvm,int stackpos)

{
  return *(longlong *)(sqvm + 0x48) + 4 + (longlong)stackpos * 0x10;
}
```
This PR fixes the casting in `squirrel.h`


pg9182 (commented):
Makes sense, and doesn't seem to cause issues. Not my area of expertise though, so waiting for @RoyalBlue1 or @BobTheBob9 to look before merging.

RoyalBlue1 (approved):
Code looks good fixes an obvious error

PR #449, opened 2023-04-19
Remove leftover LoadPlugins declaration

pg9182:
Doesn't exist since 450d0b1ed437cf37b4309af952af8904f3f07768 (#343).


pg9182 (commented):
No need to test; this function isn't defined anywhere and was explicitly removed in #343.

GeckoEidechse (approved):
Removal makes sense based on the commit linked in PR description and obviously game still launches fine.

PR #450, opened 2023-04-19
Fix parsing string ConVar/ConCommand.Flags from `mod.json`

pg9182:
The ParseConVarFlagsString function introduced in 64100065b55f79e76542ba689545c60e6fb0dcef (#373) is utterly broken. It only parses the first flag, logs misleading warnings, has an undefined return value in some codepaths, and is somewhat convoluted.

Luckily, this doesn't [appear](https://github.com/search?q=path%3A**%2Fmod.json+ConVars+Flags&type=code) to affect most (if not all) existing mods, as they all seem to be using integer values for Flags, which is taken as-is.

### DIFF
--- a/NorthstarDLL/core/convar/convar.cpp
+++ b/NorthstarDLL/core/convar/convar.cpp
@@ -497,40 +497,36 @@ bool ConVar::ClampValue(float& flValue)
 
 int ParseConVarFlagsString(std::string modName, std::string sFlags)
 {
-	sFlags += '|'; // add additional | so we register the last flag
-	std::string sCurrentFlag;
+	int iFlags = 0;
+	std::stringstream stFlags(sFlags);
+	std::string sFlag;
 
-	for (int i = 0; i < sFlags.length(); i++)
+	while (std::getline(stFlags, sFlag, '|'))
 	{
-		if (isspace(sFlags[i]))
+		// trim the flag
+		sFlag.erase(sFlag.find_last_not_of(" \t\n\f\v\r") + 1);
+		sFlag.erase(0, sFlag.find_first_not_of(" \t\n\f\v\r"));
+
+		// skip if empty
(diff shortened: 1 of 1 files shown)


pg9182 (commented):
Tested with:

```cpp
int main() {
	printf("%d\n", ParseConVarFlagsString("", ""));
	printf("%x\n", ParseConVarFlagsString("", "RELOAD_MATERIALS"));
	printf("%x\n", ParseConVarFlagsString("", "HIDDEN"));
	printf("%x\n", ParseConVarFlagsString("", "RELEASE"));
	printf("%x\n", ParseConVarFlagsString("", "RELOAD_MATERIALS|HIDDEN"));
	printf("%x\n", ParseConVarFlagsString("", "RELOAD_MATERIALS | HIDDEN"));
	printf("%x\n", ParseConVarFlagsString("", "RELOAD_MATERIALS | HIDDEN | RELEASE"));
	printf("%x\n", ParseConVarFlagsString("", "RELOAD_MATERIALS | HIDDENA"));
	printf("%x\n", ParseConVarFlagsString("", "RELOAD_MATERIALS | HIDDENA | RELEASE"));

	return 0;
}
```

<table><thead><tr><th>Before</th><th>After</th></tr></thead><tbody><tr><td>

```
Mod ConCommand  has unknown flag 
Illegal instruction (core dumped)
--- after removing the first testcase
100000
10
80000
100000
100000
100000
100000
100000
```

</td><td>

```
0
100000
10
80000
100010
100010
180010
Mod ConCommand  has unknown flag HIDDENA
100000
Mod ConCommand  has unknown flag HIDDENA
180000
```

</td></tbody></table>

RoyalBlue1 (approved):
did not try to compile or run but the logic in the code seems sound and there are no obvious errors

PR #451, opened 2023-04-21
Search for Northstar.dll in profile directory

F1F7Y:
Looks for `Northstar.dll` in the specified profile directory using `-profile=<dir>`. If it doesn't exist it defaults to the `Northstar.dll` in the root directory.


GeckoEidechse:
Hol' up this looks so easy, why haven't we been doing this earlier?! o_O

Genuinely awesome idea ngl

Given that the `NorthstarLauncher.exe` barely gets changed anyway it should stay mostly compatible between releases if combined with other versions of `Northstar.dll` 👀 

Now thing is as much as I want this ASAP there's a few things to consider:

Most importantly

- [ ] Release repo CI needs to be updated to put DLL into right folder (I could do that if you don't wanna, just LMK)
- [ ] Mod managers will need to support the change (obviously the location in the zip folder would be updated in release repo so that it's in right profile so it's mostly a question of whether mod-managers check for it and maybe have some logic to get rid of old DLL
  - [ ] @0neGal for Viper
  - [ ] @BigSpice for VTOL
  - [ ] @GeckoEidechse for FlightCore (oh hey, that's me xD)
  - [ ] r2modman probably also needs to be updated given that it puts mods into profile in its dir and places exe in Titanfall2 folder
- [ ] General checking whether this breaks stuff?
- [ ] ???

GeckoEidechse:
Actually as a temporary switchover we could search both paths (old and new) where _new_ is preferred but _old_ is used as fallback and then later deprecated old location after a month or two.

ASpoonPlaysGames:
> * Release repo CI needs to be updated to put DLL into right folder

Does it? Isn't the main benefit of this for RCs and dev work? Surely we should continue to put the dll where it currently is, except for when building/installing an RC or a version using a mod manager?

F1F7Y:
>     * [ ]  Release repo CI needs to be updated to put DLL into right folder (I could do that if you don't wanna, just LMK)

I fall back to the `Northstar.dll` in the root directory so this is completely optional

F1F7Y:
> Actually as a temporary switchover we could search both paths (old and new) where _new_ is preferred but _old_ is used as fallback and then later deprecated old location after a month or two.

Skipped over this originally. We do fallback to the original path when a `Northstar.dll` doesnt exist in the profile, or when no profile is passed.

Thinking about it if you put a `Northstar.dll` into `R2Northstar` it and dont provide a `-profile=R2Northstar` it'll fallback to the original and wont load the profile one.

F1F7Y:
> Thinking about it if you put a `Northstar.dll` into `R2Northstar` and dont provide a `-profile=R2Northstar` it'll fallback to the original and wont load the profile one.

Fixed in 20d4360

F1F7Y:
Current behaviour:
1. No args passed:
  Looks for `Northstar.dll` in `R2Northstar`
2. A profile passed
  Looks for `Northstar.dll` in the profile dir
  
If we dont find `R2Northstar` we fallback to the one in the root directory

ASpoonPlaysGames (approved):
Not tested, but code looks good, and I like the idea a lot

GeckoEidechse (changes requested):
> 2\. A profile passed
>         Looks for `Northstar.dll` in the profile dir

Note that to my knowledge this will break r2modman. Cause it runs `NorthstarLauncher.exe` with `-profile` pointing to the mods folder in its install dir. Not sure if moving the the dll into `R2Northtstar` folder is enough to fix that, I'd have to check r2modman source for that.

ASpoonPlaysGames:
> > 2. A profile passed
> > Looks for `Northstar.dll` in the profile dir
> 
> Note that to my knowledge this will break r2modman. Cause it runs `NorthstarLauncher.exe` with `-profile` pointing to the mods folder in its install dir. Not sure if moving the the dll into `R2Northtstar` folder is enough to fix that, I'd have to check r2modman source for that.

No, because it falls back to root directory if it's not there in the profile dir

pg9182 (approved):
Approving this for now since it's useful and backwards-compatible, but I am working on reworking the launch process and path handling (separating core and custom stuff, separating titanfall and the launcher, unifying profile path resolution), so let's not overload stuff too much more.

Not tested, but code looks okay.

GeckoEidechse:
> > Note that to my knowledge this will break r2modman. Cause it runs `NorthstarLauncher.exe` with `-profile` pointing to the mods folder in its install dir. Not sure if moving the the dll into `R2Northtstar` folder is enough to fix that, I'd have to check r2modman source for that.
> 
> No, because it falls back to root directory if it's not there in the profile dir

Oh right, forgot about that, disregard my comment above then ^^"

uniboi (approved):
Works in testing

0neGal:
> * Mod managers will need to support the change (obviously the location in the zip folder would be updated in release repo so that it's in right profile so it's mostly a question of whether mod-managers check for it and maybe have some logic to get rid of old DLL

Problem is, how would we go about doing this, mod managers would have to store the file structure of the archive upon installation, and then delete files when missing in newer versions. Either that or have a file in the archive with files of which to delete before/after installation.

As I do believe most if not all mod managers just extract the archive into the install folder in some way or form, without deleting any other files.

Of course, we could all just add in a check to delete the old DLL, but that sounds a lot more like a workaround than an actual solution, given this will quite possibly happen again sooner or later.

Northstar itself could also just delete the old file if it finds it, but uncertain if that's any good of a solution.

ebkr:
> > 2. A profile passed
> > Looks for `Northstar.dll` in the profile dir
> 
> Note that to my knowledge this will break r2modman. Cause it runs `NorthstarLauncher.exe` with `-profile` pointing to the mods folder in its install dir. Not sure if moving the the dll into `R2Northtstar` folder is enough to fix that, I'd have to check r2modman source for that.

What should we be doing with the DLL? It was built to the spec at the time of original development, so what’s changed in terms of how we provide -profile and how is the DLL stored there when pulled from TS?

If it’s in the root of the package then we should just support it? Has the target for -profile changed/we’re currently using a legacy approach?


> As I do believe most if not all mod managers just extract the archive into the install folder in some way or form, without deleting any other files.


Could be wrong but I’m pretty sure r2mm handles it as we have a state tracker that logs the current state of all the files

GeckoEidechse:
> > > 2. A profile passed
> > >    Looks for `Northstar.dll` in the profile dir
> > 
> > 
> > Note that to my knowledge this will break r2modman. Cause it runs `NorthstarLauncher.exe` with `-profile` pointing to the mods folder in its install dir. Not sure if moving the the dll into `R2Northtstar` folder is enough to fix that, I'd have to check r2modman source for that.
> 
> What should we be doing with the DLL? It was built to the spec at the time of original development, so what’s changed in terms of how we provide -profile and how is the DLL stored there when pulled from TS?
> 
> If it’s in the root of the package then we should just support it? Has the target for -profile changed/we’re currently using a legacy approach?

So basically currently the package for Northstar looks like
```
.
├── icon.png
├── manifest.json
├── Northstar
│   ├── bin
│   │   └── <some stuff>
│   ├── LEGAL.txt
│   ├── Northstar.dll   <--- this would be moved in the future
│   ├── NorthstarLauncher.exe
│   ├── r2ds.bat
│   └── R2Northstar
│       ├── mods
│       │   └── <some stuff>
│       └── plugins
│           └── <some stuff>
└── README.md
```

and in the future we'd change it to
```
.
├── icon.png
├── manifest.json
├── Northstar
│   ├── bin
│   │   └── <some stuff>
│   ├── LEGAL.txt
│   ├── NorthstarLauncher.exe
│   ├── r2ds.bat
│   └── R2Northstar
│       ├── Northstar.dll   <---- now moved here
│       ├── mods
│       │   └── <some stuff>
│       └── plugins
│           └── <some stuff>
└── README.md
```

So if r2mm would already support that, then there's no change necessary. If not we'll keep it in mind and will have to update r2mm accordingly before pushing a changed package structure in Northstar.

(Note that this PR is not about moving the DLL itself but simply about enabling moving it in the future ^^)

ebkr:
> So if r2mm would already support that, then there's no change necessary.

As far as I'm aware it should just work out of the box. We do things like compare what's copied across and remove things that aren't specified anyway.

Assuming we don't need to change the target of the mods directory then no change would be necessary from what I can tell as I'm pretty sure it just extracts the zip into the manager's profile dir.

F1F7Y:
Added `depends on another PR` as this depends on mod managers to add support ( even though none of them properly support profiles to this date yet )

BigSpice:
I personally have not been explained exactly how they work. I'm putting work tn, some pointers would be wonderful 👀

F1F7Y:
> I personally have not been explained exactly how they work. I'm putting work tn, some pointers would be wonderful 👀

The idea of profiles is, in short, just renaming `R2Northstar`. Let's say you pass `-profile="Testing"`, now instead of using the `R2Northstar` folder we look for mods, plugins in `Testing` folder.

This PR adds this functionality for loading the `Northstar.dll`. Lets say you set the profile directory to `Testing`. NorthstarLauncher will look for the dll in that directory, if it doesn't find it in there it'll try to load the one in the root ( where we place it rn ), if even that fails it exits.

GeckoEidechse:
> Added `depends on another PR` as this depends on mod managers to add support ( even though none of them properly support profiles to this date yet )

Looking back at this again. This PR just adds functionality to search for DLL in the profile folder but does not actually move it yet. So we can probably just skip most of the stuff in [my original comment](https://github.com/R2Northstar/NorthstarLauncher/pull/451#issuecomment-1518202024) and only do that when we actually move the DLL.

GeckoEidechse:
bump

PR #452, opened 2023-04-22
Replace `set-output` in GitHub Actions workflows

pg9182:
`set-output` is being deprecated as per https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


GeckoEidechse (approved):
Looks good

image

image

PR #453, opened 2023-04-22
Use structs for the Script Server Browser

uniboi:
Adds two squirrel api functions to deal with structs
```cpp
SQRESULT pushnewstructinstance(HSquirrelVM* sqvm, const int fieldCount)
```
creates a new struct instance with `fieldCount` fields and pushes it to the stack

```cpp
SQRESULT sealstructslot(HSquirrelVM* sqvm, const int fieldIndex)
```
pops a value from the stack and sets it as the value of the now last (struct-) object at `fieldIndex`.
I'm not sure about the function name

__Example:__
```cpp
pushnewstructinstance(sqvm, 2);
pushinteger(sqvm, 19);
sealstructslot(sqvm, 0);
pushinteger(sqvm, 22);
sealstructslot(sqvm, 1);
```
This pushes a struct with two slots to the stack that contains two integers.

This PR cleans up the script serverbrowser native API a bit to make it easier to use.
Instead of accessing info about a server by index there is now only a function that returns an array of struct instances that hold all server info.


NorthstarDLL/scripts/client/scriptserverbrowser.cpp
```diff
@@ -6,299 +6,105 @@
...
-		return SQRESULT_ERROR;
+		g_pSquirrel<context>->arrayappend(sqvm, -2);
 	}
-
-	g_pSquirrel<context>->pushstring(sqvm, g_pMasterServerManager->m_vRemoteServers[serverIndex].id);
 	return SQRESULT_NOTNULL;
 }
 
-ADD_SQFUNC("bool", NSServerRequiresPassword, "int serverIndex", "", ScriptContext::UI)
+ADD_SQFUNC("bool", NSIsMasterServerAuthenticated, "", "", ScriptContext::UI)
```
RoyalBlue1, line 81: move this and the following functions to the top to make the diff better

NorthstarDLL/scripts/client/scriptserverbrowser.cpp
```diff
@@ -6,299 +6,105 @@
...
+		// region
+		g_pSquirrel<context>->pushstring(sqvm, remoteServer.region, -1);
+		g_pSquirrel<context>->sealstructslot(sqvm, 9);
 
-ADD_SQFUNC("int", NSGetServerMaxPlayerCount, "int serverIndex", "", ScriptContext::UI)
-{
-	SQInteger serverIndex = g_pSquirrel<context>->getinteger(sqvm, 1);
+		// requiredMods
+		g_pSquirrel<context>->newarray(sqvm);
+		for (auto mod : remoteServer.requiredMods)
```
RoyalBlue1, line 60: same here

NorthstarDLL/scripts/client/scriptserverbrowser.cpp
```diff
@@ -6,299 +6,105 @@
...
-		g_pSquirrel<context>->raiseerror(
-			sqvm,
-			fmt::format(
-				"Tried to get name of server index {} when only {} servers are available",
-				serverIndex,
-				g_pMasterServerManager->m_vRemoteServers.size())
-				.c_str());
-		return SQRESULT_ERROR;
-	}
+		RemoteServerInfo remoteServer = g_pMasterServerManager->m_vRemoteServers[i];
```
RoyalBlue1, line 14: This is not really needed but if you make this a const ref the code will be optimized better

RoyalBlue1 (approved):
The things I marked with a comment don't really need to be done but they make the code a bit better. The Squirrel API stuff looks good though. I didn't try to run it so there might be issues there

catornot (approved):
tested this and it worked :)
also I really like this rewrite since the index functions are a mess.

NorthstarDLL/scripts/client/scriptserverbrowser.cpp
uniboi: Oh I misunderstood. I moved the functions like you said to clean up the diff.

uniboi:
Just to clarify a bit more, I removed all squirrel functions that only *access* a value from a server by index but mods shouldn't be affected since I know of barely nobody that uses them. (Wanderer is the only one)

PR #457, opened 2023-04-23
Fix initscript issue

### LINKED ISSUE #456: Cannot run Northstar without mods in `v1.13.0`
Running Northstar without any mods (to e.g. play on vanilla servers) gives

image

catornot:
fixes the issue described by #456 by adding a check if the mods are actually enabled.

should be checked if there are any side effects from this ( from my testing there doesn't appear to be any )

## how to test this

after installing the pr disable `Northstar.Client` if you crash then the pr doesn't work, script errors will happen (because you disabled a core mod lol)

fixes #456

### DIFF
--- a/NorthstarDLL/squirrel/squirrel.cpp
+++ b/NorthstarDLL/squirrel/squirrel.cpp
@@ -415,7 +415,7 @@ template <ScriptContext context> bool __fastcall CSquirrelVM_initHook(CSquirrelV
 	bool ret = CSquirrelVM_init<context>(vm, realContext, time);
 	for (Mod mod : g_pModManager->m_LoadedMods)
 	{
-		if (mod.initScript.size() != 0)
+		if (mod.m_bEnabled && mod.initScript.size() != 0)
 		{
 			std::string name = mod.initScript.substr(mod.initScript.find_last_of('/') + 1);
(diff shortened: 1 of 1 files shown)


pg9182 (approved):
Northstar starts successfully (client, vanilla client, dedi). Check for `m_bEnabled` matches other `g_pModManager->m_LoadedMods` loops.

GeckoEidechse (approved):
Confirmed working in testing.

`v1.13.0` with all mods disabled has the error shown in #456, using this PR under the same setup launches just fine.

GeckoEidechse:
Apparently also fixed a script compile error for someone?

image

catornot:
I think it got fixed because they removed the vannila files that northstar doesn't override anymore

minh5052vn:
My login menu still has the console problem. There is nothing on the screen except of the console and the background animation.

ASpoonPlaysGames:
> My login menu still has the console problem. There is nothing on the screen except of the console and the background animation.
> Can we reverse back to the previous version?

Do you have any mods installed? A blank main menu other than the console is a very common issue with mods

ASpoonPlaysGames:
A log file would be helpful here

minh5052vn:
> 

I don't have any mods. Just 3 Northstar basic things.

PR #458, opened 2023-04-23
Add plugin dependency constants

catornot:
Adds dependency constants for plugins so mods can rely on plugins without always producing script errors when the plugin is missing

# testing

If you trust me :)
test plugin : [rrplug_test.zip](https://github.com/R2Northstar/NorthstarLauncher/files/11304684/rrplug_test.zip)
test mod : [cat_or_not.ProximityChat.zip](https://github.com/R2Northstar/NorthstarLauncher/files/11304686/cat_or_not.ProximityChat.zip)

expected result : no compile errors and a message on CLIENT load "we have RRPLUG_TEST" with the plugin


NorthstarDLL/squirrel/squirrel.cpp
```diff
@@ -255,6 +255,14 @@ template <ScriptContext context> void SquirrelManager<context>::VMCreated(CSquir
 
 		defconst(m_pSQVM, pair.first.c_str(), bWasFound);
 	}
+
+	auto loadedPlugins = &g_pPluginManager->m_vLoadedPlugins;
+	for (const auto& pluginName : g_pModManager->m_PluginDependencyConstants)
+	{
+		auto f = [&](Plugin plugin) -> bool { return plugin.dependencyName == pluginName; };
+		defconst(m_pSQVM, pluginName.c_str(), std::find_if(loadedPlugins->begin(), loadedPlugins->end(), f) != loadedPlugins->end());
```
pg9182, line 263: Can we scope this somehow, or maybe even define a function to check if a plugin is defined?

catornot: so you want this to be moved into it's own function?

pg9182: I'm wondering if there's a better way to do this without polluting the global namespace with arbitrarily-named constants.

catornot: I don't understand. there is no global variable?

pg9182: Oh, nevermind, I misinterpreted the code and didn't look at the sample mod as I was on my phone.

pg9182 (approved):
Tested the sample mod/plugin, seems to work as intended. Code looks good.

ASpoonPlaysGames:
@catornot Merge conflicts :)

ASpoonPlaysGames (changes requested):
I don't see any code that handles characters in the plugin name that aren't valid for squirrel constants.

Things like spaces and special characters should get handled, or the squirrel constants might become unusable

catornot:
> I don't see any code that handles characters in the plugin name that aren't valid for squirrel constants.

I was thinking that the plugin creator would be responsible for that but I guess I could add a warning.

ASpoonPlaysGames (approved):
Code looks good to me, requested change was implemented

GeckoEidechse:
Merging based on reviews

PR #459, opened 2023-04-25
Restore `-noplugins` command arg

catornot:
this command arg was removed in plugins v2

basically it prevents plugins from being loaded

### testing 
a batch file to start northstar with `-noplugins`
[noplugins-test.zip](https://github.com/R2Northstar/NorthstarLauncher/files/11327229/noplugins-test.zip)
can't upload bat files lol

### DIFF
--- a/NorthstarDLL/plugins/plugins.cpp
+++ b/NorthstarDLL/plugins/plugins.cpp
@@ -207,6 +207,11 @@ bool PluginManager::LoadPlugins()
 	data.version = ns_version.c_str();
 	data.northstarModule = g_NorthstarModule;
 
+	if (strstr(GetCommandLineA(), "-noplugins") != NULL)
+	{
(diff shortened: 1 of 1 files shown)


pg9182 (approved):
Works.

NorthstarDLL/plugins/plugins.cpp
```diff
@@ -207,6 +207,11 @@ bool PluginManager::LoadPlugins()
 	data.version = ns_version.c_str();
 	data.northstarModule = g_NorthstarModule;
 
+	if (strstr(GetCommandLineA(), "-noplugins") != NULL)
+	{
+		NS::log::PLUGINSYS->warn("-noplugins detected; skipping loading plugins");
+		return false;
+	}
```
GeckoEidechse, line 214: I assume there's a reason for this but why isn't the check at the beginning of the function? ^^

catornot: look at the if cases below

PR #460, opened 2023-04-25
Support loading plugins from subdirs

catornot:
allows plugins to be actually organized/grouped instead of just dumping them into the `plugins` folder
also useful for mod managers

### DIFF
--- a/NorthstarDLL/plugins/plugins.cpp
+++ b/NorthstarDLL/plugins/plugins.cpp
@@ -52,8 +52,8 @@ std::optional<Plugin> PluginManager::LoadPlugin(fs::path path, PluginInitFuncs* 
 
 	Plugin plugin {};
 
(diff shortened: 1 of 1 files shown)


GeckoEidechse (approved):
Kinda expected this to also load plugins from none `plugins/` folders but based on testing that doesn't seem to be the case.

I.e. I used `DiscordRPC.dll` for testing and putting it in the folllowing folders I got

| Location | Plugin loaded? |
| --- | --- |
| `R2Northstar/plugins/` | ✔️ |
| `R2Northstar/plugins/test` | ✔️ |
| `R2Northstar/plugins/test/test2` | ✔️ |
| `R2Northstar/` | ❌ |
| `R2Northstar/mods` | ❌ |
| `R2Northstar/mods/Northstar.Custom` | ❌ |

So I guess works as expected in testing ^^

PR #461, opened 2023-04-28
refactor sq api init

uniboi:
Makes initializing the SquirrelManagers more readable by having definitions for CLIENT, UI & SERVER vm in one place.
Currently it looks like this
```cpp
ON_DLL_LOAD_RELIESON("client.dll", ClientSquirrel, ConCommand, (CModule module))
{
	// ...

	g_pSquirrel<ScriptContext::CLIENT>->__sq_defconst = module.Offset(0x12120).As<sq_defconstType>();
	g_pSquirrel<ScriptContext::UI>->__sq_defconst = g_pSquirrel<ScriptContext::CLIENT>->__sq_defconst;

	// repeat for every function ...
}

ON_DLL_LOAD_RELIESON("server.dll", ClientSquirrel, ConCommand, (CModule module))
{
	g_pSquirrel<ScriptContext::SERVER>->__sq_defconst = module.Offset(0x12120).As<sq_defconstType>();

	// repeat for every function ...
}
```
The redundancy is annoying and makes it hard to see at a glance what the offsets for the functions are.

This PR implements this function that initializes the functions either on the `CLIENT` & `UI` or on the `SERVER` dynamically:

```cpp
// Initialize a function on the CLIENT and UI or SERVER, depending on context
template <ScriptContext context, typename T>
void SetSharedMember(SquirrelManager<context>* manager, T* member, const CModule module, const int clOffset, const int svOffset)
{
	switch (context)
	{
	case ScriptContext::CLIENT:
	{
		*member = module.Offset(clOffset).As<T>(); // Set the member for the CLIENT vm
		*(T*)((char*)g_pSquirrel<ScriptContext::UI> + (int64_t)member - (int64_t)manager) = *member; // Set the member for the UI vm
		break;
	}
	case ScriptContext::SERVER:
		*member = module.Offset(svOffset).As<T>();
	}
}
```

Initializing a function for all vms now looks like this:

```cpp
template <ScriptContext context> void InitSharedSquirrelFunctions(SquirrelManager<context>* manager, CModule module)
{
	SetSharedMember<context, sq_defconstType>(manager, &manager->__sq_defconst, module, 0x12120, 0x1F550);
}
```

The issue with this is that the memory layout of classes is ub afaik

### DIFF
--- a/NorthstarDLL/squirrel/squirrel.cpp
+++ b/NorthstarDLL/squirrel/squirrel.cpp
@@ -658,6 +658,83 @@ template <ScriptContext context> void SquirrelManager<context>::ProcessMessageBu
 	}
 }
 
+// Initialize a function on the CLIENT and UI or SERVER, depending on context
+template <ScriptContext context, typename T>
+void SetSharedMember(SquirrelManager<context>* manager, T* member, const CModule module, const int clOffset, const int svOffset)
+{
+	switch (context)
+	{
+	case ScriptContext::CLIENT:
+	{
+		*member = module.Offset(clOffset).As<T>(); // Set the member for the CLIENT vm
+		*(T*)((char*)g_pSquirrel<ScriptContext::UI> + (int64_t)member - (int64_t)manager) = *member; // Set the member for the UI vm
+		break;
+	}
+	case ScriptContext::SERVER:
+		*member = module.Offset(svOffset).As<T>();
+	}
+}
+
+template <ScriptContext context> void InitSharedSquirrelFunctions(SquirrelManager<context>* manager, CModule module)
+{
+	// consts & root table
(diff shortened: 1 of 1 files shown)


catornot (approved):
works in testing, plugins still work too.
refactoring good :)

pg9182 (changes requested):
Doing this until I have time to look at it more closely.

- Need to check each refactored line.
- If the UB depends on the layout of TF stuff, it's fine, but if it's on our own classes, it's not worth it.

GeckoEidechse:
\*bump\*

GeckoEidechse:
@uniboi just a heads-up that this PR has some merge conflicts atm

GeckoEidechse:
merge conflicts still

PR #465, opened 2023-06-01
Don't attempt masterserver auth on lobby

ASpoonPlaysGames:
(unless we are in private match)

Currently, we attempt to auth with the masterserver when we are in the lobby (non private match). This is fine for servers who have port forwarding set up, and Atlas just doesn't show these servers in the server browser anyway.

However for normal users who aren't familiar with how servers work and are experiencing issues with northstar, the constant errors in a normal lobby saying that they failed masterserver registration tend to be a bit of a red herring, they see log spam of errors and think that it is the source of their issue.

If we ever decide to implement lobby servers then this might cause issues, but honestly those should be treated differently from a normal user in a lobby, so the extra logic would already be needed imo.


ASpoonPlaysGames:
Just realised this is kinda hard to test, because the server wouldn't appear in the server browser anyway.

To test it's probably best if you don't have port forwarding set up (so you would normally get a bunch of errors from the MS) because then there is a visible difference in logs.

NoCatt (approved):
When in the regular lobby, it will not try to auth to the master server and will not print the error in the console ( as the request was never made ).
When you enter a private match it will try to auth ( and fail or be sucessfull ).

PR #467, opened 2023-06-06
Move to cmake

F1F7Y:
Moves the entire project to cmake

!! Apart from the build directory this doesn't modify any source files or build tools !!

Both Release and Debug builds are built into `${CMAKE_BINARY_DIR}/game/` ( wsock32 is built into bin/x64_retail ). This means devs can move their game installation into the `game` directory and launch northstar directly from visual studio.

- [x] Move to cmake
- [x] Update `ci.yml`
- [x] Update `release.yml`
  - !! Untested

The goal of this PR is to just move to cmake, no compiler changes, no complicated cmake macros.


ASpoonPlaysGames:
> !! Apart from the build directory this doesn't modify any source files or build tools !!

This is very nice, makes reviewing and testing much easier

GeckoEidechse:
Holy shit this is really fucking cool :O

loader_wsock32_proxy/CMakeLists.txt
```diff
@@ -0,0 +1,49 @@
...
+
+target_precompile_headers(loader_wsock32_proxy PRIVATE pch.h)
+
+target_compile_definitions(loader_wsock32_proxy PRIVATE
+                           UNICODE
+                           _UNICODE
+)
+
+set_target_properties(loader_wsock32_proxy PROPERTIES
+                      RUNTIME_OUTPUT_DIRECTORY ${CMAKE_SOURCE_DIR}/game/bin/x64_retail
```
GeckoEidechse, line 46: Is `game` a common artifact folder name? Haven't interacted too much with CMake, I just know from past experience that one usually calls it `build` or `target` but I could be mistaken here ^^

Basically just wanna use an output folder name that is a standard as possible so that someone with CMake experience finds the build artifacts in the folder they expected.

GeckoEidechse (changes requested):
Make sure to also update `[BUILD.md](https://github.com/R2Northstar/NorthstarLauncher/blob/main/BUILD.md)` to explain how to import the project and build it ^^

AFAIK wiki just links to that file in regards to build instructions, not sure what ModdingDocs does ( cc @laundmo @uniboi for ModdingDocs)

loader_wsock32_proxy/CMakeLists.txt
F1F7Y: Cmake projects usually use folders named `bin` or `build` for output. I chose `game` mainly because r5reloaded uses the same structure

pg9182: cmake already has a concept of a build dir, so we shouldn't be overriding that IMO

F1F7Y: When using Visual studio as the generator CMAKE_BINARY_DIR defaults to `out/build/CMAKE_BUILD_TYPE/`. As I said i chose to output into `game` so that one can copy their game over and launch from vs directly.

I did try setting CMAKE_BINARY_DIR to `${CMAKE_SOURCE_DIR}/game`, but that seemed to default to the visual studio path ??, unsure what was happening there, might have been a problem somewhere else

pg9182: > When using Visual studio as the generator CMAKE_BINARY_DIR defaults to out/build/CMAKE_BUILD_TYPE/. As I said i chose to output into game so that one can copy their game over and launch from vs directly.

You should be able to set `buildRoot` in `CMakeSettings.json` to control that without breaking CMake conventions and the ability to do out-of-tree builds.

F1F7Y:
> Make sure to also update [`BUILD.md`](https://github.com/R2Northstar/NorthstarLauncher/blob/main/BUILD.md) to explain how to import the project and build it ^^

Done

loader_wsock32_proxy/CMakeLists.txt
F1F7Y: > You should be able to set `buildRoot` in `CMakeSettings.json` to control that without breaking CMake conventions and the ability to do out-of-tree builds.

Due to cmake giving developers the ability to build in way more environments i chose to hard code the build path and ignore `CMakeSettings.json`

pg9182: What?

pg9182 (commented):
Changes look okay overall; there's some inconsistencies in the import libs (casing, extra libs), but that's not a big deal rn and it can be fixed in a later PR.

I'm still not convinced about the build output dir override, though.

loader_wsock32_proxy/CMakeLists.txt
F1F7Y: > What?

Only the Visual Studio generator respects `CMakeSettings.json`. So we need to hard code the build dir into `CMakeLists.txt` to have cmake behave the same even when not being ran from visual studio.

F1F7Y:
> I'm still not convinced about the build output dir override, though.

I'd much rather move my game into the `game` dir and be able to launch from vs, then have to copy over the binaries and manually attach the debugger every time i make a small change

loader_wsock32_proxy/CMakeLists.txt
pg9182: You pass the cmake build dir when setting up cmake.

pg9182: I'm fine with keeping NS_BINARY_DIR (it's good to have for the reason you mentioned), but I'd like it to respect CMake conventions by default.

pg9182: @F1F7Y

F1F7Y: Unsure what you mean by "respect CMake conventions". The reason I made `NS_BINARY_DIR` and not just used `CMAKE_BINARY_DIR` is because that is also used for the cache. That means setting `CMAKE_BINARY_DIR` to `${CMAKE_SOURCE_DIR}/game` will have cmake generate its files inside the game directory cluttering up your titanfall install

uniboi:
When I run `cmake .` in the repository root cmake generates 44 untracked files

F1F7Y:
> When I run `cmake .` in the repository root cmake generates 44 untracked files

Did you git pull, because i added all files to gitignore

pg9182:
> When I run cmake . in the repository root

With cmake, you generally make a build folder, then run `cmake ..` from in it.

catornot (approved):
Everything worked. 
Didn't have any crashes.

CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+cmake_minimum_required(VERSION 3.23)
```
uniboi, line 1:
```suggestion
cmake_minimum_required(VERSION 3.12)
```
I'm too lazy to update cmake and it works the same in 3.12

.gitignore
```diff
@@ -16,6 +16,25 @@
 # Mono auto generated files
 mono_crash.*
 
+# CMake output
+out/
+game/
+CMakeFiles/
+cmake_install.cmake
+CMakeCache.txt
+CopyOfCMakeCache.txt
```
uniboi, line 25:
```suggestion
CopyOfCMakeCache.txt
ALL_BUILD.vcxproj
ALL_BUILD.vcxproj.filters
Northstar.sln
ZERO_CHECK.vcxproj
ZERO_CHECK.vcxproj.filters
loader_wsock32_proxy/loader_wsock32_proxy.vcxproj
loader_wsock32_proxy/loader_wsock32_proxy.vcxproj.filters
NorthstarDLL/NorthstarDLL.vcxproj
NorthstarDLL/NorthstarDLL.vcxproj.filters
```
those are all of the untracked files generated for me

uniboi (commented):
> Did you git pull, because i added all files to gitignore

yes now it's only 10 untracked files

> With cmake, you generally make a build folder, then run cmake .. from in it.

building in the root or an untracked folder has no effect here, the same untracked files get generated either way

GeckoEidechse:
Just realised that we maybe should've poked @BobTheBob9 on this as well ^^"

.gitignore
GeckoEidechse: Resolved in 7e0bc06dd62f5e9ee0075b90c00d4f44e1f1cfa3?

F1F7Y: yes

GeckoEidechse (approved):
Confirmed working in testing for:
- [x] Compiling with Visual Studio
- [x] Compiling via Visual Studio commandline
- [x] CI artifact

in all cases I replaced existing `NorthstarLauncher.exe` and `Northstar.dll` and then ran Northstar and joined some random server.

GeckoEidechse:
I just noticed that we [still build _NorthstarLauncher_ from scratch in release repo](https://github.com/R2Northstar/Northstar/blob/main/.github/workflows/build.yml#L38-L41) as well which will break with this PR.

The fix for this would be to finally update release repo to just download release files from this repo instead of building from scratch, finally decoupling the two repos.

So essentially, while merging this would break release repo, it shouldn't block the PR as the fix can be implemented independently ^^

pg9182:
> building in the root or an untracked folder has no effect here, the same untracked files get generated either way

The idea is you ignore the build folder as a whole.

loader_wsock32_proxy/CMakeLists.txt
pg9182: When you do the following, you should end up with the binaries somewhere in the `./build` folder.

```
mkdir build
cd build
cmake /path/to/NorthstarLauncher
ninja
```

With `NS_BINARY_DIR` set to `${CMAKE_SOURCE_DIR}/game` as the default, the binaries end up in `/path/to/NorthstarLauncher/game` rather than `./build/game` like they would with `${CMAKE_BINARY_DIR}/game`.

pg9182 (changes requested):
- NS_BINARY_DIR changes
- @uniboi's suggestion to lower the minimum cmake version
- also add `build/` to gitignore

F1F7Y:
Just as a note i have not tested the last 4 commits and wont be able to do so till Thursday

GeckoEidechse (changes requested):
Building e75775a55b2ea0984b4dcaa5dbf8175b4f520233 from a fresh clone using the described instructions (open folder in visual studio -> build all) results in the binaries to land in `%userprofile%\CMakeBuilds\467109644\build\x64-Debug` which is kinda annoying IMO.

Produced binaries and files should stay inside the repo folder. I'm assuming some path prefix variable isn't set by default in the CMake config?

**EDIT:** CI artifact works just fine in testing btw ^^

F1F7Y:
Fixed the issues. The binaries still build into `${CMAKE_BINARY_DIR}/game`. In visual studio `CMAKE_BINARY_DIR` is set to `${projectDir}/out` ( check comment in root `CMakeLists.txt` )

Version `3.15` is also the minimum version without getting any errors

uniboi:
What error do you get with sub 3.15?

F1F7Y:
> What error do you get with sub 3.15?

libcurl linking errors iirc

uniboi (approved):
I've been working with branch for quite some time and never had complaints

BobTheBob9:
seems quite nice at a glance, is the game directory configurable? i store my northstar dev env in Titanfall2/R2Northstar/src, would be nice to be able to keep that

F1F7Y:
> seems quite nice at a glance, is the game directory configurable? i store my northstar dev env in Titanfall2/R2Northstar/src, would be nice to be able to keep that

The game directory is hardcoded to `${CMAKE_BINARY_DIR}/game` where `CMAKE_BINARY_DIR` can be any path set by the user.

So for you to not have to move the game you'd have to point it to the root directory of `Titanfall2` and rename `Titanfall2` to `game`. Then update the registry so origin can still find the game

F1F7Y:
On that note it might make more sense to use `Titanfall2` instead of `game` as the directory we build into

pg9182:
> The game directory is hardcoded to ${CMAKE_BINARY_DIR}/game where CMAKE_BINARY_DIR can be any path set by the user.

Didn't you change it to be indirect through `NS_BINARY_DIR`, so you can use that variable specifically to override just the output folder without putting the rest of the build stuff there?

F1F7Y:
> > The game directory is hardcoded to ${CMAKE_BINARY_DIR}/game where CMAKE_BINARY_DIR can be any path set by the user.
> 
> Didn't you change it to be indirect through `NS_BINARY_DIR`, so you can use that variable specifically to override just the output folder without putting the rest of the build stuff there?

What I got from our previous conversation is that you wanted me to use `CMAKE_BINARY_DIR` so i used it

F1F7Y:
👉 👈

GeckoEidechse:
Hopefully retesting tonight to see if https://github.com/R2Northstar/NorthstarLauncher/pull/467#pullrequestreview-1478177431 is resolved. If it is, there'd be nothing on my side blocking it ^^

GeckoEidechse (approved):
> Building [e75775a](https://github.com/R2Northstar/NorthstarLauncher/commit/e75775a55b2ea0984b4dcaa5dbf8175b4f520233) from a fresh clone using the described instructions (open folder in visual studio -> build all) results in the binaries to land in `%userprofile%\CMakeBuilds\467109644\build\x64-Debug` which is kinda annoying IMO.

Tested with fresh clone of a1ebc2fa6f6f26568e3cc2408f783ca0049d251e and it all stays inside repo folder now. Both when using Visual Studio and when using CLI.

Tested as working when running CI artifact (skipped testing local builds cause time constraints)

CMakeLists.txt
```diff
@@ -10,6 +10,13 @@ set(CMAKE_VS_PLATFORM_TOOLSET v143)
 
 # This determines the real binary root directory
 set(NS_BINARY_DIR ${CMAKE_BINARY_DIR}/game)
+# NOTE [Fifty]: Visual studio deems Build root with the value "${projectDir}"
```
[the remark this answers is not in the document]
F1F7Y: Noone knows how `git blame` works. Also adding a tag to comments is faster than `git blame` :)

Doesn't hurt to do this imo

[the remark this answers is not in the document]
GeckoEidechse: Eh, just use an extension in your IDE of choice to do it for you.

E.g. vscode + gitlens gives you `git blame` on active line

image


That being said, this is not blocking so I'm fine with this either way in this case ^^

GeckoEidechse:
Poking @pg9182 to look at this again cause your _Changes Requested_ seems to be the only thing blocking rn ^^

Jan200101:
An install target and a variable could be added to specify a Titanfall 2 directory to install to.
I'd think that would be way better than having to copy the files over or copy the whole game over.

F1F7Y:
> An install target and a variable could be added to specify a Titanfall 2 directory to install to. I'd think that would be way better than having to copy the files over or copy the whole game over.

That sounds like a great solution, but imo should be a separate PR otherwise we wont get anywhere

pg9182:
> What I got from our previous conversation is that you wanted me to use CMAKE_BINARY_DIR so i used it

Kind of; it was that you should respect `CMAKE_BINARY_DIR` by default, but giving an additional variable like `NS_BINARY_DIR` which can override it if specified explicitly (i.e., have the default value use `CMAKE_BINARY_DIR`) is okay.

CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+project(Northstar CXX ASM_MASM)
+
+# Language specs
+set(CMAKE_CXX_STANDARD 20)
+set(CMAKE_C_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_VS_PLATFORM_TOOLSET v143)
+
+# This determines the real binary root directory
+set(NS_BINARY_DIR ${CMAKE_BINARY_DIR}/game)
```
pg9182, line 12: This is good.

pg9182 (approved):

CMakeLists.txt
Jan200101: wouldnt this need to be a cache entry to be user settable?
e.g.
`set(NS_BINARY_DIR "${CMAKE_BINARY_DIR}/game" CACHE STRING "Northstar binary path")`

PR #471, opened 2023-06-11
Remove `-vanilla` check

F1F7Y:
Removes `-vanilla` arg check in launcher as it disables loading of `Northstar.dll` which completely defeats the purpose of launching northstar.

The check also checks for `run_northstar.txt` which seems to have been used for launching on linux. Might be also worth removing as, again, there's no point in not loading `Northstar.dll`

### DIFF
--- a/NorthstarLauncher/main.cpp
+++ b/NorthstarLauncher/main.cpp
@@ -255,12 +255,6 @@ void PrependPath()
 bool ShouldLoadNorthstar(int argc, char* argv[])
 {
 	bool loadNorthstar = true;
-	for (int i = 0; i < argc; i++)
-		if (!strcmp(argv[i], "-vanilla"))
-			loadNorthstar = false;
-
-	if (!loadNorthstar)
-		return loadNorthstar;
 
 	auto runNorthstarFile = std::ifstream("run_northstar.txt");
 	if (runNorthstarFile)


pg9182:
`-vanilla` was used with the wsock proxy to launch non-northstar titanfall.

F1F7Y:
> `-vanilla` was used with the wsock proxy to launch non-northstar titanfall.

So, do we still want it? I wanted to use `-vanilla` for for a simple way to play northstar on vanilla, could rename this to smth like `-nodll` or `-nonorthstar`

ASpoonPlaysGames:
> `-vanilla` was used with the wsock proxy to launch non-northstar titanfall.

Yeah but if you want to launch normal titanfall you can just open the normal exe, or remove -northstar from steam/ea launch options.

I've also literally never seen anybody use it

GeckoEidechse:
As this does not affect `run_northstar.txt` I'd be fine with this change. The lack of `-northstar` in Steam for Titanfall2 should be enough to imply loading vanilla.

However the change should be reflected in wiki as well. Further when someone does, `-northstar -vanilla` on Steam for Titanfall2, this would now mean that they want to run Northstar but in "vanilla compatibility mode", right?

ASpoonPlaysGames:
> However the change should be reflected in wiki as well. Further when someone does, `-northstar -vanilla` on Steam for Titanfall2, this would now mean that they want to run Northstar but in "vanilla compatibility mode", right?

Currently, -vanilla basically just "cancels out" -northstar (which is kinda dumb imo but whatever), the idea behind this PR is to make it so that `-northstar -vanilla` equals northstar launcher but vanilla servers, or "vanilla compatibility mode", yes

PR #473, opened 2023-06-20
Make clang-format less strict

F1F7Y:
Docs: <https://clang.llvm.org/docs/ClangFormatStyleOptions.html>

Makes clang-format less strict and also changes some things to make the code more readable ( this will very obviously require some parts to be changed manually ).

As I took my own spin on formatting and these changes are very individual based i expect this to take time, unlike other PRs.


uniboi:
ong

F1F7Y:
also thinking about setting
```
PointerAlignment: Right
ReferenceAlignment: Right
```

.clang-format
```diff
@@ -4,32 +4,24 @@ IndentWidth: 4
...
-BinPackArguments: false
-BinPackParameters: false
-ExperimentalAutoDetectBinPacking: false
-IndentCaseLabels: false
-ObjCBinPackProtocolList: Auto
+AlwaysBreakTemplateDeclarations: Yes
+IndentCaseLabels: true
 SpacesInSquareBrackets: false
-SpacesInParentheses: false
+SpacesInParentheses: true
```
pg9182, line 12: Please not this one...

.clang-format
```diff
@@ -4,32 +4,24 @@ IndentWidth: 4
...
 SpaceAfterCStyleCast: false
 SpaceInEmptyParentheses: false
 SpaceAfterLogicalNot: false
 SpaceBeforeAssignmentOperators: true
 SpaceBeforeCpp11BracedList: true
 SpaceBeforeCtorInitializerColon: true
 SpaceBeforeInheritanceColon: true
-SpaceBeforeParens: ControlStatements
+SpaceBeforeParens: Never
```
pg9182, line 20: Why

pg9182 (changes requested):

pg9182:
Can we limit this PR to relaxing rules rather than drastically changing formatting?

NorthstarDLL/core/structs.h
```diff
@@ -18,45 +18,45 @@
...
 
-#define OFFSET_STRUCT(name) union name
-#define STRUCT_SIZE(size) char __size[size];
-#define STRUCT_FIELD_OFFSET(offset, signature)                                                                                             \
-	struct                                                                                                                                 \
-	{                                                                                                                                      \
-		char CONCAT2(pad, __LINE__)[offset];                                                                                               \
-		signature;                                                                                                                         \
-	};
+#define OFFSET_STRUCT( name ) union name
```
pg9182, line 26: What happened to this file?

NorthstarDLL/util/version.cpp
```diff
@@ -46,7 +46,8 @@ void InitialiseVersion()
 	if (ntdll)
 	{
 		// real win32 version info (i.e., ignore manifest)
-		DWORD(WINAPI * RtlGetVersion)(LPOSVERSIONINFOEXW);
+		DWORD(WINAPI * RtlGetVersion)
+		(LPOSVERSIONINFOEXW);
```
pg9182, line 50: Umm

pg9182 (commented):
Looks good in general.

NorthstarDLL/core/structs.h
F1F7Y: It had `//clang-format off` instead of `// clang-format off` so current clang format still formatted it, while i fixed it here, will go through it and try to reduce the diff

NorthstarDLL/util/version.cpp
F1F7Y: Seems to be caused by `ColumnLimit: 0`, removing `ColumnLimit` completely seems to default it to something, but fixes this

pg9182 (approved):
LGTM. There's some other clang-format settings we should change too, but this is good.

Jan200101 (approved):
Looks good but will probably cause conflicts in some PRs

ASpoonPlaysGames:
Before this gets merged someone should merge main into it, and then fix the formatting

Jan200101 (approved):
re-reviewing since changes were made since my last review.

Still looks good, some additional format changes could be made but its best to split any more changes into new PRs

GeckoEidechse:
Quick skim and the changes are an improvement. That being said:

1. I found out today that `clang-format` does tab indentation incorrectly, check my rant here https://github.com/R2Northstar/NorthstarLauncher/pull/550#issuecomment-1720274765
   (spaces superiority)
3. When merging I will first merge only the config change and then the formatting changes in a separate commit so that I can then add that commit to the [`.git-blame-ignore-revs`](https://github.com/R2Northstar/NorthstarLauncher/blob/1979247175d2eb491f3ffb2ae874ac7c5e03c47a/.git-blame-ignore-revs) file so that git blame is still readable
4. I'll probably merge the other PRs that are ready to merge first so that we don't cause a bunch of PRs that are ready to merge to merge conflict.

GeckoEidechse:
For merging this we would first only merge the `.clang-format` related changes and then in a second commit, made by `Northstar <northstar@northstar.tf>` we would apply the formatting changes so that we then can add the formatting commit to `.git-blame-ignore-revs` so that it doesn't show up in git blame ^^

F1F7Y:
Superseded by #772

PR #477, opened 2023-06-22
Rewrite crash handler

F1F7Y:
The goal of this PR is to make the crash handler code more readable and fix some issues.
- [x] Use an external process for the message box to not block the game
- [x] Format registers
- [x] Format Modules
- [x] Format stacktrace
- [x] Format mods
- [x] Create minidump

This probably wont fix the issue of logs not flushing at times, but i plan to tackle that in my logging PR which comes after this and is one of the reasons i made this one.


F1F7Y:
New crash message looks like this:
image

Unless i spot a problem in testing this should be it. The main change you should notice is that you always only get 1 message box, which won't block the game from closing.

GeckoEidechse:
For testing, maybe mention a way to consistently crash Northstar to test this PR. At least the only one from the top of my head that I know of is increasing host timescale to values above 10 or so.

ASpoonPlaysGames:
> At least the only one from the top of my head that I know of is increasing host timescale
Oh you sweet summer child...

a good one is just duplicate audio overrides, install an exrill colour pack (and dont remove the other colours) and you will get like 10 crashes at the same time

F1F7Y:
> a good one is just duplicate audio overrides, install an exrill colour pack (and dont remove the other colours) and you will get like 10 crashes at the same time

This ^

F1F7Y:
Calling this script function:
```cpp
AUTOHOOK_INIT();

FUNCTION_AT(server.dll + 0x996e0, void*, __fastcall, CreateNewRecordedAnimation, (void))

ADD_SQFUNC("void", AnimTest, "", "", ScriptContext::SERVER)
{
	void* anim = CreateNewRecordedAnimation();
	return SQRESULT_NULL;
}
```
causes `GetModuleHandleExA` [here](https://github.com/R2Northstar/NorthstarLauncher/pull/477/files#diff-28f35e0a11bd4ffd6449a5f11f55a32d5da3b495fd8ed84a0229b293813ff99eR146) to fail. Thing is we pass it the exception pointer given to us in `ExceptionFilter` so there's not much we can do.

Old crash handler didn't check the return value of winapi functions and crashed inside of it-self so this is still an improvement.

F1F7Y:
Message thread ( in modding-chat, discord ): [start](https://discord.com/channels/920776187884732556/922663696273125387/1123334529306869770)
~ 30 message conversation

Old crash handler caught the exception, but didn't terminate titanfall 2 leading to multiple MessageBox pop-ups

New crash handler only showed 1 pop-up and properly terminated titanfall 2

NorthstarDLL/logging/crashhandler.cpp
```diff
@@ -1,377 +1,527 @@
...
+		PROCESS_INFORMATION pi;
 
-	void* exceptionAddress = exceptionInfo->ExceptionRecord->ExceptionAddress;
+		ZeroMemory(&si, sizeof(si));
+		si.cb = sizeof(si);
+		ZeroMemory(&pi, sizeof(pi));
 
-	storedException->cause = GetExceptionName(*storedException);
+		std::string svCmdLine =
+			fmt::format("bin\\CrashMsg.exe {} {} {} {}", GetNorthstarPrefix(), GetExceptionString(), m_svCrashedModule, m_svCrashedOffset);
```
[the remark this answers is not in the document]
Jan200101: Also how would distrobution of this be handled?

NorthstarDLL/logging/crashhandler.cpp
```diff
@@ -1,377 +1,527 @@
...
+	if (!GetModuleHandleExA(GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS, pCrashAddress, &hCrashedModule))
 	{
+		m_svCrashedModule = CRASHHANDLER_GETMODULEHANDLE_FAIL;
+		m_svCrashedOffset = "";
 
-		HMODULE backtraceModuleHandle;
-		GetModuleHandleExA(GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS, static_cast<LPCSTR>(framesToCapture[i]), &backtraceModuleHandle);
+		wchar_t wszError[256];
+		// NOTE [Fifty]: MAKELANGID is deprecated
+		FormatMessageW(FORMAT_MESSAGE_FROM_SYSTEM, NULL, GetLastError(), MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT), wszError, 255, NULL);
```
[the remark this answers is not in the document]
F1F7Y: Looking at the docs I think I should do error checking on this. To answer your question I very much do not know how this would act. I'll test when I get the chance though

NorthstarDLL/logging/crashhandler.cpp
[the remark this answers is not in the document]
F1F7Y: > Is there any major benefit to executing a remote process instead of just doing it in Northstar?

MessageBox blocks the thread it is called on, from experience in some cases the game likes to hang for quite a while before terminating.

Separating the message box into a separate process means the game can close independent of the user interacting with the message box.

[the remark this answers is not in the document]
F1F7Y: > Also how would distrobution of this be handled?

Preferably with a second PR I have in the works, which unfortunately depends on this being merged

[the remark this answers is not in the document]
GeckoEidechse: Feel free to draft PR that one already in case you haven't yet ^^

NorthstarDLL/logging/crashhandler.cpp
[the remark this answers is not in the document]
F1F7Y: addressed in 688690d

F1F7Y:
> For testing, maybe mention a way to consistently crash Northstar to test this PR. At least the only one from the top of my head that I know of is increasing host timescale to values above 10 or so.

`crast_test` is probably the easiest, although it gives only 1 stack frame

emma-miler:
> For testing, maybe mention a way to consistently crash Northstar to test this PR. At least the only one from the top of my head that I know of is increasing host timescale to values above 10 or so.

`say test` on main menu :)

F1F7Y:
I think I found a consistent fix for crash messages but showing up in some crashes ( present in both old and new crash handler ).

I'm on holiday so this won't be implemented in the next 5 days. This is also a fix for very rare cases so I'm not marking as draft.

ASpoonPlaysGames (commented):
When testing, I don't get a crash dialog popup when crashing. Is this intentional?

F1F7Y:
> When testing, I don't get a crash dialog popup when crashing. Is this intentional?

What did you do to crash ?

F1F7Y:
> `say test` in main menu

The crash message is now a separate exe located in `bin/`. If this is missing you wont get any message box.

testing with flightcore it seems to properly download the northstar dll and launcher exe, but doesnt extract the crashmsg exe ?
cc @GeckoEidechse ^

GeckoEidechse:
> > `say test` in main menu
> 
> The crash message is now a separate exe located in `bin/`. If this is missing you wont get any message box.
> 
> testing with flightcore it seems to properly download the northstar dll and launcher exe, but doesnt extract the crashmsg exe ? cc @GeckoEidechse ^

Yes that's kinda on purpose cause the zip contains a bunch of debugging files etc and I only grab what is relevant in order to avoid filling up Titanfall2 dir.

Which brings me to a good point. What's the reason for the separate exe? Cause adding new files is kinda a big deal in the sense that we technically can never remove them again cause 1/4 of our playerbase insists to keep using manual installs.

Also, I have no idea how mod-managers outside of FlightCore, in particular r2mm handle new files that aren't `NorthstarLauncher.exe`, `Northstar.dll`, or `R2Northstar`.

CrashMessage/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
...
+
+add_executable(CrashMessage
+               "main.cpp"
+)
+
+set_target_properties(CrashMessage PROPERTIES
+                      RUNTIME_OUTPUT_DIRECTORY ${NS_BINARY_DIR}/bin
+                      OUTPUT_NAME CrashMsg
+                      LINK_FLAGS "/MANIFEST:NO /DEBUG /SUBSYSTEM:WINDOWS /ENTRY:mainCRTStartup"
+)
```
GeckoEidechse, line 11: This adds a new exe which might be an issue, see https://github.com/R2Northstar/NorthstarLauncher/pull/477#issuecomment-1654548775

F1F7Y:
> Which brings me to a good point. What's the reason for the separate exe? Cause adding new files is kinda a big deal in the sense that we technically can never remove them again cause 1/4 of our playerbase insists to keep using manual installs.

Separating the message box from the game process. This means the game no-longer stays open untill the message box is closed.

Imo this is better for user experience as the game can close independently of the user interacting with the message box freeing up both system resources and not being annoying with fullscreen grabbing focus / taking extra time to close in some circumstances

GeckoEidechse:
Ok, that's actually a good reason. Only issue is that we need to make sure that adding a new exe doesn't break existing workflows and doesn't cause issues longterm as we realistically cannot remove it again later. Let's have the discussion in https://github.com/R2Northstar/NorthstarLauncher/pull/477#discussion_r1276806465 as otherwise it'll clutter up the thread to figure this all out.

CrashMessage/CMakeLists.txt
GeckoEidechse: Alright, reasoning for new exe is in: https://github.com/R2Northstar/NorthstarLauncher/pull/477#issuecomment-1654565064

copy/paste from @F1F7Y's comment:

> [Additional exe is for] separating the message box from the game process. This means the game no-longer stays open untill the message box is closed.
> 
> Imo this is better for user experience as the game can close independently of the user interacting with the message box freeing up both system resources and not being annoying with fullscreen grabbing focus / taking extra time to close in some circumstances

GeckoEidechse: Tbh, the freeing up resources part is not that important IMO, as they will be freed up when box is closed anyway, so in my opinion it doesn't make much difference.

GeckoEidechse: Again my main concern is that we basically cannot remove it again from someone's install and we have also been very adamant to keep the user install "clean" for the most part.

uniboi:
We can remove it later? Installers should clean up after themselves anyways and even if not, having a floating exe in the install dir breaks nothing

CrashMessage/CMakeLists.txt
ASpoonPlaysGames: Separate exe is the superior choice, for a number of reasons:
1. system resources
2. no potential for crash handler crashing and causing an infinite crash loop (which has happened before)
3. fullscreen with titanfall 2 is fucky
4. If in the future we want to do log uploading and such, it would be much better to do that from a separate process rather than from the game process

uniboi: We can remove it later? Installers should clean up after themselves anyways and even if not, having a floating exe in the install dir breaks nothing

grrr

GeckoEidechse: > We can remove it later? Installers should clean up after themselves anyways [...]

Installers outside of FlightCore are slow to update (VTOL still doesn't support packages for example) and unfortunately 1/4 of Northstar is still on manual installs x_x

> [...]  and even if not, having a floating exe in the install dir breaks nothing

Also true but not super nice to litter folder with "random" stuff

GeckoEidechse: At the very least I'd say we should make sure that by just looking at the exe it's clear it comes from Northstar, i.e. rename the binary from the current `CrashMsg.exe` to something like `NorthstarCrashMsg.exe`.

Also, is there a specific reason it's placed in `bin/`? Could we also stick it in profile dir like `Northstar.dll` or is there a reason preventing that?

F1F7Y: > At the very least I'd say we should make sure that by just looking at the exe it's clear it comes from Northstar, i.e. rename the binary from the current `CrashMsg.exe` to something like `NorthstarCrashMsg.exe`.

Stubs don't do this and i dont think this is necessary 

> Also, is there a specific reason it's placed in `bin/`? Could we also stick it in profile dir like `Northstar.dll` or is there a reason preventing that?

r5reloaded also has a `crashmsg.exe` in `bin/`. I don't see any reason to put it into profile dir

GeckoEidechse: > > At the very least I'd say we should make sure that by just looking at the exe it's clear it comes from Northstar, i.e. rename the binary from the current `CrashMsg.exe` to something like `NorthstarCrashMsg.exe`.
> 
> Stubs don't do this and i dont think this is necessary

Good point, paging @pg9182 for this (if he responds in time :>)

GeckoEidechse: On that note, I assume there isn't an easy way to keep logic inside currently existing binaries and have the move to separate binary be a separate PR, right?

Cause as far as I can tell, this is the only thing blocking so splitting it up would help fast tracking this PR here ^^"

F1F7Y: > On that note, I assume there isn't an easy way to keep logic inside currently existing binaries and have the move to separate binary be a separate PR, right?

This is really easy and testing locally apart from the game window staying open it seems to behave correctly ( almost as if i wrote this to prevent duplicate messages :> ). If this is really blocking i can push this change rn

ASpoonPlaysGames: > If this is really blocking i can push this change rn

It *really* shouldn't be blocking

GeckoEidechse: > > If this is really blocking i can push this change rn
> 
> It _really_ shouldn't be blocking

This is the first time we are adding a new binary file since stubs and the first time we are adding an additional binary from Launcher code since wsock32 proxy both of which are over a year ago. Usually I'd poke @BobTheBob9 or @pg9182 for this but they'd take forever to respond ._.

Hence it would probably be best to just have integrated for now and make separate PR for separating out. That way we can have the rewrite merged now as opposed to getting potentially stuck in PR hell.

As I said, we don't have an easy way of removing files again (especially outside R2Northstar folder) so I'm always double/triple checking when we do add new ones. While I don't anticipate us having to remove the exe for some reason, I just really wanna make sure that we don't run into this case if we can avoid it ^^

uniboi: We *dont* want to avoid adding a new file, we *want* a new exe. Otherwise you're stuck with a stack of crash messages from audio overrides.
Having an additional exe is really no problem.

Jan200101: I'm really not a fan of adding another binary and think the surface area that the user can interact with should be minimal.

uniboi: Do you honest to god think any user executes a random exe in the `bin/` directory that's called `CrashMsg.exe`

Jan200101: Yes, but that is not my point.

What happens when the CrashMsg app is missing or broken or intercepted or replaced?
It adds another point of failure that we cannot manage.

uniboi: The same thing that happens when someone deletes the NorthstarLauncher.exe, Northstar.dll, wsock proxy or any other binary from the game.
No app randomly deletes or modifies binaries placed in a bin directory so this is really not an issue

F1F7Y: If it's missing the crash log still gets printed, just the message box doesn't appear.

r5reloaded has been running the same setup for god knows how long and they never had any issues

Jan200101: r5reloaded also requires you to have a specific game version and download a 1.5GB zip.

Its not exactly a sane comparison to make.

ASpoonPlaysGames: > It adds another point of failure that we cannot manage.

It does also remove some points of failure though. We have had issues in the past where the crash handler crashed, as well as times where the crash handler didnt catch things properly

Also, the r5r comparison is a good one to make, given that a large chunk of our code is adapted from r5r

GeckoEidechse: > Do you honest to god think any user executes a random exe in the `bin/` directory that's called `CrashMsg.exe`

You several underestimate the skill of some users to fuck things up :trollface: 





> > On that note, I assume there isn't an easy way to keep logic inside currently existing binaries and have the move to separate binary be a separate PR, right?
> 
> This is really easy and testing locally apart from the game window staying open it seems to behave correctly ( almost as if i wrote this to prevent duplicate messages :> ). If this is really blocking i can push this change rn

Honestly the best solution IMO. Split up the rewrite and moving to separate exe into separate PRs and then we can merge what is fine rn already ^^

F1F7Y: > You several underestimate the skill of some users to fuck things up :trollface:

If they install CrashMsg.exe wrong, they installed northstar wrong, not caused by this PR. If an installer messes this up it's not caused by this PR too.

> Honestly the best solution IMO. Split up the rewrite and moving to separate exe into separate PRs and then we can merge what is fine rn already ^^

I suppose this is the best solution for this

F1F7Y: Removed the extra exe, will make a separate PR with it

ASpoonPlaysGames (approved):
Works in testing.
Code also looks good as far as I can tell.

CrashMessage/CMakeLists.txt
Alystrasz: > Removed the extra exe, will make a separate PR with it

Shall this PR be closed then?

Alystrasz:
> a good one is just duplicate audio overrides, install an exrill colour pack (and dont remove the other colours) and you will get like 10 crashes at the same time

I didn't take this literaly, I should have :')

image

CrashMessage/CMakeLists.txt
GeckoEidechse: > Shall this PR be closed then?

No this is the PR without the separate exe, i.e. the one we wanna merge :P

Alystrasz:
So this branch should only display one error window, right?
I still have several modals displayed after installing several audio override mods:

image

ASpoonPlaysGames:
> So this branch should only display one error window, right?

Nah, it just shouldn't loop infinitely and stuff iirc

The multiple crashes here are due to audio override things crashing in threads

Alystrasz (approved):
I confirm this PR prevents error modal windows from popping indefinitely, and instead displays a single one.

#### Mod test sample

* [Za Warudo Phase](https://northstar.thunderstore.io/package/cooldudepugs/Za_Warudo_Phase/)
* [Finnegans Ultimate Viper OS](https://northstar.thunderstore.io/package/Finnegan/Finnegans_Ultimate_Viper_OS/)
* [Exrills Neon Ion Series](https://northstar.thunderstore.io/package/EXRILL/Exrills_Neon_Ion_Series/)

#### Screenshots

##### Before

image

##### After

image

Jan200101 (approved):
Tested scenarios that previously caused it to spit out minidumps on loop.
Thats fixed and now it creates just a single minidump and goes catatonic.

GeckoEidechse:
Merging based on reviews.

Issue #475, opened 2023-06-22
undefined reference `__sq_getentity`


Jan200101:
https://github.com/R2Northstar/NorthstarLauncher/blob/89996405ee6341252836d606d5de02fff83ec3ab/NorthstarDLL/squirrel/squirrel.h#L281-L284

No definition for `__sq_getentity` appears to exist on master.

`__sq_getentity` was introduced by #171
and removed in https://github.com/R2Northstar/NorthstarLauncher/commit/0c729c609cd052f4e720ce885c1d5322b63c9edd

I am unsure if this `__sq_getentity` is the proper passthrough for `getthisentity`, shouldn't it be `__sq_getthisentity`?

GeckoEidechse:
Assigning @BobTheBob9 cause git blame says this was changed in #171 (big refactor). Probably just got moved around but too lazy to check ^^

PR #478, opened 2023-06-24
Code cleaning

Jan200101:
Cleaning up a bunch of code to be compatible with other compilers and systems.

### DIFF
--- a/NorthstarDLL/CMakeLists.txt
+++ b/NorthstarDLL/CMakeLists.txt
@@ -148,7 +148,7 @@ add_library(NorthstarDLL SHARED
 target_link_libraries(NorthstarDLL PRIVATE
                       ${CMAKE_SOURCE_DIR}/include/MinHook.x64.lib
                       ${CMAKE_SOURCE_DIR}/include/libcurl/lib/libcurl_a.lib
-                      Ws2_32.lib
+                      WS2_32.lib
                       Crypt32.lib
                       Cryptui.lib
                       dbghelp.lib
--- a/NorthstarDLL/core/hooks.cpp
+++ b/NorthstarDLL/core/hooks.cpp
@@ -218,7 +218,7 @@ void MakeHook(LPVOID pTarget, LPVOID pDetour, void* ppOriginal, const char* pFun
 		spdlog::error("MH_CreateHook failed for function {}", pStrippedFuncName);
 }
 
-AUTOHOOK_ABSOLUTEADDR(_GetCommandLineA, GetCommandLineA, LPSTR, WINAPI, ())
+AUTOHOOK_ABSOLUTEADDR(_GetCommandLineA, (LPVOID)GetCommandLineA, LPSTR, WINAPI, ())
 {
 	static char* cmdlineModified;
 	static char* cmdlineOrg;
@@ -386,7 +386,7 @@ void CallAllPendingDLLLoadCallbacks()
 }
 
 // clang-format off
-AUTOHOOK_ABSOLUTEADDR(_LoadLibraryExA, LoadLibraryExA,
+AUTOHOOK_ABSOLUTEADDR(_LoadLibraryExA, (LPVOID)LoadLibraryExA,
 HMODULE, WINAPI, (LPCSTR lpLibFileName, HANDLE hFile, DWORD dwFlags))
 // clang-format on
 {
@@ -415,7 +415,7 @@ HMODULE, WINAPI, (LPCSTR lpLibFileName, HANDLE hFile, DWORD dwFlags))
 }
 
 // clang-format off
-AUTOHOOK_ABSOLUTEADDR(_LoadLibraryA, LoadLibraryA,
+AUTOHOOK_ABSOLUTEADDR(_LoadLibraryA, (LPVOID)LoadLibraryA,
 HMODULE, WINAPI, (LPCSTR lpLibFileName))
 // clang-format on
 {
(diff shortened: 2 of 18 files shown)


NorthstarDLL/core/hooks.cpp
```diff
@@ -7,7 +7,7 @@
 #include <fstream>
 #include <sstream>
 #include <filesystem>
-#include <Psapi.h>
+#include <psapi.h>
```
pg9182, line 10: Isn't `Psapi` the correct casing for this one?

loader_wsock32_proxy/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@ target_link_libraries(loader_wsock32_proxy PRIVATE
                       ${CMAKE_CURRENT_SOURCE_DIR}/include/MinHook.x64.lib
                       mswsock.lib
                       ws2_32.lib
-                      Shlwapi.lib
+                      shlwapi.lib
```
pg9182, line 17: Isn't `ShLwApi` the correct casing for this one (note that this differs from the header, `Shlwapi`?

NorthstarLauncher/main.cpp
```diff
@@ -10,7 +10,7 @@
 #pragma comment(lib, "Ws2_32.lib")
 
 #include <winsock2.h>
-#include <WS2tcpip.h>
+#include <ws2tcpip.h>
```
pg9182, line 13: Isn't `WS2tcpip` correct?

NorthstarLauncher/CMakeLists.txt
```diff
@@ -24,7 +24,7 @@ target_link_libraries(NorthstarLauncher PRIVATE
                       uuid.lib
                       odbc32.lib
                       odbccp32.lib
-                      Ws2_32.lib
+                      ws2_32.lib
```
pg9182, line 27: Shouldn't this be `WS2_32`?

NorthstarDLL/CMakeLists.txt
```diff
@@ -148,7 +148,7 @@ add_library(NorthstarDLL SHARED
 target_link_libraries(NorthstarDLL PRIVATE
                       ${CMAKE_SOURCE_DIR}/include/MinHook.x64.lib
                       ${CMAKE_SOURCE_DIR}/include/libcurl/lib/libcurl_a.lib
-                      Ws2_32.lib
+                      ws2_32.lib
```
pg9182, line 151: Shouldn't this be `WS2_32`?

pg9182 (changes requested):
Changes look good and are correct overall.

GeckoEidechse:
Sidenote, to get some of this merged faster we could cherry-pick the good looking changes into new PR and merge that already.

Personally I'm the kinda person that would make a new PR for every commit in this pull request cause that tends to get most of the stuff merged pretty fast but I also know that for some contributors making PRs this is kinda annoying / too much effort... ^^

Jan200101:
> Sidenote, to get some of this merged faster we could cherry-pick the good looking changes into new PR and merge that already.
> 
> Personally I'm the kinda person that would make a new PR for every commit in this pull request cause that tends to get most of the stuff merged pretty fast but I also know that for some contributors making PRs this is kinda annoying / too much effort... ^^

Yeah but these things are so small that I don't know if they really varant a dedicated PR.

F1F7Y:
> but I also know that for some contributors making PRs this is kinda annoying / too much effort... ^^

In some cases it is also counter productive especially when two systems depend on each other

Imo this PR is small enough and if it were to be split id split it into cmake and code changes

NorthstarDLL/core/hooks.cpp
Jan200101: according to Microsoft `psapi.h` appears correct
https://learn.microsoft.com/en-us/windows/win32/api/psapi/

pg9182:
> Yeah but these things are so small that I don't know if they really varant a dedicated PR.

I'm with @Jan200101 on this one. For the compiler-compatibility/correctness PRs, it's better to do it in chunks of changes as they're ready, as they're all way too small for individual PRs, and it's much more annoying to test when it's split up. Also, I think I'll be the main one reviewing these ones anyways.

loader_wsock32_proxy/CMakeLists.txt
Jan200101: this is correct, online information on this has appeared inconcistent.

The correct name as part of the SDK is `ShLwApi.Lib`

NorthstarLauncher/main.cpp
Jan200101: Yep, information available was incorrect, will amend.

NorthstarLauncher/CMakeLists.txt
Jan200101: Yep, information available was incorrect, will use the correct naming.

Jan200101:
should close #475

GeckoEidechse:
> Also, if the PR splits changes into individual working commits like this one, it can be okay to PR them together if they're dependent of each other.

100%. I meant changes within the PR that aren't dependent on each other, e.g. type casts vs fixing invalid includes ^^

NorthstarDLL/core/hooks.cpp
Jan200101: Given that online documentation appears rarely correct I've gone to verfiy this and
image

Since the header can be used on case sensitive system I'll also revert this.

Jan200101:
Took me a while to realize VS switched back to Debug instead of using Release by default (can't we force the build type in cmake?) but I tested this on a fresh Visual Studio install against a fresh Titanfall install and it works without a hitch.

pg9182 (approved):
Looks good, changes are correct, build works.

PR #479, opened 2023-06-24
Implement a basic timer class

F1F7Y:
Implements a basic timer class for measuring time

Example usage:
```cpp
#include "utils/timer.h"

...

CTimer timer;
timer.Start();

...

spdlog::info("Thing took {:d} ms to complete", timer.GetDuration());
```


GeckoEidechse:
Out of curiosity what's reason for adding this? E.g. debugging runtime of certain functions?
(Having that kind of info in the PR description helps with better understanding the pull request and whether to add it ^^)

F1F7Y:
We already track time intervals in some places ( [example](https://github.com/R2Northstar/NorthstarLauncher/blob/main/NorthstarDLL/engine/hoststate.cpp#L82) ). This is just a helper class so that one doesn't have to manually store the start point and then subtract to get the duration.

NorthstarDLL/util/timer.h
```diff
@@ -0,0 +1,32 @@
...
+	void Start(void);
+	long long GetDuration(void) const;
+
+  private:
+	std::chrono::time_point<std::chrono::steady_clock> m_Clock;
+};
+
+//-----------------------------------------------------------------------------
+// Purpose:
+//-----------------------------------------------------------------------------
```
uniboi, line 20: mfw

F1F7Y: image

This is on main, unless you have a better idea and want to format all of this we stick with this :)

uniboi: at least write what is does exactly

PR #480, opened 2023-06-24
Rename `InitilaseNorthstar` to `Northstar_Init`

F1F7Y:
Renames `InitilaseNorthstar` to `Northstar_Init`. 

Personally i prefer `<subsystem>_Init` / `<subsystem>_Shutdown` over what we do now. As `InitilaseNorthstar` is exported i made a separate PR.

Unless one is using #451 this shouldnt impact the end user.


pg9182:
> Unless one is using https://github.com/R2Northstar/NorthstarLauncher/pull/451 this shouldnt impact the end user.

That's actually my main objection against #451. Nevertheless, for compatibility, I don't think we should rename this for now unless we're also making other major changes too. I'm going to close this PR for now, but when we rework the load process to make ns relocatable, we can do this change.

PR #482, opened 2023-06-24
Add more DebugDraw functions

F1F7Y:
Adds more Debug Draw functions used to render debug shapes.
Apart from `RenderUnknown` all of these match the [source sdk versions](https://github.com/ValveSoftware/source-sdk-2013/blob/master/mp/src/public/tier2/renderutils.h#L26-L52).

Also hooks `DrawAllOverlays` where one should do custom drawing from native ( navs, triggers, etc... ).


F1F7Y:
Converted to draft as im getting some crashes i need to fix

ASpoonPlaysGames:
Going to close because hella conflicts + this is mostly (completely?) implemented in the navmesh debug draw PR anyway

PR #483, opened 2023-06-25
Add RelWithDebInfo configuration

F1F7Y:
Some contributors are having trouble with the game freezing when running Debug binaries. RelWithDebInfo fixes this for some.


F1F7Y:
Github actions use Release which wasnt changed at all ( neither was Debug ). So there should be no change in artifacts

Jan200101 (approved):

pg9182:
Honestly, we should consider just making RelWithDebInfo the default and only supported target.

pg9182 (approved):
Looks good, builds correctly.

PR #486, opened 2023-06-26
give variables a value with a compatible type

Jan200101:


NorthstarDLL/logging/logging.cpp
```diff
@@ -134,7 +134,7 @@ void InitialiseConsole()
 	if (!strstr(GetCommandLineA(), "-noansiclr"))
 	{
 		g_bSpdLog_UseAnsiColor = true;
-		DWORD dwMode = NULL;
+		DWORD dwMode = 0;
```
Jan200101, line 137: DWORD is an integer type, NULL is for pointers

NorthstarDLL/server/serverchathooks.h
```diff
@@ -9,7 +9,7 @@ enum class CustomMessageType : char
 };
 
 constexpr unsigned char CUSTOM_MESSAGE_INDEX_BIT = 0b10000000;
-constexpr unsigned char CUSTOM_MESSAGE_INDEX_MASK = ~CUSTOM_MESSAGE_INDEX_BIT;
+constexpr unsigned char CUSTOM_MESSAGE_INDEX_MASK = (unsigned char)~CUSTOM_MESSAGE_INDEX_BIT;
```
Jan200101, line 12: bitwise operations implicitly turn into an integers so put it back into its place.

pg9182 (approved):

PR #490, opened 2023-06-26
replace dependent names with explicit template definition

Jan200101:
some calls to `g_pSquirrel` methods depend on what context is and this may change how `<` is interepreted.

Explicitly labeling it as a template makes it clear to the compiler what it is.

### DIFF
--- a/NorthstarDLL/scripts/server/scriptuserinfo.cpp
+++ b/NorthstarDLL/scripts/server/scriptuserinfo.cpp
@@ -7,7 +7,7 @@ ADD_SQFUNC("string", GetUserInfoKVString_Internal, "entity player, string key, s
 	"Gets the string value of a given player's userinfo convar by name", ScriptContext::SERVER)
 // clang-format on
 {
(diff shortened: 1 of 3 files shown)


NorthstarDLL/scripts/server/scriptuserinfo.cpp
```diff
@@ -7,7 +7,7 @@ ADD_SQFUNC("string", GetUserInfoKVString_Internal, "entity player, string key, s
 	"Gets the string value of a given player's userinfo convar by name", ScriptContext::SERVER)
 // clang-format on
 {
-	const R2::CBasePlayer* pPlayer = g_pSquirrel<ScriptContext::SERVER>->getentity<R2::CBasePlayer>(sqvm, 1);
+	const R2::CBasePlayer* pPlayer = g_pSquirrel<ScriptContext::SERVER>->template getentity<R2::CBasePlayer>(sqvm, 1);
```
Jan200101, line 10: This isn't strictly needed but its better to keep the same everywhere.

pg9182 (approved):

PR #491, opened 2023-06-26
Reorganize third-party dependencies into `thirdparty` directory

### LINKED ISSUE #484: 2 copies of MinHook?
there are currently 2 copies of MinHook vendored in NorthstarLauncher:
- [include/MinHook.x64.lib](https://github.com/R2Northstar/NorthstarLauncher/blob/main/include/MinHook.x64.lib)
- [loader_wsock32_proxy/include/MinHook.x64.lib](https://github.com/R2Northstar/NorthstarLauncher/blob/main/loader_wsock32_proxy/include/MinHook.x64.lib)

Was this an oversight or are these two different versions?

F1F7Y:
Everyone does it, we should too ( also not much of an include folder if it has compiled libs now is it )

Closes #484


pg9182 (changes requested):
While we're at it, can we also remove the unused libraries and reorganize the thirdparty dir to contain a subdir for each individual dependency?

F1F7Y:
> While we're at it, can we also remove the unused libraries and reorganize the thirdparty dir to contain a subdir for each individual dependency?

Wanted to do this in a separate Pr but can do it i this one if you want

F1F7Y:
> While we're at it, can we also remove the unused libraries and reorganize the thirdparty dir to contain a subdir for each individual dependency?

My morning self didnt read this fully. Imo removing unused libraries should be a separate PR from this

pg9182 (approved):
Good enough for this PR. Internal shouldn't be it's own folder, and ideally, we'd have a separate include dir for each dep, but we're going to be removing it anyways and this is an improvement over the old layout.

pg9182:
Ah, fuck, I was a bit trigger-happy. We also need to update the `.gitattributes` file too.

GeckoEidechse:
> Ah, fuck, I was a bit trigger-happy. We also need to update the `.gitattributes` file too.

Made PR for it in #496

PR #492, opened 2023-06-28
Replace field preprocessors with less hacky solutions

Jan200101:
Here be dragons
This behavior is a lot more reliable and less dependent on how constants are represented during the preprocessing step.

### DIFF
--- a/NorthstarDLL/core/structs.h
+++ b/NorthstarDLL/core/structs.h
@@ -35,19 +35,42 @@ OFFSET_STRUCT(Name)
 // Special case for a 0-offset field
 #define STRUCT_FIELD_NOOFFSET(offset, signature) signature;
 
+// Just puts two tokens next to each other, but
+// allows us to force the preprocessor to do another pass
 #define FX(f, x) f x
 
+// Macro used to detect if the given offset is 0 or not
+#define TEST_0 ,
 // MSVC does no preprocessing of integer literals.
 // On other compilers `0x0` gets processed into `0`
-#define TEST_0 ,
 #define TEST_0x0 ,
 
+// Concats the first and third argument and drops everything else
+// Used with preprocessor expansion in later passes to move the third argument to the fourth and change the value
 #define ZERO_P_I(a, b, c, ...) a##c
+
+// We use FX to prepare to use ZERO_P_I.
+// The right block contains 3 arguments:
+// NIF_
+// CONCAT2(TEST_, offset)
+// 1
+//
+// If offset is not 0 (or 0x0) the preprocessor replaces
+// it with nothing and the third argument stays 1
+//
+// If the offset is 0, TEST_0 expands to , and 1 becomes the fourth argument
+//
+// With those arguments we call ZERO_P_I and the first and third arugment get concat.
+// We either end up with:
+// NIF_ (if offset is 0) or
+// NIF_1 (if offset is not 0)
 #define IF_ZERO(m) FX(ZERO_P_I, (NIF_, CONCAT2(TEST_, m), 1))
 
+// These macros are used to branch after we processed if the offset is zero or not
 #define NIF_(t, ...) t
 #define NIF_1(t, ...) __VA_ARGS__
 
+// FIELD(S), generates an anonymous struct when a non 0 offset is given, otherwise just a signature
 #define FIELD(offset, signature) IF_ZERO(offset)(STRUCT_FIELD_NOOFFSET, STRUCT_FIELD_OFFSET)(offset, signature)
 #define FIELDS FIELD
 
--- a/NorthstarDLL/core/structs.h
+++ b/NorthstarDLL/core/structs.h
@@ -35,28 +35,20 @@ OFFSET_STRUCT(Name)
 // Special case for a 0-offset field
 #define STRUCT_FIELD_NOOFFSET(offset, signature) signature;
 
-// Based on: https://stackoverflow.com/questions/11632219/c-preprocessor-macro-specialisation-based-on-an-argument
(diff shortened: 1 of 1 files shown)


Jan200101:
```diff
-#define STRUCT_FIELD_OFFSET(offset, signature) \
-	char CONCAT2(pad, __LINE__)[offset]; \
+#define STRUCT_FIELD_OFFSET(offset, signature)                                                                                             \
+	char CONCAT2(pad, __LINE__)[offset];                                                                                                   \
 	signature;
```

the spacing for the backslashes are insanely stupid.

clang-format has an option for this, can we either disable it or set it to left?
https://clang.llvm.org/docs/ClangFormatStyleOptions.html#alignescapednewlines

NorthstarDLL/core/structs.h
```diff
@@ -25,38 +25,24 @@ OFFSET_STRUCT(Name)
...
-// Based on: https://stackoverflow.com/questions/11632219/c-preprocessor-macro-specialisation-based-on-an-argument
-// Yes, this is hacky, but it works quite well actually
-// This basically makes sure that when the offset is 0x0, no padding field gets generated
-#define OFFSET_0x0 ()
+#define FX(f,x) f x
 
-#define IIF(c) CONCAT2(IIF_, c)
-#define IIF_0(t, ...) __VA_ARGS__
-#define IIF_1(t, ...) t
+#define TEST_0 ,
```
Jan200101, line 37: ignore the build failure, despite my writing I forgot that there also needs to be a `TEST_0x0`, because MSVC does not convert the integer constant `0x0` to `0` before.

NorthstarDLL/core/structs.h
```diff
@@ -25,38 +25,24 @@ OFFSET_STRUCT(Name)
...
-#define DO_PROBE(offset) PROBE_PROXY(OFFSET_##offset) // concatenate prefix with offset
-#define PROBE_PROXY(...) PROBE_PRIMITIVE(__VA_ARGS__) // expand arguments
-#define PROBE_PRIMITIVE(x) PROBE_COMBINE_##x // merge
-#define PROBE_COMBINE_(...) PROBE(~) // if merge successful, expand to probe
+#define ZERO_P_I(a,b,c,...) a##c
+#define IF_ZERO(m) FX(ZERO_P_I, (NIF_, CONCAT2(TEST_, m), 1))
 
-#define CHECK_N(x, n, ...) n
+#define NIF_(t,...) t
+#define NIF_1(t,...) __VA_ARGS__
```
Jan200101, line 43: The core logic of this boils down to the same thing as the original, except fewer passes are required and it does not rely on MSVC specific behavior.

NorthstarDLL/core/structs.h
Jan200101: Tested, with this it builds on MSVC again.
Holding off on making a commit because of the formatting issue.

Jan200101:
I managed to send a DLL to Lagaffe on the Discord for testing and it resulted in a crash.
Unsure why, could also have been caused by a botched VS install on my Laptop.
Currently unable to investigate duo to a hardware migration.

Jan200101:
> resulted in a crash.
> Unsure why,

Used my brain and figured out why.

Offset structs are a union that relies on offset anonymous structs to directly access the value at specific byte offsets.

This is extremely hacky.

I have 2 solutions:
- Remove constructors from members that are used within offsets (Vector3 is the only one)
- Rework offset system (a lot of work)

Jan200101:
Rebased to fix conflicts, removed the union meddling, changes to clang-format are not needed anymore.

Added missing preprocessor for MSVCs lack of integer processing at the preprocessor stage and a comment on why.

If needed I can give a partially complete explanation of the previous and new behavior and why this is better than what was there before.

GeckoEidechse:
> the spacing for the backslashes are insanely stupid.
> 
> clang-format has an option for this, can we either disable it or set it to left? https://clang.llvm.org/docs/ClangFormatStyleOptions.html#alignescapednewlines

Yup, 100% would take an align left in separate PR ^^

EladNLG (commented):
i dont know what this does

hummusbird:
how to open  can of beans

hummusbird:
wait a minute! this isnt the google!

emma-miler (changes requested):
Code in general looks good, but could do with some more comments explaining what it does and why, and maybe how.

emma-miler (dismissed):

ASpoonPlaysGames (approved):
Code looks good, comments have been added as per Emma's request.

ASpoonPlaysGames:
Note: for testing, just launching the game and playing a match or two should suffice? this shouldn't change any actual logic

Jan200101:
The Field macros are only used in a few places but joining a match should suffice.
It might be worth seeing if servers work with this applied, since the code that uses Field is used a lot more on server only code.

F1F7Y:
Related to this PR, on my ns fork ( branch: [`primedev`](https://github.com/F1F7Y/NorthstarLauncher/tree/primedev) ) I got rid of `OFFSET_STRUCT` fully by replacing it with fully / partially reversed equivalents. Everything i have on there fully lines up as far as i know

Jan200101:
I welcome the removal of it, I opted against it after multiple contributors said they find it to be a useful macro.

EM4Volts (commented):
code looks like code (dont know whats goin on, maybe someday ill learn something that isnt python)

other then that tested. copied over files from the artifact, goes ingame. no errors.

GeckoEidechse:
Merging based on reviews.

PR #494, opened 2023-06-30
Remove Vector3 Constructor

Jan200101:
Packing was enabled because the compiler may add padding which we don't want.

Constructors were removed because they were either rather primitive or broken (the `pRawFloats` one).

Use of constructor was replaced with a simple cast, which works just as well.


Jan200101:
After spending a few seconds with the compiler explorer
I came to the conclusion that packing is entirely pointless here, so I'm removing it.

I'll also merge the CI correction commit into the main one and rebase, for the sake of clarity.

Alystrasz (approved):
Confirmed building and running.

uniboi (approved):
Code looks good & fixes an incorrect allocation

GeckoEidechse:
Not testing myself cause others already did. Just trusting they tested well enough :>

PR #497, opened 2023-07-02
CMAKE: set build type if not previously set, turn bin dir message into status

Jan200101:


F1F7Y (approved):
Have not tested in northstar, but the same setup works correctly on my personal projects.

GeckoEidechse (approved):
Built fine in a fresh build using visual studio

PR #414, opened 2023-02-04
Update libcurl and remove openssl dependency

BobTheBob9:
Updates `libcurl` to version `7.87.0` (with websocket support) built with winssl for ssl support, also removes openssl as it is not currently required for any code or as a dependency for any libs we use

PR #498, opened 2023-07-02
replace vendored libcurl and minhook with submodules

Jan200101:
minhook uses the latest commit since CMake support is relatively new.

This was compiled against msvc-wine and tested using the Parkour gamemode and a round of attrition.

Only libcurl and minhook have been turned into submodules, since those are the only ones we need to actually compile.

closes #414

PR #500, opened 2023-07-02
Replace vendored libcurl and minhook with submodules

Jan200101:
related to #498 

Actually includes the changes to CI for both generic builds and releases.

Things that are open to discussion
- [x] How to streamline the process
- [ ] Documentation

Closes #414


Jan200101:
- How to streamline the process
we can invoke git from cmake and init the submodules if they were previously empty
This would simplify the workflow for a lot of people that simply rely on VS to set things 

- Build Documentation
BUILD.md currently advises to download a *zip* of NorthstarLauncher, which does not automatically include any submodules.

There is now simple way to adapt the docs for this.

GeckoEidechse:
>  we can invoke git from cmake and init the submodules if they were previously empty

This would mean though that until first compile and unless the developer manually initiated the submodules, they would get linting errors due to missing files in their IDE.

Not the worst but maybe we can find a way to improve that?

Jan200101:
> Not the worst but maybe we can find a way to improve that?

Not with git submodules, I'm afraid.

Alternatives to submodules are:
- cmake FetchContent (gets dependencies at configure time)
- vendoring the dependency (slightly better than what we do right now, but has the tendency to fill the repo up with garbage that makes it heavier with updates)
- use submodule alternatives like subtree (this would require the developer to install external tools)

GeckoEidechse:
Yeah, I guess submodules are fine as long as docs are updated. I would even put `git clone --recurse-submodules https://github.com/R2Northstar/NorthstarLauncher.git` as a command in the README.md

Just cause one might not immediately expect to have to clone with submodules. That + maybe a link to the submodule page in git docs just so that we can avoid having a bunch of contributors ask why their project doesn't build ^^

(Man I wish C++ had a standard package manager like Rust x_x)

Jan200101:
> Yeah, I guess submodules are fine as long as docs are updated. I would even put `git clone --recurse-submodules https://github.com/R2Northstar/NorthstarLauncher.git` as a command in the README.md

idk, maybe?
I personally like `git submodules update --init --recursive` more (and we don't even need the recursive) just because its a standalone command that someone can simply copy paste out of the build docs.

> Just cause one might not immediately expect to have to clone with submodules.

We get add a function that runs at the start of the relevant cmake files to check if the submodules are setup and if not try to do it for them.
This would alleviate some pain for people with existing setups, those that cloned using tools (e.g. Visual Studio) and lets us give more informative errors.

---

p0 also suggested using their fork of minhook to fix some of the issues with it, any objections to this? (I can't find the fork at the point of writing this so dunno what changed)

F1F7Y:
My question is what does this bring to the table? We don't submodule all the libs ( and we can't do that due to spdlog having a config header ) so what's the point in adding complexity with no real benefits.

Jan200101:
> My question is what does this bring to the table?

.

> We don't submodule all the libs ( and we can't do that due to spdlog having a config header ) 

I actually did put in the effort at the time to also convert spdlog and RapidJSON into submodules, but RapidJSON is completely incompatible due to them relying on the cmake config system more than they need.

Spdlog was no problem, the config header can be shipped separately, or be defined in cmake.

> so what's the point in adding complexity with no real benefits.

I wouldn't say there is no benefit.
- all crypto, internal and openssl headers are dropped
- dependencies are not tied to a specific toolchain
- dependencies are more easily updated (At one point god and bob knew how he build libcurl, now only god knows)
- the git repo no longer has to store binary blobs (which git is not great at in the first place)
and probably more that I am not thinking of

GeckoEidechse:
> I personally like `git submodules update --init --recursive` more (and we don't even need the recursive) just because its a standalone command that someone can simply copy paste out of the build docs.

Both work ig. Whatever you think gets as the last amount of contributors being confused about submodules and complaining about failed compiles I'm happy with. I just took the first command that came to my head ^^

> We get add a function that runs at the start of the relevant cmake files to check if the submodules are setup and if not try to do it for them.

Forgot about that one. Ye 100% do that :+1: 

> My question is what does this bring to the table? We don't submodule all the libs ( and we can't do that due to spdlog having a config header ) so what's the point in adding complexity with no real benefits.

What @Jan200101 said + for me personally I prefer having dependencies be defined by some identifier as this means there's less "unrelated" code in the repo. Additionally, while probably not an issue for Northstar but something to consider still, when someone adds a new dependency as vendored code, they could in theory sneak some malicious code inside. So to probably add vendored code, one would technically have to check it against it source to ensure that none of the files have been modified. Again while I doubt someone would try this form of attack against Northstar, I would not trust any PR that adds vendored code written by someone that hasn't been with the project for a significant while.

pg9182:
> I personally like git submodules update --init --recursive more (and we don't even need the recursive) just because its a standalone command that someone can simply copy paste out of the build docs.

I'm with @GeckoEidechse about using `--recurse-submodules` for the initial clone, but we do need to document the standalone method for people who already have it cloned or are using git clients which don't do submodules for them. I don't have strong feelings about this though.

> My question is what does this bring to the table? We don't submodule all the libs ( and we can't do that due to spdlog having a config header ) so what's the point in adding complexity with no real benefits.

What @Jan200101 and @GeckoEidechse said, but especially:
- Cleaner diffs since dependencies have their own independent history, and only a pointer is stored in the main repo.
- No prebuilt binaries in the repo (we can be sure binaries are built from specific sources, and the build process is inherently documented, so it's also easier to update them).
- Better organization of external deps.
- Easier to switch toolchains.

> when someone adds a new dependency as vendored code, they could in theory sneak some malicious code inside ...  I would not trust any PR that adds vendored code written by someone that hasn't been with the project for a significant while

Definitely. Although, it is possible to sneak commits into a submodule not part of the main repo (do it in a fork, and GH will allow that commit to be fetched from the main repo) too. But submodules over vendoring definitely helps with verification, and is also cleaner in general.

pg9182 (approved):

pg9182 (approved):
Approved, but docs still need to be done in this PR or another.

Jan200101:
a cmake function was added that checks if the relevant submodule dir is empty, NorthstarLauncher is git a repo and if git is available before initializing git submodules

Build instructions were updated to reflect the use of git.

GeckoEidechse:
> * [ ]  How to streamline the process
> 
> * [ ] Documentation

I assume these two points are done now? Once this gets tested both from CI build and from compiling locally on a fresh clone, this should then be good to merge. ^^

Though @F1F7Y did ask me to not merge anything hooking related until they had the chance to investigate a hooking related crash this Friday (i.e. tomorrow). So if this PR is through review before that, I guess just poke them to make sure it's fine to merge ^^

Jan200101:
oh yeah and p0 wanted their Minhook fork used instead. Forgot about that until now. Anything opposing that?

F1F7Y:
> Though @F1F7Y did ask me to not merge anything hooking related until they had the chance to investigate a hooking related crash this Friday (i.e. tomorrow). So if this PR is through review before that, I guess just poke them to make sure it's fine to merge ^^

Comparing the logs I highly doubt the issue is hooking related.

When it comes to this PR I'm still of the opinion this is adding unnecessary complexity. If I understand git submodules correctly the ones in this PR aren't bound to any tag which could introduce issues in the future.

F1F7Y:
> oh yeah and p0 wanted their Minhook fork used instead. Forgot about that until now. Anything opposing that?

Should be a separate PR probably

Jan200101:
> When it comes to this PR I'm still of the opinion this is adding unnecessary complexity

While I agree that its added complexity, what alternative do you propose?

> If I understand git submodules correctly the ones in this PR aren't bound to any tag which could introduce issues in the future.

This is a misconception that I have fallen for myself.
Git submodules are pinned to a specific commit internally and cannot update on their own.
Any changes to that pin need to be commited to the main repo, making updates to it apparent.

libcurl is pinned to 801bd5 [reference](https://github.com/R2Northstar/NorthstarLauncher/pull/500/files#diff-fb1dabd3cbb0b83271779a8a49cd8e39077ee31c4098a73afc05185a487f9dce), which is the tag for `curl-7_81_0` the same version that was previously vendored
minhook is pinned to 0f25a2 [reference](https://github.com/R2Northstar/NorthstarLauncher/pull/500/files#diff-c24785fd6fadfb95fc5e044ea21a4fee53e7746be8168bd35fcc031f4027479e), the latest commit at the time of writing, since minhook doesn't really see lot of development anymore, mostly fixes.

F1F7Y:
> While I agree that its added complexity, what alternative do you propose?

Ship the libraries and build them alongside Northstar.
No extra complexity in git cloning and if we're feeling adventurous we could even make slight modifications to them like r5r does with curl

> This is a misconception that I have fallen for myself.
> Git submodules are pinned to a specific commit internally and cannot update on their own.
> Any changes to that pin need to be commited to the main repo, making updates to it apparent.

Thanks for clearing things up, I'll look into this more as the lack of a tag in .gitmodules confused me

p0358:
> Ship the libraries and build them alongside Northstar.

Adding them to the source control is the thing that actually adds complexity, by quickly and unproportionally inflating the repo size, you need to then clone all the past versions of the libraries when cloning the main repo, and 99% size of it can easily be just the libs.

It's more like a very lazy solution that works out when you just want to get things working and not spend extra time on setting up any build configs, or also on compiling the libs.

It has no advantages other than being very lazy. Especially if you ship .lib, MSVC is sometimes updated in a way older compiled .libs won't work, and you gotta recompile them and recommit them, which is tedious. There was a period of time they'd break such compatibility every few days (!), as libs aren't supposed to be used between different compiler versions unlike DLLs...

> if we're feeling adventurous we could even make slight modifications to them like r5r does with curl

The proper way to do this is to maintain a fork of such library with your changeset, and have that added as a submodule. Then it makes it very easy to keep such fork updated and pull in latest upstream changes too.

p0358:
And generally I approve, but idk if the instructions in README aren't too thorough right now, I didn't see most projects ever mention how to install git and VS2022, and frankly if someone has issues doing that on their own, I'm not even sure if they should be trying to compile stuff.

Anyway in current shape, saying "download git from this link" may make people start asking questions like "if I downloaded git from somewhere else and not there specifically, will I have issues? Is this some special git version they're using that I need to get?". I'm just not sure if this isn't gonna create more confusion amongst some than it aims to solve xd

You do you, but I think that part at least is overdone. At the very least, just make it say "1. Install git", and have that be linked to the download page, that shouldn't confuse anyone

Jan200101:
> Anyway in current shape, saying "download git from this link" may make people start asking questions like "if I downloaded git from somewhere else and not there specifically, will I have issues? Is this some special git version they're using that I need to get?". I'm just not sure if this isn't gonna create more confusion amongst some than it aims to solve xd

The idea was to give a direct link to the resource the user needs, like with VS2022.

Also, doesn't VS have a menu button to clone a git repo? Maybe that should be used instead.

Overall I think this is just a stop-gap solution until the build instructions get a rewrite

p0358:
> The idea was to give a direct link to the resource the user needs, like with VS2022.

Yeah, it can be there tbh if you want, I was mostly complaining about the wording that went with it

> Also, doesn't VS have a menu button to clone a git repo? Maybe that should be used instead.

Yeah it does, not sure how popular using it is

> Overall I think this is just a stop-gap solution until the build instructions get a rewrite

What else is there to be rewritten in it, given it's already CMake and already uses git submodules as of this PR?

Jan200101:
> Yeah, it can be there tbh if you want, I was mostly complaining about the wording that went with it

Yeah, wording things isnt my strong suit, anyone got better something better?

> Yeah it does, not sure how popular using it is

One less extra thing to download, I guess.

> What else is there to be rewritten in it, given it's already CMake and already uses git submodules as of this PR?

- preparation steps should be in their own section
- no mention of switching build types (I believe debug builds are the default and debug only result in an unresponsive program looping to infinity)
- no mention of wsock DLL (not needed on Windows but very needed under Proton)
- (this one is recent) a linux section using msvc-wine
- possible docker section if a dockerfile ever gets upstreamed

GeckoEidechse (approved):
Works in testing. I switch branch, never did an init for submodules just build the project and CMake took care of the rest.

Both building via visual studio and via CLI worked flawlessly (made sure to remove submodule files inbetween tests)

GeckoEidechse:
Given that CMake kinda takes care of everything I'm also fine with merging as is ^^

Issue #138, opened 2022-04-10
Detect Wine/Proton version to warn user if running under unsupported versions


GeckoEidechse:
To help with users running Northstar under unsupported Wine/Proton versions it was suggested to check the version the game is running under to display some information if the version is an unsupported one.

Ultimately we need to work on better Wine/Proton support but this should help as a stop-gap solution.

GeckoEidechse:
Detecting Wine version as suggested in [this Stackoverflow thread](https://stackoverflow.com/questions/7372388/determine-whether-a-program-is-running-under-wine-at-runtime/42333249#42333249)

Compiling the following code on Windows

```C
#include <windows.h>
#include <stdio.h>
int main(void)
{
    static const char *(CDECL *pwine_get_version)(void);
    HMODULE hntdll = GetModuleHandle("ntdll.dll");
    if(!hntdll)
    {
        puts("Not running on NT.");
        return 1;
    }

    pwine_get_version = (void *)GetProcAddress(hntdll, "wine_get_version");
    if(pwine_get_version)
    {
        printf("Running on Wine... %s\n",pwine_get_version());
    }
    else
    {
        puts("did not detect Wine.");
    }
    int my_int;
    scanf("%d", &my_int);

    return 0;
}
```

and then running executable on SteamOS 3.0 with the same version for Proton as we use for Northstar, we get:

```
Running on Wine... 7.3
```

i.e. detected Wine/Proton version is `7.3`.
This is an issue as that means we cannot reliably detect if it's the actual right Proton flavour (i.e. Proton-GE vs Proton vs Wine), just if it's the right version of that flavour.

Jan200101:
There are wine internal functions that can be called that would let us identify where the Proton prefix is and from there access the relevant `version` file, so there would be no problem in figuring out what Proton version we are running.

Question is, where do we log this information? How do we inform the user about this?

And while we are at it, are there any other places that would benefit from wine detection? (There is at least 1 function in engine.dll that tries to identify the Windows version and the `status` command has a hardcoded `"Windows"` string)

Jan200101:
Correct, it makes no sense to check.

Proton 8 works out of the box
Proton 7 works with a DLL override and IGO64.dll disabled
All recent releases of GE-Proton work out of the box
NorthstarProton works out of the box

Only issue that still exists is Nvidia's 535 series of drivers running out of memory during setup causing prefix creation failure, but that is out of our control.

[...]

PR #502, opened 2023-07-05
Add OS, Wine and Proton logging

Jan200101:
Relevant log on Fedora 38 running NorthstarProton 8.1
```
[15:16:48] [NORTHSTAR] [info] NorthstarLauncher version: 0.0.0.1+dev
[15:16:48] [NORTHSTAR] [info] Command line: "Z:\home\sentry\.local\share\Steam\steamapps\common\Titanfall2\Titanfall2.exe"   -northstar --use-gl=osmesa
[15:16:48] [NORTHSTAR] [info] Using profile: R2Northstar
[15:16:48] [NORTHSTAR] [info] Operating System: Linux (Wine)
[15:16:48] [NORTHSTAR] [info] Wine build: wine-8.1.r0.gdde73435 ( TkG Staging Esync Fsync )
[15:16:48] [NORTHSTAR] [info] Proton build: NorthstarProton-8.1-1
```

### DIFF
--- a/NorthstarDLL/logging/logging.cpp
+++ b/NorthstarDLL/logging/logging.cpp
@@ -3,6 +3,7 @@
 #include "core/convar/concommand.h"
 #include "config/profile.h"
 #include "core/tier0.h"
+#include "util/version.h"
 #include "spdlog/sinks/basic_file_sink.h"
 
 #include <iomanip>
@@ -219,4 +220,11 @@ void NS::log::FlushLoggers()
 		logger->flush();
 
 	spdlog::default_logger()->flush();
+}
+
+void StartupLog()
+{
+	spdlog::info("NorthstarLauncher version: {}", version);
+	spdlog::info("Command line: {}", GetCommandLineA());
+	spdlog::info("Using profile: {}", GetNorthstarPrefix());
 }
--- a/NorthstarDLL/logging/logging.cpp
+++ b/NorthstarDLL/logging/logging.cpp
@@ -232,7 +232,7 @@ typedef const char*(CDECL* wine_get_build_id_type)(void);
 wine_get_build_id_type wine_get_build_id;
 
 // Not exported Winapi methods
-typedef NTSTATUS (WINAPI* RtlGetVersion_type)(PRTL_OSVERSIONINFOW);
+typedef NTSTATUS(WINAPI* RtlGetVersion_type)(PRTL_OSVERSIONINFOW);
 RtlGetVersion_type RtlGetVersion;
 
 void StartupLog()
--- a/NorthstarDLL/logging/logging.cpp
+++ b/NorthstarDLL/logging/logging.cpp
@@ -6,6 +6,7 @@
 #include "util/version.h"
 #include "spdlog/sinks/basic_file_sink.h"
 
+#include <cstdlib>
 #include <iomanip>
 #include <sstream>
 
@@ -222,9 +223,61 @@ void NS::log::FlushLoggers()
 	spdlog::default_logger()->flush();
 }
 
+// Wine specific functions
+typedef const char*(CDECL* wine_get_host_version_type)(const char**, const char**);
+wine_get_host_version_type wine_get_host_version;
+
+typedef const char*(CDECL* wine_get_build_id_type)(void);
+wine_get_build_id_type wine_get_build_id;
+
 void StartupLog()
 {
 	spdlog::info("NorthstarLauncher version: {}", version);
 	spdlog::info("Command line: {}", GetCommandLineA());
 	spdlog::info("Using profile: {}", GetNorthstarPrefix());
+
+	HMODULE ntdll = GetModuleHandleA("ntdll.dll");
+	if (!ntdll)
+	{
+		// How did we get here
+		spdlog::info("Operating System: Unknown");
+		return;
+	}
+
+	wine_get_host_version = (wine_get_host_version_type)GetProcAddress(ntdll, "wine_get_host_version");
+	if (wine_get_host_version)
+	{
+		// Load the rest of the functions we need
+		wine_get_build_id = (wine_get_build_id_type)GetProcAddress(ntdll, "wine_get_build_id");
+
+		const char* sysname;
+		wine_get_host_version(&sysname, NULL);
+
+		spdlog::info("Operating System: {} (Wine)", sysname);
+		spdlog::info("Wine build: {}", wine_get_build_id());
+
+		char* compatToolPtr = std::getenv("STEAM_COMPAT_TOOL_PATHS");
+		if (compatToolPtr)
+		{
+			std::string compatToolPath(compatToolPtr);
+
+			spdlog::info("Proton build: {}", compatToolPath.substr(compatToolPath.rfind("/") + 1));
+		}
+	}
+	else
+	{
+		// We are real Windows (hopefully)
+		const char* win_ver = "Unknown";
+
+		OSVERSIONINFO osvi;
+		BOOL bIsWindowsXPorLater;
+
+		ZeroMemory(&osvi, sizeof(OSVERSIONINFO));
+		osvi.dwOSVersionInfoSize = sizeof(OSVERSIONINFO);
+		GetVersionEx(&osvi);
+
+		// a reference table can be found in the OSVERSIONINFOA documentation
+		// https://learn.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-osversioninfoa#remarks
+		spdlog::info("Operating System: Windows (NT{}.{})", osvi.dwMajorVersion, osvi.dwMinorVersion);
+	}
 }
--- a/NorthstarDLL/logging/logging.cpp
+++ b/NorthstarDLL/logging/logging.cpp
@@ -6,6 +6,7 @@
 #include "util/version.h"
 #include "spdlog/sinks/basic_file_sink.h"
 
+#include <winternl.h>
 #include <cstdlib>
 #include <iomanip>
 #include <sstream>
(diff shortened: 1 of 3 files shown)


NorthstarDLL/logging/logging.cpp
```diff
@@ -220,3 +222,94 @@ void NS::log::FlushLoggers()
...
 }
+
+// Wine specific functions
+typedef const char *(CDECL *wine_get_host_version_type)(const char**, const char**);
+wine_get_host_version_type wine_get_host_version;
+
+typedef const char *(CDECL *wine_get_build_id_type)(void);
+wine_get_build_id_type wine_get_build_id;
+
+typedef char *(CDECL *wine_get_unix_file_name_type)(LPCWSTR);
```
Jan200101, line 233: These are leftovers from a previous attempt of this that need to be removed.
In a default Wine environment its possible to directly read files, from the host fs simply by translating the paths over.

Proton runs in a container with a static user `steamuser`, which means it cannot read user files unless specifically exposed by the container setup.

NorthstarDLL/logging/logging.cpp
```diff
@@ -220,3 +222,94 @@ void NS::log::FlushLoggers()
...
+					win_ver = "8.1";
+					break;
+				default:
+					break;
+			}
+		}
+		else if (osvi.dwMajorVersion == 10)
+		{
+			win_ver = "10";
+		}
```
Jan200101, line 311: The relevant table I was consulting did not have an entry for Windows 11.

I assume its just `dwMajorVersion == 11`, but I have no setup to test this (neither do I really want to install Windows 11)

NorthstarDLL/logging/logging.cpp
```diff
@@ -220,3 +222,94 @@ void NS::log::FlushLoggers()
...
+	spdlog::info("Command line: {}", GetCommandLineA());
+	spdlog::info("Using profile: {}", GetNorthstarPrefix());
+
+	HMODULE ntdll = GetModuleHandleA("ntdll.dll");
+	if(!ntdll)
+	{
+		// How did we get here
+		spdlog::info("Operating System: Unknown");
+		return;
+	}
```
Jan200101, line 251: This is practically pointless, but its better to sanity check.

NorthstarDLL/logging/logging.cpp
```diff
@@ -220,3 +222,94 @@ void NS::log::FlushLoggers()
...
+    	wine_get_unix_file_name = (wine_get_unix_file_name_type)GetProcAddress(GetModuleHandleA("KERNEL32"), "wine_get_unix_file_name");
+    	wine_get_dos_file_name = (wine_get_dos_file_name_type)GetProcAddress(GetModuleHandleA("KERNEL32"), "wine_get_dos_file_name");
+
+		const char* sysname;
+		wine_get_host_version(&sysname, NULL);
+
+		spdlog::info("Operating System: {} (Wine)", sysname);
+		spdlog::info("Wine build: {}", wine_get_build_id());
+
+		char* compatToolPtr = std::getenv("STEAM_COMPAT_TOOL_PATHS");
```
Jan200101, line 267: `GetEnvironmentVariable` would be more winapi oriented, but getenv doesn't require us to manually deal with stuff, plus its standard.

Jan200101:
relevant issue #138

NorthstarDLL/logging/logging.cpp
Jan200101, line 267: Also regarding this env var: its the absolute host path to the compat tool folder e.g. `/home/user/.local/share/Steam/compatibilitytools.d/NorthstarProton8.1-1`

NorthstarDLL/logging/logging.cpp
```diff
@@ -220,3 +222,94 @@ void NS::log::FlushLoggers()
...
+				default:
+					break;
+			}
+		}
+		else if (osvi.dwMajorVersion == 10)
+		{
+			win_ver = "10";
+		}
+
+		spdlog::info("Operating System: Windows {}", win_ver);
```
GeckoEidechse, line 313: I would maybe even just stick `osvi.dwMajorVersion` and `osvi.dwMinorVersion` into the log IMO. The <Win10 percentage is pretty low AFAIK and only shrinking. If we really need to we can still look up what e.g. `NT6.3` corresponds. To me it kinda just feels like blowing up source code size without much payoff

(bike-shedding)

Jan200101: yeah sure, though perhaps we should document a link to resource we can use to reference the version numbers.

Jan200101:
Ignore me casually bumping my head while working Winapi shit out

F1F7Y (changes requested):
Starting with windows 8.1 [`GetVersionEx`](https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getversionexa) behaves differently. This is on Windows 10 Home 21H2: 
image

The only good replacement i found is [`RtlGetVersion`](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-rtlgetversion?redirectedfrom=MSDN)

Jan200101:
> Starting with windows 8.1 [`GetVersionEx`](https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getversionexa) behaves differently.
> This is on Windows 10 Home 21H2: image

God do I love Microsoft.
Apparently it now returns the relevant Version in relations to the programs Manifest, which we don't have.

Sooo.... I know that Source has an internhal helper function to identify the relevant OS Version, surely Respawn didn't gut it and replace it with their own hacky solution, right? :clueless:

Their method (found in engine.dll) does nonsense checks that at the end just return Windows 8.1.
Annoying

> The only good replacement i found is [`RtlGetVersion`](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-rtlgetversion?redirectedfrom=MSDN)

Yep, that seems to be the best alternative.
Since this is a one-time thing, going kernel space should have practically no impact at all.
Also, the one you linked appears to be the relevant function in Driver Space, here is the link to the winapi one [`RtlGetVersion`](https://learn.microsoft.com/en-us/windows/win32/devnotes/rtlgetversion)

Jan200101:
Turns out, RtlGetVersion is pain in the ass to work with.
Implemented it similar to how the wine callbacks are done, as suggested by this SO Answer https://stackoverflow.com/a/48749703/6331425

Needs re-testing

NorthstarDLL/logging/logging.cpp
```diff
@@ -220,3 +223,70 @@ void NS::log::FlushLoggers()
...
+	else
+	{
+		// We are real Windows (hopefully)
+		const char* win_ver = "Unknown";
+
+		RTL_OSVERSIONINFOW osvi;
+		osvi.dwOSVersionInfoSize = sizeof(osvi);
+
+		RtlGetVersion = (RtlGetVersion_type)GetProcAddress(ntdll, "RtlGetVersion");
+		if (RtlGetVersion && RtlGetVersion(&osvi))
```
F1F7Y, line 281: `STATUS_SUCCESS` ( What `RtlGetVersion` returns ) is 0x0 ( [docs](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/596a1078-e883-4972-9bbc-49e60bebca55) ). This means this check will fail and fall back to the else block.

Modifying the check for the return value yielded this:
image

F1F7Y (changes requested):

NorthstarDLL/logging/logging.cpp
Jan200101: Yep, I knew this but ended up fumbling it up along the way.
It appears to work as expected, so I guess thats solved.

F1F7Y (approved):
Code looks good, will have to go through it in my PR ( #504 ) once merged though.

Tested on Windows 10, works correctly.

GeckoEidechse (approved):
Works fine on Windows 10

```logs
[21:46:20] [NORTHSTAR] [info] NorthstarLauncher version: 0.0.0.1+dev
[21:46:20] [NORTHSTAR] [info] Command line: "C:\Program Files (x86)\Steam\steamapps\common\Titanfall2/NorthstarLauncher.exe" 
[21:46:20] [NORTHSTAR] [info] Using profile: R2Northstar
[21:46:20] [NORTHSTAR] [info] Operating System: Windows (NT10.0)
[21:46:21] [NORTHSTAR] [info] Enabling hook _GetCommandLineA
[21:46:21] [NORTHSTAR] [info] Enabling hook _LoadLibraryExA
...
````

**Comment by BimNoah** [comment]
Works fine on Gentoo with NorthstarProton 8.1.1

```
[21:57:37] [NORTHSTAR] [info] NorthstarLauncher version: 0.0.0.1+dev
[21:57:37] [NORTHSTAR] [info] Command line: "Z:\home\noah\.local\share\Steam\steamapps\common\Titanfall2\Titanfall2.exe"   -noborder -windowed -northstar --use-gl=osmesa
[21:57:37] [NORTHSTAR] [info] Using profile: R2Northstar
[21:57:37] [NORTHSTAR] [info] Operating System: Linux (Wine)
[21:57:37] [NORTHSTAR] [info] Wine build: wine-8.1.r0.gdde73435 ( TkG Staging Esync Fsync )
[21:57:37] [NORTHSTAR] [info] Proton build: NorthstarProton-8.1-1
[21:57:37] [NORTHSTAR] [info] Enabling hook _GetCommandLineA
```

NorthstarDLL/logging/logging.cpp
GeckoEidechse: I guess this got removed?

PR #503, opened 2023-07-05
Add support for loading Thunderstore mods natively

GeckoEidechse:
This PR allows for loading Thunderstore mods directly from a separate directory (called `thunderstore-mods` rn, I'm happy to change).

It expects a folder that represents the unpacked zip from the mod from Thunderstore. This heavily simplifies mod management as now there's a direct link between Thunderstore mod and Northstar (Thunderstore folder name).

image

Loading user installed Northstar mods from `R2Northstar/mods/` is still supported for now as mod-managers need to be updated first and as a fallback for mods not on Thunderstore.

Mod-managers could then also take care of moving existing Thunderstore mods from `mods/` to `thunderstore-mods/`

I will take care of notifying mod-manager authors once a decision about merging (or not merging) this PR has been made.

This PR is an initial implementation and could be adapted based on feedback. As a reminder I barely know how to read C++ and simply took existing code as an inspiration. If something's done in a peculiar way, it's most likely not cause I was trying to be "clever" but simply used what worked so please point out anything odd you might see in the code.

Related: https://github.com/R2Northstar/Northstar/issues/472

### DIFF
--- a/NorthstarDLL/mods/modmanager.cpp
+++ b/NorthstarDLL/mods/modmanager.cpp
@@ -413,9 +413,16 @@ void ModManager::LoadMods()
 
 	// Special case for Thunderstore mods dir
 	std::filesystem::directory_iterator thunderstoreModsDir = fs::directory_iterator(GetThunderstoreModFolderPath());
+	// Set up regex for `AUTHOR-MOD-VERSION` pattern
+	std::regex pattern(R"(.*\\([a-zA-Z0-9_]+)-([a-zA-Z0-9_]+)-(\d+\.\d+\.\d+))");
 	for (fs::directory_entry dir : thunderstoreModsDir)
 	{
 		fs::path modsDir = dir.path() / "mods"; // Check for mods folder in the Thunderstore mod
+		// Use regex to match `AUTHOR-MOD-VERSION` pattern
+		if (!std::regex_match(dir.path().string(), pattern)) {
+			spdlog::warn("The following directory did not match 'AUTHOR-MOD-VERSION' {}", modsDir.string());
+			continue; // skip loading mod that doesn't match
+		}
 		if (fs::exists(modsDir) && fs::is_directory(modsDir))
 		{
 			for (fs::directory_entry subDir : fs::directory_iterator(modsDir))
--- a/NorthstarDLL/mods/modmanager.cpp
+++ b/NorthstarDLL/mods/modmanager.cpp
@@ -419,7 +419,8 @@ void ModManager::LoadMods()
 	{
 		fs::path modsDir = dir.path() / "mods"; // Check for mods folder in the Thunderstore mod
 		// Use regex to match `AUTHOR-MOD-VERSION` pattern
-		if (!std::regex_match(dir.path().string(), pattern)) {
+		if (!std::regex_match(dir.path().string(), pattern))
+		{
 			spdlog::warn("The following directory did not match 'AUTHOR-MOD-VERSION': {}", modsDir.string());
 			continue; // skip loading mod that doesn't match
 		}
--- a/NorthstarDLL/mods/modmanager.cpp
+++ b/NorthstarDLL/mods/modmanager.cpp
@@ -413,7 +413,7 @@ void ModManager::LoadMods()
 				modDirs.push_back(dir.path());
 
 	// Special case for Thunderstore mods dir
-	// for (fs::directory_entry dir : thunderstoreModsDir) // TODO: Use this 
+	// for (fs::directory_entry dir : thunderstoreModsDir) // TODO: Use this
 	for (fs::directory_entry dir : fs::recursive_directory_iterator(GetThunderstoreModFolderPath())) // TODO: remove me, temp
 		if (fs::exists(dir.path() / "mod.json"))
 			modDirs.push_back(dir.path());
--- a/NorthstarDLL/mods/modmanager.cpp
+++ b/NorthstarDLL/mods/modmanager.cpp
@@ -413,10 +413,16 @@ void ModManager::LoadMods()
 				modDirs.push_back(dir.path());
 
 	// Special case for Thunderstore mods dir
-	// for (fs::directory_entry dir : thunderstoreModsDir) // TODO: Use this
-	for (fs::directory_entry dir : fs::recursive_directory_iterator(GetThunderstoreModFolderPath())) // TODO: remove me, temp
-		if (fs::exists(dir.path() / "mod.json"))
-			modDirs.push_back(dir.path());
+	for (fs::directory_entry dir : thunderstoreModsDir) {
+		fs::path modsDir = dir.path() / "mods"; // Check for mods folder in the Thunderstore mod
+		if (fs::exists(modsDir) && fs::is_directory(modsDir)) {
+			for (fs::directory_entry subDir : fs::directory_iterator(modsDir)) {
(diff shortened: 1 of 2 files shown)


NorthstarDLL/mods/modmanager.h
```diff
@@ -9,6 +9,7 @@
 #include <filesystem>
 
 const std::string MOD_FOLDER_SUFFIX = "/mods";
+const std::string THUNDERSTORE_MOD_FOLDER_SUFFIX = "/thunderstore-mods";
```
GeckoEidechse, line 12: Actually maybe we should call it `mods-thunderstore` so it's closer to `mods` in the file manager when sorting alphabetically ^^

uniboi: I'd prefer something like `packages` to stay more in line with the current folder naming

NorthstarDLL/mods/modmanager.cpp
```diff
@@ -410,6 +411,23 @@ void ModManager::LoadMods()
 			if (fs::exists(dir.path() / "mod.json"))
 				modDirs.push_back(dir.path());
 
+	// Special case for Thunderstore mods dir
+	std::filesystem::directory_iterator thunderstoreModsDir = fs::directory_iterator(GetThunderstoreModFolderPath());
+	for (fs::directory_entry dir : thunderstoreModsDir)
+	{
+		fs::path modsDir = dir.path() / "mods"; // Check for mods folder in the Thunderstore mod
```
GeckoEidechse, line 418: Additionally we could add a check here to enforce that the folder matches the `AUTHOR-MOD-VERSION` pattern (e.g. `GeckoEidechse-Titanfall_Slander-0.6.9`)

NorthstarDLL/mods/modmanager.h
ASpoonPlaysGames: > I'd prefer something like `packages` to stay more in line with the current folder naming

I agree, it's probably best to keep thunderstore related things aligned with thunderstore's naming conventions and definitions. There is no such thing as a thunderstore mod in reality, the term came about simply to distinguish between mods that come from the discord server and mods that come from thunderstore.

NorthstarDLL/mods/modmanager.cpp
ASpoonPlaysGames: This is a good idea. Make sure that if this is implemented, we don't just silently skip loading the package, but we log a warning. At some point I'd like to be able to keep track of all modloading warnings and display them in-game.

NorthstarDLL/mods/modmanager.h
GeckoEidechse: Should we do something like `thunderstore-packages` then or just `packages`? I have the feeling that `packages` on it's own might not be associated with modding for most users. The name _Thunderstore_ on the other side might be more associated with modding for someone who used Northstar previously. ^^

ASpoonPlaysGames: Why does it need to be associated with modding? The user shouldn't be interacting with this folder anyway, except for manual installs who can figure it out by themselves and not be supported much further.

GeckoEidechse: fair point ^^

GeckoEidechse: Done in 83f93c25c09204d746a1c89693c6a94082fa761c

NorthstarDLL/mods/modmanager.cpp
GeckoEidechse: Done in 7a6da3754753dab57ba5401c48f2d91484b93741. Mods that do not match the pattern get skipped and a warning is logged.

Example:
image
gives
```
[01:09:29] [NORTHSTAR] [warning] The following directory did not match 'AUTHOR-MOD-VERSION' R2Northstar/packages\S2-Mods-GlitchOverhaul-1.2.11\mods
[01:09:29] [NORTHSTAR] [warning] The following directory did not match 'AUTHOR-MOD-VERSION' R2Northstar/packages\S2ModsGlitchOverhaul-1.2.11\mods
```

GeckoEidechse:
I put you under review request specifically @F1F7Y as you had some ideas for a v2 rewrite of mods (which I'm all in favour). I see this PR as a good way to prepare for a v2 and as such wanna make sure that it's set up in a way that helps us to get to a v2 ^^

F1F7Y:
> I put you under review request specifically @F1F7Y as you had some ideas for a v2 rewrite of mods (which I'm all in favour). I see this PR as a good way to prepare for a v2 and as such wanna make sure that it's set up in a way that helps us to get to a v2 ^^
> 
> So feedback would be super appreciated when you get the chance to <3

Going through it it looks good, will review tomorrow when i get home.

Ignoring core mods would this fully replace the `mods/` folder or would we have two folders for for manual install and thundersotre packages?

Also considering 1 thunderstore package can have multiple mods it might be a good idea to display the thunderstore package name to the user in some way. ( Id keep this for a future PR, just worth mentioning here imo )

GeckoEidechse:
> Going through it it looks good, will review tomorrow when i get home.

Awesome, super appreciated <3

> Ignoring core mods would this fully replace the `mods/` folder or would we have two folders for for manual install and Thunderstore packages?

That's what we gotta figure out next. Right now all documentation for manually install explains how to extract the folder into the `mods/` folder. If we wanna change it to `packages/` we would first have to update the documentation, then add some deprecation notices when loading non-core mods, then stop loading mods from `mods/` dir.

However that doesn't handle the case of what to do with non-Thunderstore mods, e.g. a modder testing a current build of their mod that isn't published to Thunderstore yet. Maybe have another directory to differentiate between core, Thunderstore, and manual (non-thunderstore)? At that point things are gonna start getting chaotic though, especially when you also factor in auto-downloaded mods.

So yeah, don't really have a good solution for that yet. My suggestion would be to keep manually installed non-Thunderstore mods in `mods/` and for manually installing Thunderstore mods, update the documentation so that user can install it in a way that it passes the regex and in turn can then also be picked up by mod-managers ^^

> Also considering 1 thunderstore package can have multiple mods it might be a good idea to display the thunderstore package name to the user in some way. ( Id keep this for a future PR, just worth mentioning here imo )

100% agreed. Thunderstore packages contain a `manifest.json` So we could use that and/or the folder name to see which Thunderstore mod a Northstar mod is part of and then display that within the Northstar mod menu.

F1F7Y (approved):
Works in testing

Code looks okay, but the whone `ModManager` class should probably be refactored for auto-download

GeckoEidechse:
> Code looks okay, but the whole `ModManager` class should probably be refactored for auto-download

Would we do that in #362 or in a separate PR before/after? ^^

GeckoEidechse:
Created #505 to discuss timeline of implementation with mod-manager authors assuming that it's clear that apart from technical aspects this PR will be merged.

F1F7Y:
> Would we do that in #362 or in a separate PR before/after? ^^

Unsure, we'll see how launcher PRs progress as I'd like to implement bobs changes, but making multiple large PRs is just asking for trouble.

Regarding this PR i'm fine with this getting merged.

NorthstarDLL/mods/modmanager.h
GeckoEidechse: We should've stuck with `thunderstore-packages`...

https://github.com/R2Northstar/Northstar/issues/537

Issue #485, opened 2023-06-26
Use of templated function within header without available implementation


Jan200101:
`squirrel.h` makes use of `NS::log::squirrel_logger`
https://github.com/R2Northstar/NorthstarLauncher/blob/23594f24d3a2b09f5d8f912a773540ff3c1db841/NorthstarDLL/squirrel/squirrel.h#L351

but this method is only implemented in `squirrel.cpp`
https://github.com/R2Northstar/NorthstarLauncher/blob/23594f24d3a2b09f5d8f912a773540ff3c1db841/NorthstarDLL/squirrel/squirrel.cpp#L33-L41

Thus the header cannot reasonable make use of the method since it lacks an implementation for the type it references.

The solution would be to move the function into the header, but I have had problems with it refusing to accept the extern `NS::log::SCRIPT_*` usage.

Jan200101:
The calls to `squirrel_logger` in the header were added in #355, so I guess @emma-miler is the one I should ask about this

PR #504, opened 2023-07-07
Logging refactor

F1F7Y:
Refactor to logging to make it more readable ( eg currently you can use both `spdlog::info` and `NS::log::NS->info`, this is bad )

Rough list of things:
- [ ] Refactor logging functions
- [ ] More logging categories
- [ ] Separate log files into a folder
  - Current idea is: ( Very much up to discussion )
    ```%profile%/
    └── %Y-%m-%d_%H-%M-%S/
      ├── native.txt
      ├── warning.txt
      ├── error.txt
      ├── script.txt
      ├── script_warning.txt
      ├── crash.txt
      └── minidump.dmp```
- [ ] Print system and northstar information
- [ ] Ability to toggle ( most likely through a launch arg ) what is logged into gameconsole
- [ ] `-wconsole` arg to disable windows terminal
- [ ] Fix crashlogs not flushing to file

Depends on ( and should preferably be shipped with ): #477


F1F7Y:
> I assume whether it's stored in one file or multiple can be easily adjusted by just modifying the code related to the logger backend? In which case switching between the two should be rather simple once we decide which mode to choose (I'd still argue for single file :P)

spdlog makes it really easy to switch between the configurations, so for the sake of testing i could implement it so that we know which one has more pros

NorthstarDLL/CMakeLists.txt
```diff
@@ -61,6 +61,8 @@ add_library(NorthstarDLL SHARED
             "engine/runframe.cpp"
             "logging/crashhandler.cpp"
             "logging/crashhandler.h"
+			"logging/dbg.cpp"
+			"logging/dbg.h"
```
pg9182, line 65: Whitespace.

pg9182 (changes requested):

NorthstarDLL/CMakeLists.txt
F1F7Y: Fixed in 65497ec

F1F7Y:
I'd say this is pretty much done, just waiting on #477

NorthstarDLL/squirrel/squirrel.h
```diff
@@ -353,15 +348,20 @@ template <ScriptContext context> class SquirrelManager : public virtual Squirrel
...
+				__FUNCTION__,
+				GetContextName(context));
 		}
 
 		SQObject functionobj {};
 		int result = sq_getfunction(m_pSQVM->sqvm, funcname, &functionobj, 0);
 		if (result != 0) // This func returns 0 on success for some reason
 		{
-			NS::log::squirrel_logger<context>()->error("Call was unable to find function with name '{}'. Is it global?", funcname);
+			Error(SQ_GetLogContextNative(context), NO_ERROR, "Call was unable to find function with name '%s'. Is it global?\n", funcname);
```
Jan200101, line 364: This fixes https://github.com/R2Northstar/NorthstarLauncher/issues/485

GeckoEidechse:
@F1F7Y am I correct assuming assuming that this is basically the same as #617 ?

GeckoEidechse:
(Also I just noticed that the label for dependent PR was never removed when #477 was merged causing this PR to be stuck in PR hell, ugh. We really need some way to automate this as otherwise it will become impossible to keep track of stuff in the long run)

F1F7Y:
> @F1F7Y am I correct assuming assuming that this is basically the same as #617 ?

Unless i missed some change it should be yeah.

PR #510, opened 2023-07-21
Turn VPKData into a opaque struct

### LINKED ISSUE #495: Bad VPKData definition
VPKData is currently completely broken to hell and back

https://github.com/R2Northstar/NorthstarLauncher/blob/71f0ee98ccc85d41ba7587d122c83011ab1e25c3/NorthstarDLL/core/filesystem/filesystem.h#L24-L32

because its defined as a struct and not using `OFFSET_STRUCT` the size helper adds padding at the front that should not be there.

The definition in TTF2SDK also does not match what can be found in Northstar
https://github.com/Titanfall-Mods/TTF2SDK/blob/ffd7501b6cd71dc652a361a10b8bf0186d482470/TTF2SDK/IFileSystem.h#L16

---

Since Northstar does not make use of VPKData (and VPKFileEntry because of that) I think these could also be removed, since these are 2 of the 4 places where the OFFSET macros are used.

Jan200101:
closes #495 
Turns VPKData into an opaque struct instead of outright removing it, so references to it can be maintained.


Jan200101:
No testing needed since this data type wasn't used to begin with.

ASpoonPlaysGames (approved):
Every reference to VPKData doesn't use these fields, and I believe the fields were wrong anyway.
image

So this has no effect on compiling (and honestly could even help in the future if someone tries to use `VPKData`)

Issue #508, opened 2023-07-21
Remove `run_northstar.txt` functionality 


Jan200101:
Inspired by #507

## Why
the behavior of `run_northstar.txt` is inconcistent with launch flags depending on the method used:
- wsock32 proxy does not support `vanilla` and thus `run_northstar.txt` with `1` cannot be passed by
- NorthstarLauncher does not support `-northstar` and thus `run_northstar.txt` with `0` cannot be passed by

## How
1. Some people may still have the file from outdated linux installs (or other)
   it would be best to do some sort of warning before removing functionality.
2. The only mod manager that made use of it was FlightCore, and that was recently changed to use launch flags directly.

GeckoEidechse:
We might wanna do a scream test for this first where basically we just gate the logic behind an `if(false){...}`. If no one complains in a month or two we drop it.

The main use of it rn is Steam setup with non-NorthstarProton/latest Proton-GE. For example, I'm pretty sure I still use it on my SteamDeck cause I haven't updated Proton install yet.

Similarly we'll need to update docs accordingly. So like I'd propose feature flag for scream test + updated docs, wait a while, remove.

Jan200101:
> The main use of it rn is Steam setup with non-NorthstarProton/latest Proton-GE.

Yeah, its a bummer that its still stuck in PR hell.
I've gone ahead and made a PR that removes it via preprocessors, I'll also go ahead and see if I can find anyone to look at the Proton PR

Jan200101:
> The main use of it rn is Steam setup with non-NorthstarProton/latest Proton-GE. For example, I'm pretty sure I still use it on my SteamDeck cause I haven't updated Proton install yet.

soon no more, the PR has been merged into Proton and it will be included in the next release.

GeckoEidechse:
> soon no more, the PR has been merged into Proton and it will be included in the next release.

Wait upstream Proton to not overwrite the DLL or which one? ^^"
Link? :point_right::point_left:

Jan200101:
the flag thing.
a wine dll override can be added via the launch options

PR #512, opened 2023-07-21
Disable run_northstar.txt using preprocessor

Jan200101:
Related to #508 

Disables `run_northstar.txt` logic via preprocessors.

The main use case for it has mostly been Linux where steam_helper incorrectly handled arguments.

With NorthstarProton and GE-Proton containing a patch to fix the bug this is mostly a non issue these days.

### DIFF
--- a/primedev/primelauncher/main.cpp
+++ b/primedev/primelauncher/main.cpp
@@ -258,6 +258,7 @@ bool ShouldLoadNorthstar(int argc, char* argv[])
 		if (!strcmp(argv[i], "-nonorthstardll"))
 			return false;
 
+#if 0
 	auto runNorthstarFile = std::ifstream("run_northstar.txt");
 	if (runNorthstarFile)
 	{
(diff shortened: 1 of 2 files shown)


ASpoonPlaysGames:
Why disable using preprocessor instead of commenting/deleting the lines?

Jan200101:
Advised as a first step by iirc Gecko to make it easy to reenable if someone comes out of nowhere and gives a good reason to keep this.

ASpoonPlaysGames:
Fair enough

ASpoonPlaysGames (approved):
Code looks good

ASpoonPlaysGames:
Note for testing: probably best to test on Linux?

Alystrasz (approved):
Code looks good and works as intended on Windows: putting a `0` inside the `run_northstar.txt` file does not prevent Northstar from launching anymore.

Issue #507, opened 2023-07-18
Remove `ns_startup_args*.txt` functionality


F1F7Y:
## Why
1. We need to hook `GetCommandLineA` to support these `.txt` files, but before doing so we already read the commandline meaning some args dont work in the .txt.
2. Current way of [`strstr`](https://cplusplus.com/reference/cstring/strstr/) ing `GetCommandLineA` can lead to false positives with arguments that start with the same prefix.

## What
`tier0.dll` exports `CommandLine()` with which we can check the command line. We already have this mapped out and use in some cases so this is just improving code consistency.

## How
1. Do some sort of warning before removing functionality
  Print warning, discord announcement ?
2. Add support in mod-managers
  Add support to directly passing arguments to the process when creating it
3. Remove functionality in a minor verison increase
  Patch versions mainly serve for small fixes and small features, this is a pretty big change in behavior
4. Possibly have some sort of warning that checks if the file exists after this is implemented

F1F7Y:
@GeckoEidechse @BigSpice @0neGal

Jan200101:
I think enough time has passed and we should finally do this.

0neGal:
The only problem I have with this, is it's now more difficult to have a mod manager be able to change the launch options. As, unless it's launching the `.exe`'s directly, it wont really be able to change the launch options very easily. Say if you're launching through Steam.

Unless I'm mistaken, and there's some easy way to do this?

Jan200101:
> Unless I'm mistaken, and there's some easy way to do this?

See how FlightCore does it

0neGal:
> See how FlightCore does it

According to [this](https://r2northstar.gitbook.io/r2northstar-wiki/installing-northstar/northstar-installers), FlightCore simply doesn't support changing launch arguments. Unless thats outdated?

Jan200101:
> According to [this](https://r2northstar.gitbook.io/r2northstar-wiki/installing-northstar/northstar-installers), FlightCore simply doesn't support changing launch arguments. Unless thats outdated?

https://github.com/R2NorthstarTools/FlightCore/blob/5714309211d389c453c19bb77755855de33655fd/src-tauri/src/northstar/mod.rs#L260

0neGal:
Ah, I stand corrected, I wasn't aware you could add launch arguments to `steam://run` URIs

GeckoEidechse:
bump

GeckoEidechse:
[...]

PR #513, opened 2023-07-21
Add support for loading plugins from Thunderstore packages

### LINKED ISSUE #509: Add support for loading plugins from `packages` directory
#503 add support for loading unpacked Thunderstore mods directly from `packages`. However it doesn't support loading plugins yet.

As plugins can be uploaded to Thunderstore, this functionality should be added to NorthstarLauncher.

**EDIT:** An example of a plugin on Thunderstore: https://northstar.thunderstore.io/package/cat_or_not/kyurid/

Jan200101:
closes #509

reposition things a bit, adds a few more sanity checks and looks for dlls in packages.

Tested with the kyurid plugin.

### DIFF
--- a/NorthstarDLL/plugins/plugins.cpp
+++ b/NorthstarDLL/plugins/plugins.cpp
@@ -188,6 +188,23 @@ std::optional<Plugin> PluginManager::LoadPlugin(fs::path path, PluginInitFuncs* 
 	return plugin;
 }
 
+inline void findPlugins(fs::path pluginPath, std::vector<fs::path>& paths)
+{
+	if (fs::exists(pluginPath) && fs::is_directory(pluginPath))
+	{
+		// ensure dirs exist
+		fs::recursive_directory_iterator iterator(pluginPath);
+		if (std::filesystem::begin(iterator) != std::filesystem::end(iterator))
+		{
+			for (auto const& entry : iterator)
+			{
+				if (fs::is_regular_file(entry) && entry.path().extension() == ".dll")
+					paths.emplace_back(entry.path());
+			}
+		}
+	}
+}
+
 bool PluginManager::LoadPlugins()
 {
 	if (strstr(GetCommandLineA(), "-noplugins") != NULL)
@@ -213,37 +230,14 @@ bool PluginManager::LoadPlugins()
 	data.version = ns_version.c_str();
 	data.northstarModule = g_NorthstarModule;
 
-	if (fs::exists(pluginPath) && fs::is_directory(pluginPath))
-	{
-		// ensure dirs exist
-		fs::recursive_directory_iterator iterator(pluginPath);
-		if (std::filesystem::begin(iterator) != std::filesystem::end(iterator))
-		{
-			for (auto const& entry : iterator)
-			{
-				if (fs::is_regular_file(entry) && entry.path().extension() == ".dll")
-					paths.emplace_back(entry.path());
-			}
-		}
-	}
+	findPlugins(pluginPath, paths);
 
 	// Special case for Thunderstore plugin dirs
 
 	for (fs::directory_entry dir : fs::directory_iterator(GetThunderstoreModFolderPath()))
 	{
 		fs::path pluginDir = dir.path() / "plugins";
-		if (fs::exists(pluginDir) && fs::is_directory(pluginDir))
-		{
-			fs::recursive_directory_iterator iterator(pluginDir);
-			if (std::filesystem::begin(iterator) != std::filesystem::end(iterator))
-			{
-				for (auto const& entry : iterator)
-				{
-					if (fs::is_regular_file(entry) && entry.path().extension() == ".dll")
-						paths.emplace_back(entry.path());
-				}
-			}
-		}
+		findPlugins(pluginDir, paths);
 	}
 
 	if (paths.empty())
--- a/NorthstarDLL/plugins/plugins.cpp
+++ b/NorthstarDLL/plugins/plugins.cpp
@@ -190,18 +190,23 @@ std::optional<Plugin> PluginManager::LoadPlugin(fs::path path, PluginInitFuncs* 
 
 inline void findPlugins(fs::path pluginPath, std::vector<fs::path>& paths)
 {
-	if (fs::exists(pluginPath) && fs::is_directory(pluginPath))
+	// ensure dirs exist
+	if (!fs::exists(pluginPath) || !fs::is_directory(pluginPath))
 	{
-		// ensure dirs exist
-		fs::recursive_directory_iterator iterator(pluginPath);
-		if (std::filesystem::begin(iterator) != std::filesystem::end(iterator))
-		{
-			for (auto const& entry : iterator)
-			{
-				if (fs::is_regular_file(entry) && entry.path().extension() == ".dll")
-					paths.emplace_back(entry.path());
-			}
-		}
+		return;
+	}
+
+	fs::recursive_directory_iterator iterator(pluginPath);
+	// ensure iterator is not empty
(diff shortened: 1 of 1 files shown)


Jan200101:
force pushing because I made an error I do not want to acknowledge

Jan200101:
Worth noting: loading the same plugin multiple times can cause crashes, perhaps this should be checked for? Since this can also be replicated with the normal plugin setup this seems out of scope.

ASpoonPlaysGames:
> Worth noting: loading the same plugin multiple times can cause crashes, perhaps this should be checked for? Since this can also be replicated with the normal plugin setup this seems out of scope.

I agree that this is out of scope, could you make an issue about it? (if one doesn't exist already)

NorthstarDLL/plugins/plugins.cpp
```diff
@@ -190,6 +190,12 @@ std::optional<Plugin> PluginManager::LoadPlugin(fs::path path, PluginInitFuncs*
 
 bool PluginManager::LoadPlugins()
 {
+	if (strstr(GetCommandLineA(), "-noplugins") != NULL)
```
ASpoonPlaysGames, line 193: I know that this line existed prior to this PR, but using `strstr` for command line stuff is kinda bad (elaborated on a bit in #507)

I would appreciate if this could be changed? Not rly needed though, could be considered out of scope

NorthstarDLL/plugins/plugins.cpp
```diff
@@ -207,28 +213,45 @@ bool PluginManager::LoadPlugins()
...
+		{
+			fs::recursive_directory_iterator iterator(pluginDir);
+			if (std::filesystem::begin(iterator) != std::filesystem::end(iterator))
+			{
+				for (auto const& entry : iterator)
+				{
+					if (fs::is_regular_file(entry) && entry.path().extension() == ".dll")
+						paths.emplace_back(entry.path());
+				}
+			}
```
ASpoonPlaysGames, line 245: This should be made into a function, since it is just duplicated from above

ASpoonPlaysGames (changes requested):
Nothing major, code looks good for the most part.

feel free to mark the one about `strstr` as resolved if you don't want to do it

NorthstarDLL/plugins/plugins.cpp
Jan200101: its so little code thats only needed twice, is it really worth making into a dedicated function?

ASpoonPlaysGames: No harm in it, compiler would probably inline it anyway. Since the behaviour for finding plugins should be consistent no matter the starting dir, it should be a function

NorthstarDLL/plugins/plugins.cpp
F1F7Y: #507 will require one PR to remove the .txt functionality and one PR to properly incorporate the tier0 interface so imo we should let `strstr` slide for now

JMM889901:
Tested this locally with titanframework, works at loading from package, it also has no errors loading the same plugin from multiple packages or even from the plugins folder and a package

CooldudePUGS:
also tested locally, seems to work as intended

CooldudePUGS (approved):
ngl I'm terrible with code, just an active helper, but tested pr and it worked perfectly fine locally using [titan framework](https://northstar.thunderstore.io/package/The_Peepeepoopoo_man/Titanframework/) and [framework doublebarrel](https://northstar.thunderstore.io/package/The_Peepeepoopoo_man/framework_Doublebarrel/) (titan framework has a plugin included to replicate safeio, and is a dependency of framework doublebarrel)

Jan200101:
rebased because of the submodule PR, going back before it tends to be annoying for git to manage

NorthstarDLL/plugins/plugins.cpp
```diff
@@ -188,8 +188,31 @@ std::optional<Plugin> PluginManager::LoadPlugin(fs::path path, PluginInitFuncs*
 	return plugin;
 }
 
+inline void findPlugins(fs::path pluginPath, std::vector<fs::path>& paths)
+{
+	if (fs::exists(pluginPath) && fs::is_directory(pluginPath))
```
catornot, line 193: ( just a nitpick ) you could inverse this if statement and return to reduce nesting.

catornot (dismissed):
LGTM

ASpoonPlaysGames (approved):
LGTM

NorthstarDLL/plugins/plugins.cpp
```diff
@@ -188,8 +188,36 @@ std::optional<Plugin> PluginManager::LoadPlugin(fs::path path, PluginInitFuncs*
 	return plugin;
 }
 
+inline void findPlugins(fs::path pluginPath, std::vector<fs::path>& paths)
```
F1F7Y, line 191:
```suggestion
inline void FindPlugins(fs::path pluginPath, std::vector<fs::path>& paths)
```

F1F7Y (changes requested):
Would be cool to have a style guide for launcher just like we're doing with mods ( or at least trying to ), otherwise looks good

we should also get rid of the recursive iterator and settle on a structure

didnt test

NorthstarDLL/plugins/plugins.cpp
Jan200101: the name was explicitly chosen to match `freeLibrary` in the same file.

F1F7Y: We use `PascalCase` in the entire codebase, `freeLibrary` should also follow that

ASpoonPlaysGames:
> LGTM is not a valid review

@GeckoEidechse literally only reviewed to make my older review requesting changes (which was still blocking even tho resolved) not relevant

NorthstarDLL/plugins/plugins.cpp
Jan200101: freeLibrary is a wrapper around FreeLibrary that should probably be generalized and renamed.

F1F7Y: ik, but thats out of scope for this PR

F1F7Y (approved):
Code looks good, didnt test

GeckoEidechse:
@CooldudePUGS tested and @F1F7Y + @ASpoonPlaysGames code reviewed. Let's hope merging and releasing this goes smoothly ^^"

Jan200101:
Reducing the nesting ended up introducing an error that no one caught until rc4.
Fixed by #522

NorthstarDLL/plugins/plugins.cpp
```diff
@@ -188,8 +188,36 @@ std::optional<Plugin> PluginManager::LoadPlugin(fs::path path, PluginInitFuncs*
...
 
+inline void FindPlugins(fs::path pluginPath, std::vector<fs::path>& paths)
+{
+	// ensure dirs exist
+	if (!fs::exists(pluginPath) || !fs::is_directory(pluginPath))
+	{
+		return;
+	}
+
+	fs::recursive_directory_iterator iterator(pluginPath);
```
GeckoEidechse, line 199: Ugh, I just realised another issue with this. 

This doesn't do the regex check that the mod-manager does

https://github.com/R2Northstar/NorthstarLauncher/blob/82bff57d166ece1173ca7df33940171548015a71/NorthstarDLL/mods/modmanager.cpp#L622-L623

Also I'm pretty sure it still loads plugins recursively which we shouldn't need to do anymore. The only reason we added recursive load was to have a plugin -> Thunderstore-package mapping with the old plugin system, i.e. #460 which can also be reverted now.

PR #514, opened 2023-07-21
Replace forward slashes with windows path separators

Jan200101:
Before this PR Northstar mixes forward and backward slashes

```
[22:22:21] [PLUGINSYS] [info] Succesfully loaded R2Northstar/plugins\DiscordRPC.dll
```

With this PR these constants use the windows path separator (backwards slash) instead.


Alystrasz:
I see classical slashes appearing in the logs:
```text
[NORTHSTAR] [info] Building KeyValues for file scripts/aisettings/npc_titan.txt
[NORTHSTAR] [info] Building KeyValues for file scripts/aisettings/npc_titan_ogre_meteor_bounty.txt
```
Is this something you plan to address in this PR?

Jan200101:
> I see classical slashes appearing in the logs:

I don't know how we can solve these without having to process the strings.
The chain appears to go

Various Hooks -> TryReplaceFile -> CompileAssetsForFile -> TryBuildKeyValues (log info is here)

So those path more than likely come from inside the Source Engine.

Alystrasz (approved):
Confirmed building and running, has expected effect on modified paths.

GeckoEidechse:
Not testing myself cause others already did. Just assuming they tested well enough :>

Issue #506, opened 2023-07-12
Preventing "Lost Connection to Origin" when you have no internet connection


jamesla:
I'm hosting a lan in a few weeks which won't have an internet connection and we are trying to make titanfall2 feature as one of the games.

I have the dedicated server running in insecure mode which allows clients to connect without authenticating with the master server and is working fine by opening the client and using `/connect x.x.x.x`.

Now when I disconnect the internet connection altogether it almost connects but then throws the error:
```
Lost connection to Origin
```

And I'm assuming tht's because origin is not logged in - given that there is no internet connection.

Is there any sort of mod/command in order to /connect to a dedicated lan server and not require the client to actively be logged into Origin?

F1F7Y:
There is no cvar or mod to do this.

Going by bobs comments [here](https://github.com/R2Northstar/NorthstarLauncher/blob/main/NorthstarDLL/dedicated/dedicated.cpp#L264-L266) it sounds like even if we allow clients to start without origin the game would throw a fit and probably engine error or crash.

GeckoEidechse:
> Is there any sort of mod/command in order to /connect to a dedicated lan server and not require the client to actively be logged into Origin?

Doubt we can make that work without also enabling pirated versions of the game. Given that the game is still available for purchase (and we don't want to get sued by EA) we have a strong stance against pirating Titanfall2.

jamesla:
Ok fair point - we'll try track down a 3g router or something thanks

PR #516, opened 2023-07-26
Set `spewlog_enable` default value to 0

F1F7Y:
Closes [#506](https://github.com/R2Northstar/Northstar/issues/506)

Would like to add a launch arg in the future ( smth like `-nsdev` ) that would enable all log funcs and other features.

This log call can sometimes slow the game down

### DIFF
--- a/NorthstarDLL/logging/loghooks.cpp
+++ b/NorthstarDLL/logging/loghooks.cpp
@@ -249,7 +249,7 @@ ON_DLL_LOAD_RELIESON("engine.dll", EngineSpewFuncHooks, ConVar, (CModule module)
 {
 	AUTOHOOK_DISPATCH_MODULE(engine.dll)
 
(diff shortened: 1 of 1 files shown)


ASpoonPlaysGames:
Which logs exactly does this disable? my guess would be the `NATIVE SV` (and other VM ones) but not 100% sure. We need to make sure that nothing that may prove useful for tickets and such is getting disabled

F1F7Y:
> Which logs exactly does this disable? my guess would be the `NATIVE SV` (and other VM ones) but not 100% sure. We need to make sure that nothing that may prove useful for tickets and such is getting disabled

Only the `NATIVE SV` ones

ASpoonPlaysGames (approved):
Code looks good (lol)

no `[NATIVE SV]` logging is seen in the console or in log files with `spewlog_enable` set to 0

GeckoEidechse (commented):
Merging based on previous reviews.

PR #517, opened 2023-07-26
Remove `ns_startup_args*.txt` functionality

F1F7Y:
Removes `ns_startup_args*.txt` functionality as discussed at #507 

Main reason for doing this is getting rid of having to hook `GetCommandLineA`, which meant some args would only work when passed directly to the exe.

### DIFF
--- a/NorthstarLauncher/main.cpp
+++ b/NorthstarLauncher/main.cpp
@@ -388,19 +388,13 @@ int main(int argc, char* argv[])
 	{
 		PrependPath();
 
-		if (!fs::exists("ns_startup_args.txt"))
+		if (fs::exists("ns_startup_args.txt"))
 		{
-			std::ofstream file("ns_startup_args.txt");
-			std::string defaultArgs = "-multiple";
-			file.write(defaultArgs.c_str(), defaultArgs.length());
-			file.close();
+			std::cout << "[*] WARNING: 'ns_startup_args.txt' is no longer supported!" << std::endl;
 		}
-		if (!fs::exists("ns_startup_args_dedi.txt"))
+		if (fs::exists("ns_startup_args_dedi.txt"))
 		{
-			std::ofstream file("ns_startup_args_dedi.txt");
-			std::string defaultArgs = "+setplaylist private_match";
-			file.write(defaultArgs.c_str(), defaultArgs.length());
-			file.close();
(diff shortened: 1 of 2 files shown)


NorthstarLauncher/main.cpp
```diff
@@ -419,19 +419,14 @@ int main(int argc, char* argv[])
 	{
 		PrependPath();
 
-		if (!fs::exists("ns_startup_args.txt"))
+		// NOTE [Fifty]: Functionality removed in https://github.com/R2Northstar/NorthstarLauncher/pull/517
+		if (fs::exists("ns_startup_args.txt"))
 		{
-			std::ofstream file("ns_startup_args.txt");
-			std::string defaultArgs = "-multiple";
```
ASpoonPlaysGames, line 425: This PR means we lose the functionality of these default args right?

If so, this could cause issues with the game complaining about multiple instances if an instance is stuck in the background or something

NorthstarLauncher/main.cpp
```diff
@@ -419,19 +419,14 @@ int main(int argc, char* argv[])
...
-			file.write(defaultArgs.c_str(), defaultArgs.length());
-			file.close();
+			std::cout << "[*] WARNING: 'ns_startup_args.txt' is no longer supported!" << std::endl;
 		}
-		if (!fs::exists("ns_startup_args_dedi.txt"))
+		if (fs::exists("ns_startup_args_dedi.txt"))
 		{
-			std::ofstream file("ns_startup_args_dedi.txt");
-			std::string defaultArgs = "+setplaylist private_match";
```
ASpoonPlaysGames, line 432: What is the behaviour for dedis when this functionality is lost?

Also, a number of dedis use `ns_startup_args_dedi.txt` to set up the server. If this gets merged we should warn server hosters about this no longer being supported

GeckoEidechse:
Merge conflicts atm
(hence we should go with compile time feature flag first cause it's less likely to cause merge conflicts and allows for easy revert if need be. Once we determined that we can fully deprecate the feature, by forcing everyone onto a version that no longer has it, we can fully remove it together with the parsing logic)

GeckoEidechse:
Drafted until conflicts are resolved so it's clear which PRs are ready to review atm. Feel free to undraft again at any moment or when you resolved the conflicts.

PR #518, opened 2023-07-27
Update memory lib

F1F7Y:
We've been using an old verison of this adapted from r5reloaded by Bob. This PR moves it out into the `thirdparty` dir as a static lib and updates it to [`72c74b4`](https://github.com/IcePixelx/silver-bun/commit/72c74b455bf4d02b424096ad2f30cd65535f814c)

### Why update
Improvements in behaviour ( no longer using `WriteProcessMemory` which is quite limited ), some new featurs such as `PatchString`

### Why move to thirdparty
Imo we should also use this in other targets ( eg NorthstarLauncher and wsock proxy ) to improve code readability


GeckoEidechse (commented):
I assume this is a modifed version of https://github.com/IcePixelx/silver-bun/commit/72c74b455bf4d02b424096ad2f30cd65535f814c ?

Cause if unmodified we could even just submodule it :eyes:

F1F7Y:
> I assume this is a modifed version of [IcePixelx/silver-bun@72c74b4](https://github.com/IcePixelx/silver-bun/commit/72c74b455bf4d02b424096ad2f30cd65535f814c) ?
> 
> Cause if unmodified we could even just submodule it 👀

It is modifed as the original lib focuses on pattern scanning ( which we dont do apart from like two cases ). Second reason is to submodule this we would need to rewrite some patches as the original lib doesnt have some versions of the `Patch` function which those rely on

GeckoEidechse:
Alright, once tested/reviewed I might actually add libs separately then and commit them under icepixel's name so that we can better tell what's copied/what's modified when looking at commit history later on ^^

ASpoonPlaysGames (approved):
Code looks fine as far as I can tell. Didn't look at the files in `thirdparty` though since idk what stuff is modified or not

uniboi:
should've used `git mv` to move `memory.*` tbh so the history doesn't get deleted

GeckoEidechse:
Closing as superseded by #664
(made sure to properly co-author you on the new PR by @ASpoonPlaysGames btw so you still get it properly credited in the stats <3)

PR #519, opened 2023-07-27
Remove `pch.cpp` and all `#include "pch.h"` occurrences 

F1F7Y:
With the move to cmake there is no need for the `pch.cpp` files as it generates them automatically.

Also removes leftover `#include "pch.h"` occurrences


ASpoonPlaysGames (approved):
It does in fact compile, and it also does what it says it does

GeckoEidechse (approved):
Game didn't crash when I launched it and joined a random server so given that this change removes some header files, which if wrong would've given compile errors anyway, should be enough testing ig :P

PR #521, opened 2023-07-27
Fix crash in `ReadFileAsync`

### LINKED ISSUE #520: NULL Deref crash in `ReadFileAsyncHook`
`pPath` [here](https://github.com/R2Northstar/NorthstarLauncher/blob/main/NorthstarDLL/core/filesystem/rpakfilesystem.cpp#L259) is the culprit. This seems to be very rare

Stack trace:
```
Northstar.dll + 0xa3c23
materialsystem_dx11.dll + 0x9955b
materialsystem_dx11.dll + 0x99396
materialsystem_dx11.dll + 0x8c12
materialsystem_dx11.dll + 0x8808f
tier0.dll + 0x1ba3b
KERNEL32.DLL + 0x17614
ntdll.dll + 0x526b1
```

Ill try to figure out why this is happening, but either way will PR a fix

F1F7Y:
Fixes a NULL dereference crash in `ReadFileAsync`

This seems to be very rare, there's currently an open [ticket](https://discord.com/channels/920776187884732556/1134199951971332096) where this occurs. This version hasnt been tested yet.

I also promised to ping `@shmordan_08` when this gets released

Closes #520


GeckoEidechse:
Let me know when the person from the ticket has tested the version from this PR and we can mark the testing part as complete ^^

NorthstarDLL/core/filesystem/rpakfilesystem.cpp
```diff
@@ -255,11 +255,20 @@ void*, __fastcall, (int nPakHandle, void* pCallback))
...
 HOOK(ReadFileAsyncHook, ReadFileAsync, 
 void*, __fastcall, (const char* pPath, void* pCallback))
 // clang-format on
 {
+	// NOTE [Fifty]: For some reason some users are getting pPath as null when
+	//               loading a server, ReadFileAsync uses CreateFileA and checks
+	//               its return value so this is completely safe
+	if (pPath == NULL)
+	{
+		return ReadFileAsync(pPath, pCallback);
```
ASpoonPlaysGames, line 269: Why not just `return nullptr`? does `ReadFileAsync` do any work here that matters?

ASpoonPlaysGames (approved):
Code looks good, just a question rly

NorthstarDLL/core/filesystem/rpakfilesystem.cpp
Jan200101: because nullptr would be the wrong value.

Looking at a decompile, `ReadFileAsync` calls `CreateFileA` and if it returns `INVALID_HANDLE_VALUE` it returns -1 as void*
Code that relies on this would break instantly if we return anything else

NorthstarDLL/core/filesystem/rpakfilesystem.cpp
```diff
@@ -255,11 +255,20 @@ void*, __fastcall, (int nPakHandle, void* pCallback))
...
+// rtech_game.dll + 0x1e20
 // clang-format off
 HOOK(ReadFileAsyncHook, ReadFileAsync, 
 void*, __fastcall, (const char* pPath, void* pCallback))
 // clang-format on
 {
+	// NOTE [Fifty]: For some reason some users are getting pPath as null when
+	//               loading a server, ReadFileAsync uses CreateFileA and checks
+	//               its return value so this is completely safe
+	if (pPath == NULL)
```
Jan200101, line 267: There may be other possible path values we want to skip over

```c
if (IsBadStringPtrA(pPath, -1))
```
should do the trick

pg9182: > **Important:** This function is obsolete and should not be used. Despite its name, it does not guarantee that the pointer is valid or that the memory pointed to is safe to use. For more information, see Remarks on this page.

Jan200101: I suggested it because of its use within the wine codebase, which appears to have a reliable implementation of it, but it seems less reliable on Windows.

Here is a post by Raymond Chen about it from 2006 https://devblogs.microsoft.com/oldnewthing/20060927-07/?p=29563

GeckoEidechse:
I assume you never got the person from the ticket to test it? I'm kinda unsure how to best test it otherwise? Should we just forego testing an merge without it?

GeckoEidechse:
@F1F7Y bump

Alystrasz:
@F1F7Y sup

GeckoEidechse:
Poking @Alystrasz to recreate this PR when you get the chance ^^"

PR #523, opened 2023-07-28
Make sure `packages` dir exists when loading plugins

F1F7Y:
Fixes a crash introduced in #513 which occured when no `<Profile>/packages` dir was present

fs::create_directories doesnt throw when the dir already exists so its safe calling this in both PluginManager and ModManager

### DIFF
--- a/NorthstarDLL/plugins/plugins.cpp
+++ b/NorthstarDLL/plugins/plugins.cpp
@@ -211,6 +211,8 @@ bool PluginManager::LoadPlugins()
 		return false;
 	}
 
+	fs::create_directories(GetThunderstoreModFolderPath());
+
 	std::vector<fs::path> paths;
 
 	pluginPath = GetNorthstarPrefix() + "/plugins";


F1F7Y:
> fs::create_directories doesnt throw when the dir already exists so its safe calling this in both PluginManager and ModManager

We should have a better system for ensuring folder structure, but this fix works as is and doesnt clutter up the codebase much

Jan200101 (approved):
Can confirm this solves the crash when `packages` is missing.
Appears to have no other side effect so I didn't test further.

GeckoEidechse (approved):
Bug confirmed in testing using new profile that doesn't have `packages` folder and confirmed resolved with this PR.

As for code review, this PR only adds a single function call to create the missing directory.

F1F7Y:
> As for code review, this PR only adds a single function call to create the missing directory.

std::filesystem functions can throw exceptions so id advise reading up when they throw when reviewing :). ( The bug this PR fixes is in one way caused by me not doing that )

PR #525, opened 2023-07-29
Validate package pattern before checking for plugins

### LINKED ISSUE #524: Plugins load from packages is missing regex check
> Ugh, I just realised another issue with this. 
> 
> This doesn't do the regex check that the `modmanager.cpp` code does to ensure that package is installed correctly.
> 
> https://github.com/R2Northstar/NorthstarLauncher/blob/82bff57d166ece1173ca7df33940171548015a71/NorthstarDLL/mods/modmanager.cpp#L622-L623

> Also I'm pretty sure it still loads plugins recursively which we shouldn't need to do anymore. The only reason we added recursive load was to have a plugin -> Thunderstore-package mapping with the old plugin system, i.e. #460 which can also be reverted now.

_Originally posted by @GeckoEidechse in https://github.com/R2Northstar/NorthstarLauncher/pull/513#discussion_r1278286566_

Jan200101:
Closes #524

Added the package pattern to validate if the package is actually valid, copied the code structure from `modmanager.cpp` so its easier to catch anything misbehaving.

### DIFF
--- a/NorthstarDLL/mods/modmanager.cpp
+++ b/NorthstarDLL/mods/modmanager.cpp
@@ -16,6 +16,7 @@
 #include <string>
 #include <sstream>
 #include <vector>
+#include <regex>
 
 ModManager* g_pModManager;
 
--- a/NorthstarDLL/mods/modmanager.cpp
+++ b/NorthstarDLL/mods/modmanager.cpp
@@ -628,7 +628,7 @@ void ModManager::LoadMods()
 		// Use regex to match `AUTHOR-MOD-VERSION` pattern
 		if (!std::regex_match(dir.path().string(), pattern))
 		{
-			spdlog::warn("The following directory did not match 'AUTHOR-MOD-VERSION': {}", modsDir.string());
+			spdlog::warn("The following directory did not match 'AUTHOR-MOD-VERSION': {}", dir.path().string());
 			continue; // skip loading mod that doesn't match
 		}
 		if (fs::exists(modsDir) && fs::is_directory(modsDir))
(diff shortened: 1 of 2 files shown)


Jan200101:
There appears to be a mising include for regex in plugins and modmanager, I'll add them in a bit.

Jan200101 (commented):
Tested locally and played a few matches of Attrition and half a game of gun game
```
[14:22:33] [PLUGINSYS] [info] Succesfully loaded R2Northstar\plugins\DiscordRPC.dll
[14:22:33] [PLUGINSYS] [info] Loading plugin DiscordRPC version 2.0
[14:22:33] [DiscordRPC] [info] plugin logging initialized
[14:22:33] [DiscordRPC] [info] waiting for handshake...
[14:22:33] [PLUGINSYS] [info] Succesfully loaded R2Northstar\packages\cat_or_not-kyurid-0.1.1\plugins\kyurid.dll
[14:22:33] [PLUGINSYS] [info] Loading plugin KYURIDSYS version 1.0
[14:22:33] [KYURIDSYS] [info] plugin logging initialized
```
Didn't check if the regex works.

NorthstarDLL/mods/modmanager.cpp
```diff
@@ -16,6 +16,7 @@
 #include <string>
 #include <sstream>
 #include <vector>
+#include <regex>
```
GeckoEidechse, line 19: I'm a bit confused why this include is needed bow but wasn't before ^^"

GeckoEidechse (approved):
Confirmed working in testing by installing [`cat_or_not-Furnace-1.0.0`](https://northstar.thunderstore.io/package/cat_or_not/Furnace/1.0.0) in `packages` folder, then renaming it so it doesn't match the checked pattern, then loading the game with/without the PR.

Without it still loads the plugin, without it doesn't.

Renaming the directory back to original, plugin loads fine again with PR.

NorthstarDLL/mods/modmanager.cpp
GeckoEidechse: TL;DR: doesn't hurt and can help with preventing issues in case we shuffle the headers around later.

PR #526, opened 2023-07-29
Warn about the package dir, not the mod dir

Jan200101:
was told to separate it this from #525


Jan200101:
We already know that `dir.path().string()` is valid in this scope, so I don't think there is a need to re-test.

ASpoonPlaysGames (approved):
Code looks good, I don't think I rly need to elaborate more on this but I will anyway to appease the reptile.

it is not the modsDir that should be required to follow `AUTHOR-MOD-VERSION`, just above we validate the actually correct dir, but then we print the wrong one here. This PR fixes that and is good

GeckoEidechse (approved):
Confirmed working as intended by creating an empty folder whose name doesn't match the `AUTHOR-MOD-VERSION` pattern.

PR #534, opened 2023-08-20
Fix proton detection output

### LINKED ISSUE #533: Proton detection prints incorrect basename
Detecting while working on Proton support

Newer proton versions also list the steam linux runtime in their compat tool paths.

Here is an example of the value it may be set to
```
STEAM_COMPAT_TOOL_PATHS=/home/user/.local/share/Steam/compatibilitytools.d/NorthstarProton:/home/user/.local/share/Steam/steamapps/common/SteamLinuxRuntime_sniper
```

Jan200101:
fixes #533 

image

Also adds some comments to explain what is going on and puts values into variables instead of using the return value directly.
Yes I used auto, sue me.

### DIFF
--- a/NorthstarDLL/logging/logging.cpp
+++ b/NorthstarDLL/logging/logging.cpp
@@ -261,12 +261,19 @@ void StartupLog()
 		spdlog::info("Operating System: {} (Wine)", sysname);
 		spdlog::info("Wine build: {}", wine_get_build_id());
 
+
+		// STEAM_COMPAT_TOOL_PATHS is a colon separated lists of all compat tool paths used
+		// The first one tends to be the Proton path itself
+		// We extract the basename out of it to get the name used
 		char* compatToolPtr = std::getenv("STEAM_COMPAT_TOOL_PATHS");
 		if (compatToolPtr)
 		{
 			std::string compatToolPath(compatToolPtr);
 
-			spdlog::info("Proton build: {}", compatToolPath.substr(compatToolPath.rfind("/") + 1));
+			auto protonBasenameEnd = compatToolPath.find(":");
+			auto protonBasenameStart = compatToolPath.rfind("/", protonBasenameEnd) + 1;
+
+			spdlog::info("Proton build: {}", compatToolPath.substr(protonBasenameStart, protonBasenameEnd - protonBasenameStart));
 		}
 	}
 	else
--- a/NorthstarDLL/logging/logging.cpp
+++ b/NorthstarDLL/logging/logging.cpp
@@ -270,7 +270,11 @@ void StartupLog()
 			std::string compatToolPath(compatToolPtr);
 
 			auto protonBasenameEnd = compatToolPath.find(":");
+			if (protonBasenameEnd == std::string::npos)
+				protonBasenameEnd = 0;
 			auto protonBasenameStart = compatToolPath.rfind("/", protonBasenameEnd) + 1;
+			if (protonBasenameStart == std::string::npos)
+				protonBasenameStart = 0;
 
 			spdlog::info("Proton build: {}", compatToolPath.substr(protonBasenameStart, protonBasenameEnd - protonBasenameStart));
 		}
--- a/NorthstarDLL/logging/logging.cpp
+++ b/NorthstarDLL/logging/logging.cpp
@@ -267,13 +267,13 @@ void StartupLog()
 		char* compatToolPtr = std::getenv("STEAM_COMPAT_TOOL_PATHS");
 		if (compatToolPtr)
 		{
-			std::string compatToolPath(compatToolPtr);
+			std::string_view compatToolPath(compatToolPtr);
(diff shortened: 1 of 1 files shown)


NorthstarDLL/logging/logging.cpp
```diff
@@ -261,12 +261,18 @@ void StartupLog()
 		spdlog::info("Operating System: {} (Wine)", sysname);
 		spdlog::info("Wine build: {}", wine_get_build_id());
 
+		// STEAM_COMPAT_TOOL_PATHS is a colon separated lists of all compat tool paths used
+		// The first one tends to be the Proton path itself
```
ASpoonPlaysGames, line 265: Tends to be? Doesn't this mean that this PR only fixes the cases where it was broken, but breaks cases where it wasn't?

Could we just look for the `NorthstarProton` + `:`/`0x00` substring to find it more consistently?

NorthstarDLL/logging/logging.cpp
```diff
@@ -261,12 +261,18 @@ void StartupLog()
...
+		// We extract the basename out of it to get the name used
 		char* compatToolPtr = std::getenv("STEAM_COMPAT_TOOL_PATHS");
 		if (compatToolPtr)
 		{
 			std::string compatToolPath(compatToolPtr);
 
-			spdlog::info("Proton build: {}", compatToolPath.substr(compatToolPath.rfind("/") + 1));
+			auto protonBasenameEnd = compatToolPath.find(":");
```
ASpoonPlaysGames, line 272: If there is only one compat tool, wont this be `std::string::npos`? This wouldn't break the later code would it?

NorthstarDLL/logging/logging.cpp
Jan200101: > Tends to be? Doesn't this mean that this PR only fixes the cases where it was broken, but breaks cases where it wasn't?

Entries to the list are appended to the end, so when Proton steam puts it as the first entry.
But this isn't set in stone and may change in the future.

> Could we just look for the NorthstarProton + :/0x00 substring to find it more consistently?

NorthstarProton is not a requirement, only a suggestion.
GE-Proton works just as well (arguably even better).

a colon is only added between multiple entries.

NorthstarDLL/logging/logging.cpp
Jan200101: yeah, it would.

Jan200101: Should be fixed by c92531ff6a1ade76546589fae16d50e5cc7b630f

NorthstarDLL/logging/logging.cpp
ASpoonPlaysGames: > NorthstarProton is not a requirement, only a suggestion.
> GE-Proton works just as well (arguably even better).

Fair enough. I guess we can't really guarantee 100% accuracy here then

ASpoonPlaysGames (approved):
Code looks good :)

Jan200101:
Maybe the log entry should say "Compatibility tool" instead of "Proton version"

ASpoonPlaysGames:
> Maybe the log entry should say "Compatibility tool" instead of "Proton version"

If so, is there a downside to just logging the entirety of `STEAM_COMPAT_TOOL_PATHS`?

Jan200101:
> If so, is there a downside to just logging the entirety of `STEAM_COMPAT_TOOL_PATHS`?
no, but it could become a bit cluttered since the main compat tool can pull in a variety of other things.

NorthstarDLL/logging/logging.cpp
```diff
@@ -261,12 +261,22 @@ void StartupLog()
 		spdlog::info("Operating System: {} (Wine)", sysname);
 		spdlog::info("Wine build: {}", wine_get_build_id());
 
+		// STEAM_COMPAT_TOOL_PATHS is a colon separated lists of all compat tool paths used
+		// The first one tends to be the Proton path itself
+		// We extract the basename out of it to get the name used
 		char* compatToolPtr = std::getenv("STEAM_COMPAT_TOOL_PATHS");
 		if (compatToolPtr)
 		{
 			std::string compatToolPath(compatToolPtr);
```
p0358, line 270: You should change `std::string` to `std::string_view`, as there's no point allocating a new string, whereas string_view was designed exactly for this. Remember to change it for `npos` too.

p0358: Btw spdlog's formatter would cast std::string to basic_string_view anyways

Jan200101: Yeah, in theory there should be nothing that could mess with the env pointers so string_view should be fine.

H0L0theBard (approved):
Code looks good and works as intended :+1:

image

GeckoEidechse:
Merging based on reviews.

PR #535, opened 2023-08-31
feat: Enforce Thunderstore format for remote mods

Alystrasz:
Implements https://github.com/R2Northstar/NorthstarLauncher/pull/503 for remote mods (*i.e.* mods located in the `%profile%/runtime/remote/mods` directory), expecting them to be named after the `AUTHOR-MOD-VERSION` pattern.

### Testing this

Put mods respecting the Thunderstore format in the remote mods directory, and they should be loaded and usable just like mods located in the `packages` directory.

### DIFF
--- a/NorthstarDLL/mods/modmanager.cpp
+++ b/NorthstarDLL/mods/modmanager.cpp
@@ -612,32 +612,36 @@ void ModManager::LoadMods()
 	// get mod directories
 	std::filesystem::directory_iterator classicModsDir = fs::directory_iterator(GetModFolderPath());
 	std::filesystem::directory_iterator remoteModsDir = fs::directory_iterator(GetRemoteModFolderPath());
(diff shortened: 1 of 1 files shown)


GeckoEidechse (approved):
Confirmed working in testing.
I took `GeckoEidechse-Titanfall_Slander-0.0.1` and put it under `R2Northstar\runtime\remote\mods` and it loaded fine.
Renaming to `GeckoEidechse-Titanfall_Slander-lol-0.0.1`  cause the mod loading to fail as expected due to not matching the required format.
I also verified that normal package and mod loading still works by sticking the same mod in`packages/` and `mods/` while formatting it accordingly.

As for code review, all this PR does is basically move the mod loading for remote mods from the logic that deals with "classic" mods to the one that deals with Thunderstore mods. So it looks like a lot of changed lines while the logical change is actually rather small.

GeckoEidechse:
ugh, just as I merged I forgot to remove myself as co-author for updating branch, whoops x_x

PR #538, opened 2023-09-03
Refactor `debugoverlay.cpp` (fix smart pistol lines)

ASpoonPlaysGames:
closes https://github.com/R2Northstar/NorthstarMods/issues/689

This code is probably bad idk dude this is weird stuff. Would be cool to get someone better at cpp than me to review this

### DIFF
--- a/NorthstarDLL/client/debugoverlay.cpp
+++ b/NorthstarDLL/client/debugoverlay.cpp
@@ -75,6 +75,13 @@ typedef void (*RenderBoxType)(Vector3 vOrigin, QAngle angles, Vector3 vMins, Vec
 static RenderBoxType RenderBox;
 static RenderBoxType RenderWireframeBox;
 
+typedef bool (*OverlayBase_t__IsDeadType)(OverlayBase_t* a1);
+static OverlayBase_t__IsDeadType OverlayBase_t__IsDead;
+typedef void (*OverlayBase_t__DestroyOverlayType)(OverlayBase_t* a1);
+static OverlayBase_t__DestroyOverlayType OverlayBase_t__DestroyOverlay;
+
+static ConVar* Cvar_enable_debug_overlays;
+
 // clang-format off
 AUTOHOOK(DrawOverlay, engine.dll + 0xABCB0, 
 void, __fastcall, (OverlayBase_t * pOverlay))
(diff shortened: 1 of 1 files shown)


Alystrasz (commented):
I cannot approve merging this code since I don't understand it; I can however confirm it seems to be working:

##### Before

Capture d’écran du 2023-09-18 20-30-57

##### After

Capture d’écran du 2023-09-18 20-53-12

ASpoonPlaysGames:
Re-requesting testing due to large code changes

NorthstarDLL/client/debugoverlay.cpp
```diff
@@ -119,21 +212,136 @@ void, __fastcall, (OverlayBase_t * pOverlay))
...
+				{
+					bShouldDraw = m_pUnk == *g_nOverlayTickCount;
+				}
+			}
+			else
+			{
+				bShouldDraw = pCurrOverlay->m_nCreationTick == *g_nRenderTickCount;
+			}
+
+			if (bShouldDraw && bRender && (Cvar_enable_debug_overlays->GetBool() || pCurrOverlay->m_Type == OVERLAY_SMARTAMMO))
```
Jan200101, line 307: Running this shouldn't impact performance, but its pontless to do so if bRender is false.

Maybe this should be in an `if (bRender) {}`? might hurt readability though.

Jan200101 (approved):
tested, re-adds the smart pistol overlay. Did not compare to vanilla.

Did not verify offsets, code looks good, just had 1 thought

NorthstarDLL/client/debugoverlay.cpp
ASpoonPlaysGames: > Running this shouldn't impact performance, but its pontless to do so if bRender is false.

C++ has short circuiting, so this really doesn't matter tbh

Jan200101: eh, sure.
I was mostly thinking about not running code thats not needed, but its so little code t hat any CPU capable of running TF|2 should have this done in a few cycles.

ASpoonPlaysGames: If you mean running the majority of the function, it's needed so that expired debug draws get cleared from the list

Jan200101: Nah, I just mean the part directly above the if condition

GeckoEidechse:
Technically missing proper code review but given that code was originally written by @F1F7Y and adapted by @ASpoonPlaysGames I think it should be fine ^^

ASpoonPlaysGames:
> Technically missing proper code review but given that code was originally written by @F1F7Y and adapted by @ASpoonPlaysGames I think it should be fine ^^

???
image

GeckoEidechse:
:clueless:

PR #539, opened 2023-09-04
Convar print commands

H0L0theBard:
Adds various concommands such as 
- `convar_findByFlags`
- `convar_list`
- `convar_differences`
- `convar_find`

The first 3 listed above are already registered as concommands natively but didn't seem to do anything when tested.
`convar_findByFlags` and `convar_find` were already implemented by Bob under the names `findflags` and `find` respectively but the names have been changed to reflect already existing convars. 

All outputs into the console are also alphabetically sorted.
image

### DIFF
--- a/NorthstarDLL/util/printcommands.cpp
+++ b/NorthstarDLL/util/printcommands.cpp
@@ -160,14 +160,50 @@ void ConCommand_findflags(const CCommand& arg)
 	delete[] upperFlag;
 }
 
+void ConCommand_list(const CCommand& arg)
+{
+	for (auto& map : R2::g_pCVar->DumpToMap())
+	{
+		PrintCommandHelpDialogue(map.second, map.second->m_pszName);
+	}
+}
+
+void ConCommand_differences(const CCommand& arg)
+{
+	for (auto& map : R2::g_pCVar->DumpToMap())
+	{
+		ConVar* cvar = R2::g_pCVar->FindVar(map.second->m_pszName);
+		if (cvar && strcmp(cvar->GetString(), "FCVAR_NEVER_AS_STRING") != NULL)
+		{
+			if (strcmp(cvar->GetString(), cvar->m_pszDefaultValue) != NULL)
+			{
+				PrintCommandHelpDialogue(map.second, map.second->m_pszName);
+				spdlog::info("Current Value: {}", cvar->m_Value.m_pszString);
+				spdlog::info("Default Value: {}", cvar->m_pszDefaultValue);
+			}
+		}
+	}
+}
+
 void InitialiseCommandPrint()
 {
-	RegisterConCommand("find", ConCommand_find, "Find concommands with the specified string in their name/help text.", FCVAR_NONE);
-	RegisterConCommand("findflags", ConCommand_findflags, "Find concommands by flags.", FCVAR_NONE);
+	RegisterConCommand("convar_find", ConCommand_find, "Find concommands with the specified string in their name/help text.", FCVAR_NONE);
 
-	// help is already a command, so we need to modify the preexisting command to use our func instead
+	// these commands already exist, so we need to modify the preexisting command to use our func instead
 	// and clear the flags also
 	ConCommand* helpCommand = R2::g_pCVar->FindCommand("help");
 	helpCommand->m_nFlags = FCVAR_NONE;
 	helpCommand->m_pCommandCallback = ConCommand_help;
+
+	ConCommand* findCommand = R2::g_pCVar->FindCommand("convar_findByFlags");
+	findCommand->m_nFlags = FCVAR_NONE;
+	findCommand->m_pCommandCallback = ConCommand_findflags;
+
+	ConCommand* listCommand = R2::g_pCVar->FindCommand("convar_list");
+	listCommand->m_nFlags = FCVAR_NONE;
+	listCommand->m_pCommandCallback = ConCommand_list;
+
+	ConCommand* diffCommand = R2::g_pCVar->FindCommand("convar_differences");
+	diffCommand->m_nFlags = FCVAR_NONE;
+	diffCommand->m_pCommandCallback = ConCommand_differences;
 }
--- a/NorthstarDLL/util/printcommands.cpp
+++ b/NorthstarDLL/util/printcommands.cpp
@@ -1,4 +1,5 @@
 #include "printcommands.h"
+#include "core/convar/cvar.h"
 #include "core/convar/convar.h"
 #include "core/convar/concommand.h"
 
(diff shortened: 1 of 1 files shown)


ASpoonPlaysGames (approved):
Code looks good!

Asha097:
I tested this and now want to implode (It works)

Goose645:
works (no crashes for 30 min)

NorthstarDLL/util/printcommands.cpp
```diff
@@ -2,6 +2,16 @@
 #include "core/convar/convar.h"
 #include "core/convar/concommand.h"
 
+std::vector<std::pair<std::string, ConCommandBase*>> ConvarSort(std::unordered_map<std::string, ConCommandBase*> map)
```
catornot, line 5: `std::vector<std::pair<std::string, ConCommandBase*>>` could be replaced with a shorter type like `ConCommandBases`. I just think these type names are too long.
 
 ```suggestion
typedef std::vector<std::pair<std::string, ConCommandBase*>> ConCommandBases;
 
ConCommandBases ConvarSort(std::unordered_map<std::string, ConCommandBase*> map)
 ```

catornot (approved):
the code looks good, I just have a nitpick.

NorthstarDLL/util/printcommands.cpp
F1F7Y: That just makes it less readable. If you want to go for readability the best course of action would be to get rid of `CCvar::DumpToMap` and use the internal iterator instead.

Another thing is imo ( gecko also mentioned this somewhere ) is we should strive to mimic engine structure part of that being using engine provided interfaces as much as possible. Making wrappers around wrappers just worsens code quality and maintainability.

H0L0theBard: If it's preferred, I can rework all the convars to use the internal iterator instead of using `CCvar::DumpToMap`.
I had only used `CCvar::DumpToMap` as it matched the previous bob code™ which was already in the file.

H0L0theBard:
Going to be reworking the code to use `g_pCVar->FactoryInternalIterator()` and fixing minor issues which are still in this pr

Alystrasz:
@H0L0theBard 'sup?

NorthstarDLL/util/printcommands.cpp
H0L0theBard: Switched to Internal Iterator in latest commit

cyrv6737:
PR still working for me after 00bab5dc76bc13dec12199cceeb70f96d9833184

Asha097:
Tested again and still want to explode (works)

itscynxx:
was able to run all listed commands with proper outputs and no issues 👍

ASpoonPlaysGames:
(Whoops, missed that code needs re-reviewing since changes)

NorthstarDLL/util/printcommands.cpp
```diff
@@ -160,14 +200,93 @@ void ConCommand_findflags(const CCommand& arg)
...
+					}
+					if (cvar->m_bHasMax)
+					{
+						formatted.append(fmt::format(" max. {}", cvar->m_fMaxVal));
+					}
+					formatted.append(fmt::format(" - {}", cvar->GetHelpText()));
+					spdlog::info(formatted);
+				}
+			}
+		}
```
ASpoonPlaysGames, line 266: Would prefer guard clauses here instead of nesting

NorthstarDLL/util/printcommands.cpp
```diff
@@ -94,7 +105,22 @@ void ConCommand_find(const CCommand& arg)
 	char pTempName[256];
 	char pTempSearchTerm[256];
 
-	for (auto& map : R2::g_pCVar->DumpToMap())
+	ConCommandBase* var;
+	CCVarIteratorInternal* itint = R2::g_pCVar->FactoryInternalIterator();
+	std::unordered_map<std::string, ConCommandBase*> unsortedConvars;
```
ASpoonPlaysGames, line 110: Since `ConvarSort` is a basic sort, could this become a normal map instead of an unordered one?

ASpoonPlaysGames (changes requested):
Minor change and a question

H0L0theBard:
I just realised that convar_list (cvarlist in valve source) has the ability to make a list from a search and log to a file.

`cvarlist mat_` - creates a cvarlist of all "mat_" commands
`cvarlist log allconvars.txt` - logs cvarlist to allconvars.txt in game directory 
and also convar_findByFlags is missing an autocomplete callback.

Probably should implement these eventually. 
Wouldn't really change much so I can perhaps do it in another pr as the basic functionality is added in this one

NorthstarDLL/util/printcommands.cpp
H0L0theBard: That entire sort is pointless as std::map already does that. 
Me when writing that code:


Will fix shortly

H0L0theBard: Resolved in dc94412a9bb16d6ffca2dc3b8d3171ce7e128ccd

NorthstarDLL/util/printcommands.cpp
H0L0theBard: Resolved in fa9b19e5b7b3b86610b096debca674f5155340be

NorthstarDLL/util/printcommands.cpp
```diff
@@ -160,14 +186,98 @@ void ConCommand_findflags(const CCommand& arg)
...
+			continue;
+		}
+
+		if (strcmp(cvar->GetString(), "FCVAR_NEVER_AS_STRING") == NULL)
+		{
+			continue;
+		}
+
+		if (strcmp(cvar->GetString(), cvar->m_pszDefaultValue) != NULL)
+		{
```
ASpoonPlaysGames, line 241: one more guard clause?

H0L0theBard: didn't like the look of the code with that one as a guard clause but sure 
1 sec

H0L0theBard: :+1: 135be3ef1ebfe9bebb8c8978c14cb9d7a70755ca

ASpoonPlaysGames (approved):
Code looks good, requested changes were addressed :)

itscynxx:
Was able to open the console and run all commands with proper outputs with the newest version 👍

ASpoonPlaysGames:
seems that @GeckoEidechse forgot to add the "ready to merge" label when removing "needs code review"

GeckoEidechse (approved):
Seems to be working fine in quick testing.

PR #544, opened 2023-09-08
Fix `CSquirrelVM::Init` hook in `server.dll`

F1F7Y:
Fixes typo in a offset we hook

**Testing:**

For testing git clone https://github.com/F1F7Y/ServerUtilities and checkout the `fsu3` branch. Then go to lobby ( or activate the server however you want )

On current main you'll script error because the initscript wasn't compiled because the hook is wrong. On this branch you won't script error.


F1F7Y:
For testing git clone https://github.com/F1F7Y/ServerUtilities and checkout the `fsu3` branch. Then go to lobby ( or activate the server however you want )

On current main you'll script error because the initscript wasnt compiled because the hook is wrong. On this branch you wont script error.

ASpoonPlaysGames:
Well, this would explain some stuff

ASpoonPlaysGames (approved):
image
New offset seems to line up properly, compared to current offset, which is just in the middle of nowhere: 
image

Also works in testing, InitScript stuff works on SERVER VM

GeckoEidechse (approved):
Looking at the diff, it looks very obviously that previous version was a typo. I trust @ASpoonPlaysGames to properly have tested this, so this approval is just sign-off without testing.

(technically I guess I did code review now to some degree :P)

PR #546, opened 2023-09-10
Fix tokens for bots

catornot:
If bots connect before a player ever connects to the server, a uninit [pointer](https://github.com/R2Northstar/NorthstarLauncher/blob/07e76e3a8e2738dbb7a1d5a6aeaa908a838f5a02/NorthstarDLL/server/auth/serverauthentication.cpp#L207C7-L207C23) will be accessed. this happens because bots skip `CBaseServer__ConnectClient`.

this pr fixes this issue.

### DIFF
--- a/NorthstarDLL/server/auth/serverauthentication.cpp
+++ b/NorthstarDLL/server/auth/serverauthentication.cpp
@@ -101,7 +101,7 @@ bool ServerAuthenticationManager::IsDuplicateAccount(R2::CBaseClient* pPlayer, c
 	return false;
 }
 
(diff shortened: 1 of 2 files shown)


F1F7Y:
Tbh this is a bandage solution and a better way of fixing this would be to move auth entirely into `CBaseServer::ConnectClient` while keeping bot specific code in `CClient::Connect`

catornot:
I think at least token and uid should be always init.

F1F7Y:
> I think at least token and uid should be always init.

Moving auth into `CBaseServer::ConnectClient` would mean we wouldnt have to have file scope variables for serverfilter and uid completely fixing this issue

catornot:
ok then I'll do another pr

PR #548, opened 2023-09-14
Move player auth to `CServer::ConnectClient`

catornot:
Fixes bots crashing servers when they are the first to connect to it. Also moves player auth to `CServer::ConnectClient`. This allows the removal of `iNextPlayerUid` and `pNextPlayerToken` which were the cause of issues previously since they were not initialized by bots.

I have tested this quickly and it worked for me :)


Alystrasz:
Could you provide a detailed test scenario for me to test your PR?

catornot:
# how to test this
1. try to connect to anything a local server to test if this still works
2. spawn bots on a dedicated server before any real players ( crash == doesn't work D: )
bots plugin: [bots_plugin_v2.zip](https://github.com/R2Northstar/NorthstarLauncher/files/12653599/bots_plugin_v2.zip)

Alystrasz:
Gotta check the code of this pull request yet.
I already gave it a try though, and it seems like it's working as intented.

#### Without this PR

spawn_bot_without_pr

#### With this PR

spawn_bot_with_pr

NorthstarDLL/server/auth/serverauthentication.cpp
```diff
@@ -229,51 +227,72 @@ void*,, (
...
+	// since bots take player slots and it will give if not counted a false player count on the server browser.
+
+	if (!bFakePlayer)
+		return CBaseClient__Connect(self, pName, pNetChannel, bFakePlayer, a5, pDisconnectReason, a7);
+
+	// try to actually connect the bot
+	else if (!CBaseClient__Connect(self, pName, pNetChannel, bFakePlayer, a5, pDisconnectReason, a7))
 		return false;
 
 	// we already know this player's authentication data is legit, actually write it to them now
```
ASpoonPlaysGames, line 292: This comment doesn't seem accurate, given the logic changes?

NorthstarDLL/server/auth/serverauthentication.cpp
```diff
@@ -229,51 +227,72 @@ void*,, (
...
+{
+	// only remains to count bots in player count,
+	// since bots take player slots and it will give if not counted a false player count on the server browser.
+
+	if (!bFakePlayer)
+		return CBaseClient__Connect(self, pName, pNetChannel, bFakePlayer, a5, pDisconnectReason, a7);
+
+	// try to actually connect the bot
+	else if (!CBaseClient__Connect(self, pName, pNetChannel, bFakePlayer, a5, pDisconnectReason, a7))
 		return false;
```
ASpoonPlaysGames, line 290: This control flow could be improved a bit I think? (At the very least the `else` is unneeded since the above `if` always returns)

ASpoonPlaysGames (changes requested):
Code mostly looks good, just a couple of things

NorthstarDLL/server/auth/serverauthentication.cpp
catornot: ok I'll remove it

ASpoonPlaysGames (approved):
Code looks good to me

Alystrasz (approved):
Code looks good, confirmed working in testing.

NorthstarDLL/server/auth/serverauthentication.cpp
```diff
@@ -229,51 +227,71 @@ void*,, (
...
+		return CBaseClient__Connect(self, pName, pNetChannel, bFakePlayer, a5, pDisconnectReason, a7);
+
+	// try to actually connect the bot
+	if (!CBaseClient__Connect(self, pName, pNetChannel, bFakePlayer, a5, pDisconnectReason, a7))
+		return false;
+
+	g_pServerAuthentication->AuthenticatePlayer(self, 0, "0");
 
-	g_pServerAuthentication->AddPlayer(self, pNextPlayerToken);
+	g_pServerAuthentication->AddPlayer(self, "0");
```
GeckoEidechse, line 294: I'm a bit confused here. Did we never properly use `pNextPlayerToken` or was it always init with `"0"`? At based on the diff and PR description I'm struggling a bit to understand this change ^^"

catornot: No, `pNextPlayerToken` was not initlizated, this just removes it by reworking when authentication happens.

GeckoEidechse:
Merging based on reviews.

GeckoEidechse:
Reverted in #610 as it made it trivial to crash a server by just running `connect IP_ADDRESS` as it somehow allowed player to bypass the auth check until they reached script where it would script error presumably due to missing pdata stuff.

AllusiveWheat:
I know this pr is old but I went ahead and looked into it since it would fix bot crashes. The function used for pushing a disconnect reason is just wrong and you should use `CBaseServer__RejectConnection`. Not sure why the server was crashing if you direct connected but I suppose that might be due to not rejecting the connection.

PR #549, opened 2023-09-14
Implement server draining

pg9182:
TODO: description

R2Northstar/Northstar#544


r3muxd:
there is absolutely no point in this; this is replicated by running the command sv_shutdown, which will quit the server when the server next enters hibernation (which will occur when the server is empty)
to prevent new players from joining, the "sv_rejectClientConnects" cvar can be used

GeckoEidechse:
> to prevent new players from joining, the `sv_rejectClientConnects` cvar can be used

Haven't gotten around to check yet whether that convar exists. Nevertheless, we still need to communicate to master server that gameserver is no longer accepting new players. Simply refusing connection on join gives a bad user experience.

The whole point of the PR is to give a mechanism to reduce server playercount to zero to allow for a restart without interrupting player experience. If done on a round end / map change, it should at least be announced to players that server will go down at the end of round.

Server draining (maybe even through an external trigger) is desired to allow for future updating mechanisms of servers that does not interrupt active sessions.

PR #550, opened 2023-09-14
Define indent size in editorconfig

GeckoEidechse:
Setting indent size in `.editorconfig` means that things like GitHub web preview will default to showing tab indent size as 4 spaces.

By default GitHub uses tabsize 8 for indentation making diffs look weird (cause `clang-format` is doing wrong things)
image

See also https://stackoverflow.com/questions/8833953/how-to-change-tab-size-on-github

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,3 +9,4 @@ charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
 indent_style = tab
+indent_size = 4


GeckoEidechse:
As a sidenote, I just found out that `clang-format` does tab indentation incorrectly, so enjoy this rant copy-pasted from Discord:

> holy fuck, not even clang-format does tab indentation correctly
> Like, as far as I have been told, the whole reason why some people (who are clearly wrong) prefer tabs is cause you can set your tab size to whatever you prefer be it, 2, 3 (omfg why?!), 4, 8 (spotted the old school C dev) etc
> and still have the code aligned
> right?
> 
> And yet, tab indentation is still done incorrectly
> Like look at this 
> image
> the correct way to indent lines 32 and 33 would be to use one tab (like line 31) and then do the rest with spaces so that the aligned stays correct with the line above which has characters on it. That way changing tab size does not fuck the formatting
> But guess what...
> image
> Even fucking `clang-format` used tabs here
> So for tab size 4 it's perfectly fine
> But the moment you change it
> 
> `tab size == 2`
> image
> `tab size == 8`
> image
> 
> boom, it's all fucked
> defeating the entire purpose of using tabs in the first place
> 
> Anyway, so this is why spaces are the only correct way of indentation. Cause these fuck-ups will never happen.

ASpoonPlaysGames:
not adding "needs testing" because this is just an editor thing

ASpoonPlaysGames (approved):
Code looks good.

On the note about clang-format doing tabs wrong, I found a post on reddit about this sort of thing and someone linked to somewhere saying that it worked for them, but that link is now dead :(

GeckoEidechse:
Given #556 I'm not sure if we still need/want this tbh

ASpoonPlaysGames:
Personally I'd still like this, if not for anything other than making github display tabs at not absurd sizes

Issue #553, opened 2023-09-16
`disconnect` command shouldn't be `FCVAR_SERVER_CAN_EXECUTE`


ASpoonPlaysGames:
We already have a function for disconnecting people properly, so we should use that in script instead.

This allows people to disconnect others via kill replays and such

offending line:
https://github.com/R2Northstar/NorthstarLauncher/blob/main/NorthstarDLL/shared/misccommands.cpp#L189C44-L189C44

F1F7Y:
On another note `disconnect` can just be ignored by the client so `CClient::Disconnect` ( or script equivalent `NSDisconnectPlayer` ) should always be used to make sure the client is actually disconnected

r3muxd:
> This allows people to disconnect others via kill replays and such

wrong

PR #554, opened 2023-09-16
Fix crash when running `disconnect %%s`

### LINKED ISSUE #552: Running `disconnect` with string format specifiers does weird things
Examples:
`disconnect %s` shows the following in the error dialog
image
`disconnect "%n"` causes this assert:
image

ASpoonPlaysGames:
Closes #552 

Basically we were running a string format and then passing the string to the original function, which would then try and format it again (and get bad args). By passing `"%s"` and then our formatted string we basically just protect the vanilla string format

### DIFF
--- a/NorthstarDLL/client/rejectconnectionfixes.cpp
+++ b/NorthstarDLL/client/rejectconnectionfixes.cpp
@@ -25,7 +25,7 @@ void,, (bool a1, const char* fmt, ...))
 		R2::Cbuf_AddText(R2::Cbuf_GetCurrentPlayer(), "disconnect", R2::cmd_source_t::kCommandSrcCode);
 	}
 
-	return COM_ExplainDisconnection(a1, buf);
+	return COM_ExplainDisconnection(a1, "%s", buf);
 }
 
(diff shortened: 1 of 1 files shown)


Mauler125:
Code looks good, should be merged as this will fix potential security problems from taking place in the engine.
A further analysis of the code base is recommended for uncontrolled format strings or similar code bugs (requires /Wall to see).

ASpoonPlaysGames:
Whoever merges this should add @Mauler125 as coauthor :)
(I forgot when making the commit)

F1F7Y (approved):
Works, changes look good.

Preferably #553 should be implemented

Alystrasz (approved):
Running `disconnect %%s` does not crash the client with this.

GeckoEidechse (approved):
Confirmed working in testing by running `disconnect %%s` with and without the PR (with prints `%s`, without crashes). No code review done.

GeckoEidechse:
> Whoever merges this should add @Mauler125 as coauthor :)
> (I forgot when making the commit)

Done o7

PR #558, opened 2023-09-26
Expose `Northstar.dll` version to squirrel

ASpoonPlaysGames:
I intend on using this to finally set the main menu version number thing based on actual northstar version. This has its own issues such as providing rc information, but I'm not sure what can be done about that since rc stuff isn't known in launcher afaik?

### Testing:
do
`script_ui printt( NS_VERSION_MAJOR, NS_VERSION_MINOR, NS_VERSION_PATCH, NS_VERSION_DEV )`
`script_client printt( NS_VERSION_MAJOR, NS_VERSION_MINOR, NS_VERSION_PATCH, NS_VERSION_DEV )`
`sv_cheats 1; script printt( NS_VERSION_MAJOR, NS_VERSION_MINOR, NS_VERSION_PATCH, NS_VERSION_DEV )`

you should get the same print for all of them, and it should look something like `0 0 0 1` (last number probably different because of CI)

### DIFF
--- a/NorthstarDLL/scripts/scriptutility.cpp
+++ b/NorthstarDLL/scripts/scriptutility.cpp
@@ -1,7 +1,6 @@
 #include "squirrel/squirrel.h"
 #include "client/r2client.h"
 #include "engine/r2engine.h"
-#include "ns_version.h"
 
 // asset function StringToAsset( string assetName )
 ADD_SQFUNC(
@@ -26,22 +25,4 @@ ADD_SQFUNC(
 	}
 
 	return SQRESULT_NULL;
-}
-
-ADD_SQFUNC(
-	"string",
-	GetNorthstarVersion,
-	"",
-	"Returns the current version string of Northstar(.dll)",
-	ScriptContext::UI | ScriptContext::CLIENT | ScriptContext::SERVER)
-{
-	constexpr int version[4] {NORTHSTAR_VERSION};
-	std::string versionStr = std::format("{}.{}.{}", version[0], version[1], version[2]);
-
-	// for dev builds, show the fourth version number and the +dev bit
-	if (version[3])
-		versionStr += std::format(".{}+dev", version[3]);
-
-	g_pSquirrel<context>->pushstring(sqvm, versionStr.c_str());
-	return SQRESULT_NOTNULL;
 }
--- a/NorthstarDLL/scripts/scriptutility.cpp
+++ b/NorthstarDLL/scripts/scriptutility.cpp
@@ -1,6 +1,7 @@
 #include "squirrel/squirrel.h"
 #include "client/r2client.h"
 #include "engine/r2engine.h"
+#include "ns_version.h"
 
 // asset function StringToAsset( string assetName )
 ADD_SQFUNC(
@@ -25,4 +26,22 @@ ADD_SQFUNC(
 	}
 
 	return SQRESULT_NULL;
+}
+
+ADD_SQFUNC(
+	"string",
+	GetNorthstarVersion,
+	"",
+	"Returns the current version string of Northstar(.dll)",
+	ScriptContext::UI | ScriptContext::CLIENT | ScriptContext::SERVER)
+{
+	constexpr int version[4] {NORTHSTAR_VERSION};
+	std::string versionStr = std::format("{}.{}.{}", version[0], version[1], version[2]);
+
+	// for dev builds, show the fourth version number and the +dev bit
+	if (version[3])
+		versionStr += std::format(".{}+dev", version[3]);
+
+	g_pSquirrel<context>->pushstring(sqvm, versionStr.c_str());
+	return SQRESULT_NOTNULL;
 }
--- a/NorthstarDLL/squirrel/squirrel.cpp
+++ b/NorthstarDLL/squirrel/squirrel.cpp
@@ -376,7 +376,7 @@ bool IsUIVM(ScriptContext context, HSquirrelVM* pSqvm)
 	return ScriptContext(pSqvm->sharedState->cSquirrelVM->vmContext) == ScriptContext::UI;
 }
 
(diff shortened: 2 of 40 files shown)


NorthstarDLL/scripts/scriptutility.cpp
```diff
@@ -26,3 +27,21 @@ ADD_SQFUNC(
 
 	return SQRESULT_NULL;
 }
+
+ADD_SQFUNC(
+	"string",
+	GetNorthstarVersion,
```
GeckoEidechse, line 33: Should we call the function `GetNorthstarLauncherVersion` or something alike to indicate that it's the version of the native code and not script? ^^"

ASpoonPlaysGames: I considered that, but `GetNorthstarLauncherVersion` implies that it is getting the version of `NorthstarLauncher.exe` which isn't the case.

I don't think that this would cause issues, given that getting script version is more verbose `NSGetModVersion` or whatever it is exactly.

For most users, the versions should all match up anyway, this is mostly meant for dev builds and such. The overarching goal for me here is to display more build info in squirrel, so that the exact build of clients can be determined from something as simple as a screenshot of the error dialogue

RoyalBlue1 (approved):
code looks good no obvious issues

H0L0theBard (approved):
Code looks good :+1: 

Works in testing 
image

GeckoEidechse:
Merging based on reviews

GeckoEidechse:
Actually, looking at this again, couldn't we pass the version number as custom struct or array instead? Or is there a reason to pass it as a string?

Cause then we could do further checks in Squirrel based on individual version numbers `MAJOR.MINOR.PATCH.BUILD` without having to parse the string back to numbers ^^

ASpoonPlaysGames:
> Actually, looking at this again, couldn't we pass the version number as custom struct or array instead? Or is there a reason to pass it as a string?

We could, but i don't see the point really. Do you have a use-case in mind?

GeckoEidechse:
> > Actually, looking at this again, couldn't we pass the version number as custom struct or array instead? Or is there a reason to pass it as a string?
> 
> We could, but i don't see the point really. Do you have a use-case in mind?

My first train of thought was comparing launcher and squirrel version and checking if they match (with some additional handling for dev builds and release candidates). On mismatch we would print error or show both in menu or something.

However if I'm not mistaken the version in Squirrel number is still a string atm, not sure if it gets parsed.

So main reason would be future proofing for the above use case once version number handling in Squirrel improves.

We could also merge current PR as is and the switch to sending as struct/array later but that would mean API change and break any mods that would since start to rely on this function returning a string ^^

ASpoonPlaysGames:
> We could also merge current PR as is and the switch to sending as struct/array later but that would mean API change and break any mods that would since start to rely on this function returning a string ^^

or we could just add another function that returns it as a struct or whatever in the future and not have that problem

GeckoEidechse:
> or we could just add another function that returns it as a struct or whatever in the future and not have that problem

That kinda sounds like duplicating logic tbh ^^"

Imma leave the call up to you, whether to adjust function now or merge as is ^^

GeckoEidechse:
So merge as is then? ^^

ASpoonPlaysGames:
> So merge as is then? ^^

Yeah

GeckoEidechse:
Aight, I'll merge as soon as I got `v1.19.9` released (cherry-picking commits to branch is pain compared to fast-forward merges ^^")

uniboi:
Why is this not a constant?

ASpoonPlaysGames:
> Why is this not a constant?

Idk, it could be i guess.

Making it into a set of constants could also deal with what gecko wanted as well, might rework pr to use that instead

NorthstarDLL/squirrel/squirrel.cpp
```diff
@@ -256,7 +257,16 @@ template <ScriptContext context> void SquirrelManager<context>::VMCreated(CSquir
 
 		defconst(m_pSQVM, pair.first.c_str(), bWasFound);
 	}
+
```
ASpoonPlaysGames, line 260: no, this was not an accidental change. I am splitting up the const definitions for readability

uniboi (approved):
did not test but code looks very cool and good
should allow things like this in scripts
```
#if NS_VERSION_MAJOR == 1
print("among us")
#endif
```

ASpoonPlaysGames:
Testing instructions have been updated btw

EM4Volts:
tested. works.

taskinoz:
It would probably be good to also pass the full build number as NS_VERSION in case mods don't want to check against each category of the build number

GeckoEidechse:
> It would probably be good to also pass the full build number as NS_VERSION in case mods don't want to check against each category of the build number

Don't think this is necessary. It's just a

```js
if (
  NS_VERSION_MAJOR == 1 &&
  NS_VERSION_MINOR == 2 &&
  NS_VERSION_PATCH ==  3
) {
  // Do something here
}
```

And most likely the mod doesn't want to check against a specific patch version. If they do it's so that base Northstar contains a certain fix and then you wanna check if `NS_VERSION_PATCH >= 456` which doesn't work when passing the whole number as a string anyway.

The only reason I could think of to pass the whole thing is so that you can do

```
printt( NS_VERSION )
```
vs
```
printt( NS_VERSION_MAJOR, NS_VERSION_MINOR, NS_VERSION_PATCH, NS_VERSION_DEV )
```
but tbh that's not kinda worth exposing another variable IMO.

ASpoonPlaysGames:
> but tbh that's not kinda worth exposing another variable IMO.

Yeah, just define it in script tbh

uniboi:
shocking `format( "%d.%d.%d+%d",  NS_VERSION_MAJOR, NS_VERSION_MINOR, NS_VERSION_PATH, NS_VERSION_DEV)`

GeckoEidechse:
Merging based on reviews and discussions.

PR #561, opened 2023-10-03
Revert "Add a safeguard to map command"

GeckoEidechse:
#529 breaks Docker image as per https://github.com/pg9182/northstar-dedicated/issues/70

As such we should revert the change until we can figure out the reason for breakage.


Jan200101 (approved):
Can confirm, does revert.

PR #562, opened 2023-10-03
add flag to disable crash handler

Jan200101:
Skips Crash Handler init when `-disablecrashhandler` is passed

Tier0::CommandLine cannot be used because we are too early for the hook.

On Linux the crash handler has a tendency to loop, shitting out dmp's over and over again.
Debuggers, including winedbg, don't make `IsDebuggerPresent` return true.

### DIFF
--- a/NorthstarDLL/logging/crashhandler.cpp
+++ b/NorthstarDLL/logging/crashhandler.cpp
@@ -366,6 +366,9 @@ BOOL WINAPI ConsoleHandlerRoutine(DWORD eventCode)
 
 void InitialiseCrashHandler()
 {
+	if (strstr(GetCommandLineA(), "-disablecrashhandler"))
+		return;
+
 	hExceptionFilter = AddVectoredExceptionHandler(TRUE, ExceptionFilter);
 	SetConsoleCtrlHandler(ConsoleHandlerRoutine, true);
 	std::set_terminate(RuntimeExceptionHandler);


Jan200101:
Further testing has shown that the crash handler was also the main reason why Northstar couldn't get in-game on Linux prior to Proton 8.

For debugging purposes this flag is useful even with future reworks.

ASpoonPlaysGames:
> On Linux the crash handler has a tendency to loop, shitting out dmp's over and over again.
Would prefer #477 tbh

> Debuggers, including winedbg, don't make IsDebuggerPresent return true.
Why not solve this problem instead of doing this?

> For debugging purposes this flag is useful even with future reworks.
How?

F1F7Y:
> On Linux the crash handler has a tendency to loop, shitting out dmp's over and over again.

As it has been over 3 months since i re-worked the crash handler i don't remember what exactly caused this. But this isn't a linux only issue.

Jan200101:
> > On Linux the crash handler has a tendency to loop, shitting out dmp's over and over again.
> 
> Would prefer #477 tbh
> As it has been over 3 months since i re-worked the crash handler i don't remember what exactly caused this. But this isn't a linux only issue.

Now that you mention it, I do recall running into this while working on a Plugin on Windows.
Its likely that this behavior is not Linux specific but more common on Linux due to wine.

> > Debuggers, including winedbg, don't make IsDebuggerPresent return true.
> 
> Why not solve this problem instead of doing this?

How would we solve it?
Wine seemingly does so intentionally so there is nothing to upstream.
We could do it like Valve does any attempt to leave the wine prefix and check if the current process is being ptraced


> > For debugging purposes this flag is useful even with future reworks.
> 
> How?

You may want to attach instrumentation that does not set the BeingDebugged flag in the proceess
but not deal with the termination handler and so on.

ASpoonPlaysGames:
> How would we solve it?
> Wine seemingly does so intentionally so there is nothing to upstream.

Not particularly sure, but from the winedbg documentation, "Wine does not report a debugger by default" which to me implies that you can make it report a debugger somehow. Not my domain really but yeah

>You may want to attach instrumentation that does not set the BeingDebugged flag in the proceess
but not deal with the termination handler and so on.

Yeah sure, fair enough

ASpoonPlaysGames (approved):
Code looks good

`strstr` is not ideal but we don't have a nicer alternative atm afaik. Also with a long flag like this one, it is unlikely to cause issues in the future

GeckoEidechse:
merge conflicts on this atm btw

Jan200101:
yeah idk, I don't really care about having this anymore.

PR #563, opened 2023-10-03
Give `g_pMemAllocSingleton` a default value

Jan200101:
The standard does not guarantee that `g_pMemAllocSingleton` will default to a falsey value, which our `*alloc` implementations demands.

I believe this change is so small and contained that its not worth testing, can you agree with that @GeckoEidechse ?

### DIFF
--- a/NorthstarDLL/core/tier0.cpp
+++ b/NorthstarDLL/core/tier0.cpp
@@ -3,7 +3,7 @@
 // use the Tier0 namespace for tier0 funcs
 namespace Tier0
 {
-	IMemAlloc* g_pMemAllocSingleton;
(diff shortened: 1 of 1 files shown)


GeckoEidechse:
Depends if/how this could be tested. Although given that I'm pushing out many smaller releases atm I might be tempted to just test in prod (i.e. merge and release) :>

Jan200101:
This can't really be tested
In theory this could cause a crash but on the intended platform (x86_64 Windows) its not a problem.
This is rather for correctness of the code.

GeckoEidechse:
Yeah, then we just skip testing. Imma just try to get someone to sign-off the code review and then I'll merge it ^^

RoyalBlue1 (approved):
Yea code looks good

PR #564, opened 2023-10-03
Add a safeguard to map command

GeckoEidechse:
Adds safeguard to the `map` command that prevents it from executing if the requested map is invalid or no map argument is given.

This is basically #529 again which got reverted in #561 due to https://github.com/pg9182/northstar-dedicated/issues/70

The issue in https://github.com/pg9182/northstar-dedicated/issues/70 is likely caused by Wine and not launcher code. Nevertheless the PR got reverted due to not being able to locate the issue in Wine in a timely manner.

As such this PR here is marked as draft until we can figure out and solve the issue in Wine in the Docker image.


GeckoEidechse:
cc @catornot

Jan200101:
Figured it out (shoutouts to x64dbg) @GeckoEidechse 

<details>
<summary>Stack Trace and Registers</summary>

image
image
</details>

Its crashing right here
https://github.com/R2Northstar/NorthstarLauncher/blob/024605399a4878f97fd789416d9ea4234241f039/NorthstarDLL/util/printmaps.cpp#L106-L116

because in the Docker container there is no `r2/maps` it throws an exception which goes uncaught.
This was not caught on Windows because it was done on a full install with the maps folder intact.

Fixes are:
- putting it in a `try ... catch (fs::filesystem_error)`
- checking if the folder exists before iterating
- use the `std::error_code` overload of `directory_iterator`

GeckoEidechse:
That's amazing, many thanks for taking your time to look into this <3

cc @catornot

GeckoEidechse:
With #575 merged and this PR updated accordingly, this should work now.

GeckoEidechse:
Test with
```
ghcr.io/pg9182/northstar-dedicated-ci:1-tf2.0.11.0-ns1.19.8-ci.c14bdc62.112.6533697962-nslauncher.0d07f4b2.6526237893
```
created via https://github.com/pg9182/northstar-dedicated-ci/actions/runs/6533697962

catornot (approved):
still works locally
image
👍

PR #565, opened 2023-10-04
Write date to logs

### LINKED ISSUE #560: Write date to logs
As of right now, log lines only contain the time of the log, the date is inferred from the filename.

`nslog2023-08-18 22-07-19.txt`
```
[22:07:19] [NORTHSTAR] [info] NorthstarLauncher version: 1.18.1.0
[22:07:19] [NORTHSTAR] [info] Command line: NorthstarLauncher.exe
[22:07:19] [NORTHSTAR] [info] Using profile: R2Northstar
[22:07:19] [NORTHSTAR] [info] Operating System: Windows (NT10.0)
[22:07:19] [NORTHSTAR] [info] Enabling hook _GetCommandLineA
[22:07:19] [NORTHSTAR] [info] Enabling hook _LoadLibraryExA
...
```

This is perfectly fine for clients where play sessions are expected to only last a few hours at a time.

However for servers which typically run for days or weeks, this might become a bit of an issue when trying to find out when a particular event happened. While it would possible to manually keep track of new days by observing when the time switches from `23:59` to `00:00`, this quite cumbersome. For this reason we should consider writing date to log line as well.

(Suggestion provided by @Vexbane via Discord DMs)

H0L0theBard:
Suggestion from #560 

This PR adds dates to logs in the same format which the file name uses. 

image

Closes #560

### DIFF
--- a/NorthstarDLL/logging/logging.cpp
+++ b/NorthstarDLL/logging/logging.cpp
@@ -52,7 +52,7 @@ void CreateLogFiles()
 
 			stream << std::put_time(&currentTime, (GetNorthstarPrefix() + "/logs/nslog%Y-%m-%d %H-%M-%S.txt").c_str());
 			auto sink = std::make_shared<spdlog::sinks::basic_file_sink_mt>(stream.str(), false);
-			sink->set_pattern("[%H:%M:%S] [%n] [%l] %v");
+			sink->set_pattern("[%Y-%m-%d] [%H:%M:%S] [%n] [%l] %v");
 			for (auto& logger : loggers)
 			{
 				logger->sinks().push_back(sink);


Asha097:
Tested this and can confirm its not holowarebitcoinminer.exe
A-okay chief

Jan200101 (approved):
Code looks good, date format uses the long variant 👍

Alystrasz (approved):
Code looks good, testing shows date now appears in log files.

ASpoonPlaysGames:
This only affects the log files (and not things like the in-game console) right?

H0L0theBard:
This only shows in logs.

GeckoEidechse:
One trade-off is that this change will increase log size due to preventing more characters for each line. Personally I think this trade-off is worth it.
It might also break existing log parsers but that's not our problem.

If there's no opposition to this PR I'll merge it within the next few hours or so.

ASpoonPlaysGames:
> One trade-off is that this change will increase log size

Would it be worth restricting this to only dedicated servers? Listen servers and clients shouldn't really be having the game open for multiple days at a time

H0L0theBard:
Users can sometimes send their logs as a `message.txt` in the Northstar discord so having the date in the log can help in verifying if it is actually a recent log. 

That's basically the only idea I have in why we'd have it for clients and listen servers. We also recently reduced the spam of some lines in a couple of PRs so I don't personally see any issues with adding a little bit extra onto each line.

itscynxx:
> Users can sometimes send their logs as a `message.txt` in the Northstar discord so having the date in the log can help in verifying if it is actually a recent log. 

as an alternative if this isn't "enough" reason, I can have spectre check `message.txt` files, check if they're northstar logs, and ask for an actual log upload 

imo it makes less sense to do it this way (spectre way), and instead just keep date in client logs, but it's an alternative

ASpoonPlaysGames:
> Users can sometimes send their logs as a `message.txt` in the Northstar discord so having the date in the log can help in verifying if it is actually a recent log. 

Fair enough, good enough for me

GeckoEidechse:
Imma count that whole discussion as primarily in favour of having date in logs on both server and client and merge as is ^^

PR #566, opened 2023-10-05
Fix SERVER | CLIENT context specifier error

EladNLG:
see title.

### DIFF
--- a/NorthstarDLL/squirrel/squirrelautobind.h
+++ b/NorthstarDLL/squirrel/squirrelautobind.h
@@ -21,16 +21,16 @@ class __squirrelautobind;
 		__squirrelautobind CONCAT2(__squirrelautobind, __LINE__)(                                                                          \
 			[]()                                                                                                                           \
 			{                                                                                                                              \
-				if constexpr (runOnContext & ScriptContext::UI)                                                                            \
+				if constexpr ((runOnContext) & ScriptContext::UI)                                                                            \
 					g_pSquirrel<ScriptContext::UI>->AddFuncRegistration(                                                                   \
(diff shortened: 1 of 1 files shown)


NorthstarDLL/squirrel/squirrelautobind.h
```diff
@@ -21,16 +21,16 @@ class __squirrelautobind;
 		__squirrelautobind CONCAT2(__squirrelautobind, __LINE__)(                                                                          \
 			[]()                                                                                                                           \
 			{                                                                                                                              \
-				if constexpr (runOnContext & ScriptContext::UI)                                                                            \
+				if constexpr ((runOnContext) & ScriptContext::UI)                                                                            \
```
GeckoEidechse, line 24: Ok, this is hitting the limits of my C++ knowledge. Why does adding (seemingly unnecessary) brackets fix the issue? ^^"

GeckoEidechse (commented):
heads-up that there are format errors rn

NorthstarDLL/squirrel/squirrelautobind.h
EladNLG: because of how macros work

runOnContext is replaced witb an expression, for example:
`ScriptContext::SERVER | ScriptContext::CLIENT`.

`ScriptContext::SERVER | ScriptContext::CLIENT & ScriptContext::UI` causes a compile error because it doesn't know how to combine it.

Jan200101 (approved):
code looks good
`runOnContext` should have been isolated in the first place to ensure order of operations.

GeckoEidechse:
uh, any suggestions on how I best test this? ^^

uniboi:
This really does not need to be tested. The macro arguments are expanded so you have to make sure they're evaluated in the correct order or it might change. That's it

PR #568, opened 2023-10-11
Update libcurl to 8.4.0

Jan200101:
Fixes:
- https://curl.se/docs/CVE-2023-38545.html
- https://curl.se/docs/CVE-2023-38546.html


GeckoEidechse (approved):
Seems to work fine on client. I was able to connect to master server and join a server just fine. Editing loadouts and sending the resulting pdata to masterserver was also fine.

Windows dedi was also able to register with master server just fine.

Only thing not tested is Docker dedi but assuming newer curl version doesn't trigger a Wine bug I think we should be fine. It's getting quite late here so I'm skipping testing Docker dedi :P

PR #569, opened 2023-10-11
Correct submodule detection

Jan200101:
the contents of the submodules were incorrectly counted resulting in a git submodule update on every cmake invocation

This can be easily seen by checking out a different commit in a dependency and trying to build NS with it.

PR #570, opened 2023-10-11
Fixed mixed indentation in utils.cmake

GeckoEidechse:
The file had mixed indentation (spaces and tabs) when really it should've been spaces only from the beginning

I noticed the indentation change in #569 and spliced it out to make the diff of #569 proper on merge.

No testing needed as this only performs a whitespace change in `utils.cmake`. Check the diff with whitespace changes hidden to verify for yourself :P


Jan200101 (approved):
just indention change :+1: 
Looks gucci

PR #574, opened 2023-10-14
Revert "Update libcurl to 8.4.0 (#568)"

Jan200101:
newer releases of libcurl break the docker container
Its been triaged to only occur within the container so it can be argued that its not a general wine issue.

Since the CVE 8.4.0 fixes does not affect us, even if someone sets the environment variable, its fine to downgrade until the container situation is fixed or migrated

TODO @Jan200101 recreate PR once this is merged
 
This reverts commit 99c5c86e188e090c93a9778da3402d62409ece8b.


Jan200101:
@GeckoEidechse since this is a revert, is there any testing needed?

GeckoEidechse:
> @GeckoEidechse since this is a revert, is there any testing needed?

nope, should not be necessary

GeckoEidechse:
well actually maybe we should double check that docker image actually runs with this tbh

GeckoEidechse:
Image for testing
```
ghcr.io/pg9182/northstar-dedicated-ci:1-tf2.0.11.0-ns1.19.7-ci.c14bdc62.111.6520969261-nslauncher.56267614.6520921720
```

GeckoEidechse (approved):
Confirmed working in testing. Docker image runs again

PR #575, opened 2023-10-14
Add check for r2/maps folder

H0L0theBard:
Adds a fix so that #564 can be re-merged

# Testing Instructions
1. Rename r2/maps  
2. Run `maps *` in the console 

If the game does not die, :tada:. 
It should just print the maps found from the vpks.

### DIFF
--- a/NorthstarDLL/util/printmaps.cpp
+++ b/NorthstarDLL/util/printmaps.cpp
@@ -104,6 +104,12 @@ void RefreshMapList()
 	}
 
 	// get maps in game dir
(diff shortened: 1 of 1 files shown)


Jan200101 (approved):
confirmed working
code is clean and simple, checks if the folder exists.

itscynxx:
It does in fact work properly in testing
I tested this of my own free will and there were no other factors in doing so

GeckoEidechse:
Merging based on reviews.

PR #576, opened 2023-10-14
Don't run add-to-project on forks

GeckoEidechse:
As they don't have access to the necessary token


GeckoEidechse:
Looks like `if: github.event.repository.fork == false` is not actually a thing that works. RIP.

PR #577, opened 2023-10-14
Only run `add-to-project` action on issues

GeckoEidechse:
PRs from forks are triggered from that fork which means that they don't have access to the necessary token causing them to fail for the first commit.


RoyalBlue1 (approved):
Simply signing this PR off so it can be merged.

PR #579, opened 2023-10-15
Add squirrel function GetPlayerStatusEffects

EladNLG:
Requires mods PR [didnt make it yet], https://github.com/R2Northstar/NorthstarLauncher/pull/566

`array<StatusEffectData> function GetPlayerStatusEffects( entity player )`

<hr> 

**Returns:** an array which holds all status effects that a player has.

**NOTE:** the array may contain status effects which have expired; check if the end time is smaller than the current time.


EladNLG:
side note, CI for `build` fails because #566 isn't merged

GeckoEidechse:
@EladNLG #566 got merged but I cannot cleanly update this PR here to newest main cause merge conflicts. Can you update this PR so that we can take another look? ^^

GeckoEidechse:
What's the reason this was closed?

PR #581, opened 2023-10-15
Move plugin communication handler init out of plugin manager init

### LINKED ISSUE #580: `-noplugins` crashes the game on startup
image
screenshot of the exception that is thrown

ASpoonPlaysGames:
Plugin state requires this to be initialised even if `-noplugins` is passed, but if `-noplugins` is passed, it doesn't get initialised.

Moving this to `dllmain` guarantees that it is initialised properly and prevents #580 

closes #580


Jan200101 (approved):
Tested and it works without a hitch.

Code is clean and moves the global init into the InitNS function, like many other things are.

GeckoEidechse:
Merging based on reviews.

PR #582, opened 2023-10-19
add flag to disable in-game overlay

Jan200101:
The Origin / EA Desktop in-game overlay is known to be quite shit.

Here are a few sources that talk about it and EA's unwillingness to fix it (at least in the past)
- https://www.reddit.com/r/origin/comments/ho3ed5/overlay_wont_go_away/
- https://steamcommunity.com/sharedfiles/filedetails/?id=2664686758
- https://www.reddit.com/r/origin/comments/h7qoni/origin_overlay_is_permanently_enabled_for_games/
(the last one is important because ea games launched over steam are put into a special Link2Ea state)

Code is simple:
- after hook dispatch check if `-disable_igo` is on the command line and store if it was
- if igo was disabled check if the `lpLibFileName` ends with `L"IGO64.dll"`
  - if it does return nullptr

### DIFF
--- a/NorthstarDLL/core/hooks.cpp
+++ b/NorthstarDLL/core/hooks.cpp
@@ -451,10 +451,10 @@ HMODULE, WINAPI, (LPCWSTR lpLibFileName))
 {
 	if (disable_igo)
 	{
-		LPCWSTR LibFileNameEnd = lpLibFileName + wcslen(lpLibFileName);
-		LPCWSTR LibName = LibFileNameEnd - wcslen(IGO_DLL);
+		LPCWSTR lpLibFileNameEnd = lpLibFileName + wcslen(lpLibFileName);
+		LPCWSTR lpLibName = lpLibFileNameEnd - wcslen(IGO_DLL);
 
-		if (!wcsncmp(LibName, IGO_DLL, wcslen(IGO_DLL) + sizeof(*LPCWSTR)))
+		if (!wcsncmp(lpLibName, IGO_DLL, wcslen(IGO_DLL) + sizeof(wchar_t)))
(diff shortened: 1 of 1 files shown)


NorthstarDLL/core/hooks.cpp
```diff
@@ -9,6 +9,10 @@
 #include <filesystem>
 #include <Psapi.h>
 
+#define IGO_DLL L"IGO64.dll"
+
+bool disable_igo = false;
```
ASpoonPlaysGames, line 14: Might be better to just use a static bool inside the function scope for this?
also camelCase please :(

NorthstarDLL/core/hooks.cpp
```diff
@@ -459,4 +472,6 @@ void InstallInitialHooks()
 		spdlog::error("MH_Initialize (minhook initialization) failed");
 
 	AUTOHOOK_DISPATCH()
+
+	disable_igo = strstr(GetCommandLineA(), "-disable_igo");
```
ASpoonPlaysGames, line 476: as discussed on discord a bit, I'm not a fan of `-disable_igo`.
An alternative could be `-nogameoverlay` or something?

Anyway yeah use this comment as a discussion point for the name or something i guess

Jan200101:
Further testing has revelead that `IGO64.dll` is only loaded at runtime if its launched outside of Origin/EA Desktop.

When Launching thruogh Origin/EA Desktop or Steam (since it uses link2ea://) it gets loaded ahead of time.

Unloading the library then causes a dangling pointer somewhere in the internals resulting in a hard crash.
Because I don't see a solution, I'm closing this.

PR #584, opened 2023-10-23
Run add-to-project on new PRs

ASpoonPlaysGames:
Same as https://github.com/R2Northstar/NorthstarMods/pull/753 which seems to be working just fine


GeckoEidechse (approved):
Same as https://github.com/R2Northstar/NorthstarMods/pull/753 which seems to be working fine so not much to test/review here ^^

PR #585, opened 2023-10-25
Unify entrypoints (Launcher, wsock32 proxy)

Jan200101:
Merges the wsock and launcher directories together.

Full changes:
- separated code that was common to loader.cpp
- fail conditions were moved into relevant functions
- put common code into a NorthstarLoader library to be reused


Jan200101:
rebased, fixed some things up, tested it myself (though I want someone else to test it too)

NorthstarLauncher/dllmain.cpp
```diff
@@ -3,48 +3,25 @@
...
-		}
-
-		SetCurrentDirectoryW(exePath);
+		SetCurrentDirectoryW(GetExePathWide());
 
 		if (!ProvisionNorthstar()) // does not call InitialiseNorthstar yet, will do it on LauncherMain hook
 			return true;
 
+		wchar_t buffer1[4096 * 2];
+		wchar_t buffer2[4096 * 3];
```
p0358, line 21: Hm, Visual Studio/MSVC will warn you if your stack size in a single function exceeds 16384. Technically this function doesn't go deep in chain and our stack size is big enough. At the same time I get it's an improvement over wasting heap space forever I guess, but it's really not ideal either way.

If we don't care about heap space waste, then it'd be the easiest to just change them to static probably.

At the same time, the original code was written to avoid any allocations. That constraint was already violated in several places, and wasn't really probably actually really needed to begin with, so the whole code here could probably be refactored to use std::wstrings and stuff perhaps?

If you include `<fmt/xchar.h>` from spdlog's fmt (if it's not already), you can probably also use `fmt::format(L"{} etc", ...)`

p0358: Well, you moved the commented lines from heap to stack. If you're okay with the compiler warning as a result of that, fine I guess...?

Jan200101: right, completely forgot about that.
I mean, it *shouldn't* cause issues, but it would be better to just move it to be static.

Jan200101: thought about it
best to just turn static for now and implement a more proper solution later.

Also went and turned some other globals static, to ensure they don't leak across translation units for some reason.

Jan200101:
remade from scratch
split things up into fewer segments so its easier to digest

PR #590, opened 2023-11-03
Load plugin dependencies from lib folder

RoyalBlue1:
Disables recursive search for plugins in plugin folders and if it exists adds lib folder within plugin folder to Dll load dirs


catornot (approved):
works well, only destroyed one of my plugins because it probably unsound before but it wasn't important :)

code looks good

GeckoEidechse:
Merging based on review

PR #472, opened 2023-06-18
Plugins v3

catornot:
yes

 
## what changed
- nuked presence ( will be move to the discord rpc plugin )
- more exposed sq functions
- exposed dll addresses
- `g_pCVar` is exposed
- added "userdata" to plugin's async call
- added runframe to plugins

discord rpc pr : https://github.com/R2Northstar/NorthstarDiscordRPC/pull/10
mods pr : https://github.com/R2Northstar/NorthstarMods/pull/652
test plugins : https://github.com/R2Northstar/NorthstarLauncher/pull/472#issuecomment-1661401264

PR #591, opened 2023-11-05
Plugin related log error expecting more args

GeckoEidechse:
The related line expected 3 args but only 2 args are passed.

Blocks:

- #472 

### Testing:

The change should be obvious and not need testing. If you still wanna though. Simply replace the current DiscordRPC plugin we ship with the v1 version (grab from a release pre `v1.13`).

Before this PR you'll see something along the lines of "LOG ERROR" in your console output where plugin load happens.
With this PR it should log the error just fine.


pg9182 (approved):
Is clearly correct.

PR #593, opened 2023-11-05
refactor: Re-order stuff

GeckoEidechse:
based on plugins v3

### DIFF
--- a/NorthstarDLL/plugins/plugin_abi.h
+++ b/NorthstarDLL/plugins/plugin_abi.h
@@ -48,10 +48,6 @@ struct SquirrelFunctions
 	sq_pushassetType __sq_pushasset;
 	sq_pushvectorType __sq_pushvector;
 	sq_pushobjectType __sq_pushobject;
-	sq_getthisentityType __sq_getthisentity;
-	sq_getobjectType __sq_getobject;
-
-	sq_stackinfosType __sq_stackinfos;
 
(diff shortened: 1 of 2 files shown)


NorthstarDLL/plugins/plugin_abi.h
```diff
@@ -61,6 +57,10 @@ struct SquirrelFunctions
 	sq_getassetType __sq_getasset;
 	sq_getuserdataType __sq_getuserdata;
 	sq_getvectorType __sq_getvector;
+	sq_getthisentityType __sq_getthisentity;
+	sq_getobjectType __sq_getobject;
+
+	sq_stackinfosType __sq_stackinfos;
```
catornot, line 63: This is a breaking change therefore this pr should also increment the plugin version or not get released without plugins v3.

catornot (changes requested):

NorthstarDLL/plugins/plugin_abi.h
GeckoEidechse: Oh wait this struct is accessed by plugins, right? Yeah then it makes no sense to splice out, lol.

I just made the PR on the side while going through the diff ^^

GeckoEidechse:
Close as per https://github.com/R2Northstar/NorthstarLauncher/pull/593#discussion_r1382657441

PR #601, opened 2023-11-18
Rework `-vanilla` to be a vanilla compatibility mode

### LINKED ISSUE #588: Rework `-vanilla` launch option to be a vanilla compatibility mode
Currently, `-vanilla` is a pretty useless launch option that nobody uses. It basically is used by `NorthstarLauncher.exe` to launch the vanilla game without loading `Northstar.dll`

`-norestrictservercommands` is the current way of being "vanilla compatible" but is missing some things, so this would become a replacement for `-norestrictservercommands` being used for this

ASpoonPlaysGames:
Old `-vanilla` behaviour is now handled by `-nonorthstardll`

`VANILLA` is exposed as a squirrel constant, set to true when in vanilla compatibility mode.

Differences when in vanilla compatibility mode:
- Doesn't restrict server commands (same as `-norestrictservercommands`)
- Doesn't block FairFight screenshot functions 
- Doesn't do Atlas-related stuff (except for mainmenupromos)

With this PR various script changes will still be needed for proper vanilla compatibility, but it's a good start.

closes #588

### DIFF
--- a/NorthstarDLL/core/vanilla.h
+++ b/NorthstarDLL/core/vanilla.h
@@ -4,4 +4,26 @@
 /// In this mode we shouldn't auth with Atlas, which prevents users from joining a
 /// non-trusted server. This means that we can unrestrict client/server commands
 /// as well as various other small changes for compatibility
-static bool g_bIsVanillaCompatible = false;
+class VanillaCompatibility
+{
+  public:
+	void SetVanillaCompatibility(bool isVanilla)
+	{
+		static bool bInitialised = false;
+		if (bInitialised)
+			return;
+
+		bInitialised = true;
+		m_bIsVanillaCompatible = isVanilla;
+	}
+
+	bool GetVanillaCompatibility()
+	{
+		return m_bIsVanillaCompatible;
+	}
+
+  private:
+	bool m_bIsVanillaCompatible = false;
+};
+
+inline VanillaCompatibility* g_pVanillaCompatibility;
--- a/NorthstarDLL/core/vanilla.h
+++ b/NorthstarDLL/core/vanilla.h
@@ -1,0 +1,7 @@
+#pragma once
+
+/// Determines if we are in vanilla-compatibility mode.
+/// In this mode we shouldn't auth with Atlas, which prevents users from joining a
+/// non-trusted server. This means that we can unrestrict client/server commands
+/// as well as various other small changes for compatibility
+static bool g_bIsVanillaCompatible = false;
--- a/NorthstarDLL/client/clientauthhooks.cpp
+++ b/NorthstarDLL/client/clientauthhooks.cpp
@@ -19,7 +19,7 @@ void, __fastcall, (void* a1))
 {
 	// don't attempt to do Atlas auth if we are in vanilla compatibility mode
 	// this prevents users from joining untrustworthy servers (unless they use a concommand or something)
-	if (g_bIsVanillaCompatible)
+	if (g_pVanillaCompatibility->GetVanillaCompatibility())
 	{
 		AuthWithStryder(a1);
 		return;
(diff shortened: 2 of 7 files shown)


itscynxx:
Are the [auto compiled files](https://github.com/R2Northstar/NorthstarLauncher/actions/runs/6915829330?pr=601) not a good way to test this pr?

Using both `-vanilla` and `-noworkerdll` are both just launching regular vanilla on my end

ASpoonPlaysGames:
> Are the [auto compiled files](https://github.com/R2Northstar/NorthstarLauncher/actions/runs/6915829330?pr=601) not a good way to test this pr?
> 
> Using both `-vanilla` and `-noworkerdll` are both just launching regular vanilla on my end

It should work, make sure you grab both the exe and the dll since this PR touches both

itscynxx:
> It should work, make sure you grab both the exe and the dll since this PR touches both

Even going from Northstar -> Delete exe and dll -> vanilla -> install exe and dll from builds, still not seeing anything with either launch option applied

Maybe its just a "womm" moment?

ASpoonPlaysGames:
> Using both `-vanilla` and `-noworkerdll` are both just launching regular vanilla on my end

Its weird that you are getting both doing the same thing, if `-noworkerdll` launches vanilla then even if `-vanilla` doesnt work how I want it to, the worst it could do is launch normal northstar...

How are you launching Northstar?

itscynxx:
> To clarify:
> 
>     * when trying to test `-noworkerdll` you should probably be launching through `NorthstarLauncher.exe`
> 
>     * when trying to test `-vanilla` you will still need `-northstar` as well if launching through steam/ea/whatever

Wasn't aware `-northstar` was required as it wasn't required with `-vanilla` previously

`-northstar -vanilla` results in Northstar being launched with no main menu promos, with no `Launch Multiplayer` button (though this is handled in mods, so it makes sense). Trying to hit `Launch Northstar` results in a blank error message

It seems to be doing _something_, however I'm not sure to what degree it's meant to work without a mods pr

`-northstar -nonorthstardll` via Steam's launch options resulted in regular Northstar being launched with core mods and functional, just `-nonorthstardll` resulted in regular vanilla

`-nonorthstardll` passed to the `.exe` via a `.bat` file resulted in vanilla launching with the Northstar logging in a command prompt, which I'm pretty sure is the desired result

ASpoonPlaysGames:
> `-northstar -vanilla` results in Northstar being launched with no main menu promos, with no `Launch Multiplayer` button (though this is handled in mods, so it makes sense). Trying to hit `Launch Northstar` results in a blank error message
> 
> It seems to be doing _something_, however I'm not sure to what degree it's meant to work without a mods pr

This is the intended result of this PR, yes

GeckoEidechse:
I'm a bit hesitant of changing the behaviour of an existing command as we have a lot of baggage of old tutorials that would need updating etc.

Would it make sense to use something like `-vanillacompat` for Northstar in vanilla compatible mode? And then after change `-vanilla` to `-nonorthstardll` and put a deprecation notice on `-vanilla`? Or is that too overkill?

ASpoonPlaysGames:
> I'm a bit hesitant of changing the behaviour of an existing command as we have a lot of baggage of old tutorials that would need updating etc.

Literally nobody in the history of Northstar has used `-vanilla` intentionally for its old purpose. Definitely overkill

GeckoEidechse:
Aight, if you say this won't have a breaking impact, I'm fine with doing it in the current way ^^

NorthstarDLL/core/vanilla.h
```diff
@@ -0,0 +1,29 @@
+#pragma once
+
+/// Determines if we are in vanilla-compatibility mode.
+/// In this mode we shouldn't auth with Atlas, which prevents users from joining a
+/// non-trusted server. This means that we can unrestrict client/server commands
+/// as well as various other small changes for compatibility
+class VanillaCompatibility
```
uniboi, line 7: game dev try to not make everything a class challenge

ASpoonPlaysGames: I did this for two reasons:
1. Everything else does it
2. I want it only to be set once

itscynxx:
It should also be noted that this does not break VanillaPlus, in fact makes it possibly more straightforward for the end user installing it. Additionally, VanillaPlus can instead just use the `VANILLA` constant, not requiring having dependency checks for core mods, or doing individual checks for the core mods themselves

I have a 2.4 version of VanillaPlus using the `VANILLA` constant ready for when this gets released as well

GeckoEidechse:
Removing the testing label as @itscynxx already tested.

GeckoEidechse (approved):
The code changes look fine from what I can tell, although I have to say that I only looked at what shows up in the diff, meaning I will have missed anything that was forgotten to be updated.

PR #603, opened 2023-11-19
Replace audio event fetching with hook

Jan200101:
Takes the previous audio event code, which relied on reading out a register using masm, and replaces it with a new hook.

Adapted from NorthstarPrime https://github.com/F1F7Y/NorthstarPrime

### DIFF
--- a/NorthstarDLL/CMakeLists.txt
+++ b/NorthstarDLL/CMakeLists.txt
@@ -148,7 +148,6 @@ add_library(NorthstarDLL SHARED
             "util/version.h"
             "util/wininfo.cpp"
             "util/wininfo.h"
-            "audio_asm.asm"
             "dllmain.cpp"
             "dllmain.h"
             "ns_version.h"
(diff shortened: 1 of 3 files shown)


GeckoEidechse:
What's a good way to test this? Run game with mod that replaces audio?

GeckoEidechse (commented):
Seems to work in testing (i.e. no regression) when used together with [cooldudepugs-how_is_prangent_formed-1.0.1](https://northstar.thunderstore.io/package/cooldudepugs/how_is_prangent_formed/).

(Not marking as approving PR until I got confirmation that this is the right way to test ^^")

GeckoEidechse (approved):
Ok, I did test correctly in https://github.com/R2Northstar/NorthstarLauncher/pull/603#pullrequestreview-1740896655

F1F7Y (approved):
Changes look good, taken from: [sample.cpp](https://github.com/F1F7Y/NorthstarPrime/blob/primedev/primedev/codecs/miles/sample.cpp).

Issue #530, opened 2023-08-02
My take on the plugin system


F1F7Y:
We all can agree that plugins can be a useful tool. In my opinion the plugin system as it is now is insufficient and too bothersome ( especially the version system ) to use.

# How

All engine libraries ( can't check rtech rn ) expose a function called `CreateInterface` which takes a string as an argument. Calling this with a valid interface string returns a pointer to a global class instance connected to the string constant. We already use this ( although our way is taken from ttf2sdk and should be updated to be better ).

Personally I think we should copy this system and expose our own `CreateInterface` function and our own interfaces. This would mean a plugin would be partially independent of the version of the interfaces we expose meaning unless one of the interfaces you were using had a breaking update you wouldn't have to update.

# Example

Let's say we expose `"NSSquirrelFuncs001"` and `"NSCorePluginInterface001"`

If one wants to add new functions to the squirrel interface we would just expose both `NSSquirrelFuncs001"` and `"NSSquirrelFuncs002"` keeping compatibility.

If we for example wanted to change a function in `"NSCorePluginInterface001"` we would only expose `NSCorePluginInterface002"` breaking only the plugins which use this interface.

Using deprecation and other methods we could slim down the plugins effected by interface updates even more.

# What even is the pointer that `CreateInterface` returns

It's a pointer to a class. The idea is you only should use the abstract class from which it inherits but since you get the full instance of the class you can even use other functions ( eg: iterator in g_pcvar ).

# C

C compatibility would still be kept as classes are basically just structs with a pointer to the vtable ( struct of virtual functions ) at the start.

This was written on a phone so there may be mistakes. I also probably forgot some other talking points I had.

# Ending notes

Mimicking the `CreateInterface` system also encourages people to check engine libraries.

I'm open to implementing this on my own, just wanted to spark discussion, if someone has an opinion too long or not fitting for GitHub make a thread in discord referencing this issue.

F1F7Y:
As this has been getting quite some attention I'd like to clear some things up:

Only custom systems ( eg Atlas ) and northstar helper functions ( plugin inform helpers, ... ) should be exposed. For example the squirrel setter/getter s should not be in an interface and the plugin should rather either get them from the modules or reconstruct them.

I dont think northstar should split up into separate dlls each implementing a set of interfaces. The only split i think makes sense would be client ( including listen server ) and dedicated.

[...]

PR #605, opened 2023-11-29
Replace Plugin manifest JSON with symbols in the lib binary

uniboi:
For some reason only known to greater minds, Emma decided to store info about plugins in a json file, compiled into the binary itself.
This PR just changes the behaviour to symbols in the binary making them easier to compile, write and load.

For example, this is now a valid plugin in C
```c
#include <stdint.h>

enum PluginContext {
  PCTX_DEDICATED = 0x1, // load on dedicated servers
  PCTX_CLIENT // unused atm
};

const char* PLUGIN_NAME = "P4_TEST"; // unused atm
const char* PLUGIN_LOG_NAME = "P4"; // used for logging (and I think it's a good idea if the dependency constant would look like 'NSPLUG_${ABBREVIATION}'
const char* PLUGIN_DESCRIPTION = "a plugin that does things"; // unused atm
const char* PLUGIN_VERSION = "0.0.1"; // unused atm
const char* PLUGIN_DEPENDENCY_NAME = "P4_TEST"; // unused atm
const uint32_t PLUGIN_ABI_VERSION = 4; // this has to exactly match the one from Northstar. This sucks and should be changed in the future.

const uint8_t PLUGIN_CONTEXT = PCTX_DEDICATED | PCTX_CLIENT;

void PLUGIN_INIT(void* _1, void* _2) {}
```

### DIFF
--- a/NorthstarDLL/plugins/plugins.cpp
+++ b/NorthstarDLL/plugins/plugins.cpp
@@ -125,7 +125,7 @@ std::optional<Plugin> PluginManager::LoadPlugin(fs::path path, PluginInitFuncs* 
 
 	Plugin plugin = Plugin(pluginLib, m_vLoadedPlugins.size(), pathstring);
 
-	if (!plugin.valid || (!plugin.run_on_server && IsDedicatedServer())
+	if (!plugin.valid || (!plugin.run_on_server && IsDedicatedServer()))
 	{
 		freeLibrary(pluginLib);
 		return std::nullopt;
@@ -135,7 +135,6 @@ std::optional<Plugin> PluginManager::LoadPlugin(fs::path path, PluginInitFuncs* 
 	// NS::log::PLUGINSYS->info("Succesfully loaded {}", pathstring); // we log the same thing 20 lines later why is this here
 
 	plugin.logger = std::make_shared<ColoredLogger>(plugin.logName, NS::Colors::PLUGIN);
-	d
 	RegisterLogger(plugin.logger);
 	NS::log::PLUGINSYS->info("Loading plugin {} version {}", plugin.logName, plugin.version);
 	m_vLoadedPlugins.push_back(plugin);
--- a/NorthstarDLL/plugins/plugins.cpp
+++ b/NorthstarDLL/plugins/plugins.cpp
@@ -42,7 +42,7 @@ template <typename T> T GetSymbolValue(HMODULE lib, const char* symbol)
 	return addr ? *addr : NULL;
 }
 
-Plugin::Plugin(HMODULE lib, int handle, std::string path)
+Plugin::Plugin(HMODULE lib, int handle, std::string identifier)
 	: name(GetSymbolValue<char*>(lib, Symbol::NAME)), logName(GetSymbolValue<char*>(lib, Symbol::LOG_NAME)),
 	  dependencyName(GetSymbolValue<char*>(lib, Symbol::DEPENDENCY_NAME)), description(GetSymbolValue<char*>(lib, Symbol::DESCRIPTION)),
 	  api_version(GetSymbolValue<uint32_t>(lib, Symbol::REQUIRED_ABI_VERSION)), version(GetSymbolValue<char*>(lib, Symbol::VERSION)),
@@ -58,7 +58,7 @@ Plugin::Plugin(HMODULE lib, int handle, std::string path)
 {
 #define LOG_MISSING_SYMBOL(e, symbol)                                                                                                      \
 	if (!e)                                                                                                                                \
-		NS::log::PLUGINSYS->error("'{} does not export required symbol '{}'", path, symbol);
+		NS::log::PLUGINSYS->error("'{} does not export required symbol '{}'", identifier, symbol);
 	LOG_MISSING_SYMBOL(api_version, Symbol::REQUIRED_ABI_VERSION)
 	LOG_MISSING_SYMBOL(name, Symbol::NAME)
 	LOG_MISSING_SYMBOL(description, Symbol::DESCRIPTION)
@@ -72,7 +72,7 @@ Plugin::Plugin(HMODULE lib, int handle, std::string path)
 	{
 		NS::log::PLUGINSYS->error(
 			"'{}' has an incompatible API version number ('{}') in its manifest. Current ABI version is '{}'",
-			path,
+			identifier,
 			api_version,
 			ABI_VERSION);
 	}
@@ -119,7 +119,7 @@ std::optional<Plugin> PluginManager::LoadPlugin(fs::path path, PluginInitFuncs* 
 	{
 		NS::log::PLUGINSYS->error(
 			"'Failed to load library '{}'",
-			std::system_category().message(GetLastError())); // GetLastError??? We know the path that failed to load
+			std::system_category().message(GetLastError()));
 		return std::nullopt;
 	}
 
@@ -130,9 +130,6 @@ std::optional<Plugin> PluginManager::LoadPlugin(fs::path path, PluginInitFuncs* 
 		freeLibrary(pluginLib);
 		return std::nullopt;
 	}
-
-	// Passed all checks, going to actually load it now
-	// NS::log::PLUGINSYS->info("Succesfully loaded {}", pathstring); // we log the same thing 20 lines later why is this here
 
 	plugin.logger = std::make_shared<ColoredLogger>(plugin.logName, NS::Colors::PLUGIN);
 	RegisterLogger(plugin.logger);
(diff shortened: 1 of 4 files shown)


uniboi:
The command for clang-format in the readme doesn't work :)

Jan200101:
bikeshedding:

I don't like the names, they are very generic.
How about prefixing them with `PLUGIN_` ?
e.g. `const char PLUGIN_NAME[] = "ABC";`

---

regarding the
> // this has to exactly match the one from Northstar. This sucks and should be changed in the future.

comment: see #530

uniboi:
Good point

NorthstarDLL/plugins/plugins.cpp
```diff
@@ -54,116 +55,69 @@ std::optional<Plugin> PluginManager::LoadPlugin(fs::path path, PluginInitFuncs*
 	Plugin plugin {};
 
 	std::string pathstring = path.string();
-	std::wstring wpath = path.wstring();
+	std::wstring wpath = path.wstring(); // is there any reason to load the libs with wide strings?
```
Jan200101, line 58: There doesn't seem to be one.
You can replace the call to `LoadLibraryExW` with `LoadLibraryExA` and skip the conversion from `std::string` to `std::wstring`

NorthstarDLL/plugins/plugins.cpp
```diff
@@ -54,116 +55,69 @@ std::optional<Plugin> PluginManager::LoadPlugin(fs::path path, PluginInitFuncs*
...
+	char** name = (char**)GetProcAddress(pluginLib, "PLUGIN_NAME");
+	char** description = (char**)GetProcAddress(pluginLib, "PLUGIN_DESCRIPTION");
+	char** abbreviation = (char**)GetProcAddress(pluginLib, "PLUGIN_ABBREVIATION");
+	char** version = (char**)GetProcAddress(pluginLib, "PLUGIN_VERSION");
+	char** dependencyName = (char**)GetProcAddress(pluginLib, "PLUGIN_DEPENDENCY_NAME");
+	uint8_t* context = (uint8_t*)GetProcAddress(pluginLib, "PLUGIN_CONTEXT");
+	PLUGIN_INIT_TYPE plugin_init = (PLUGIN_INIT_TYPE)GetProcAddress(pluginLib, "PLUGIN_INIT");
+
+	// ensure all required values are found
+#define ASSERT_HAS_PROP(e, prop) if (!e) { NS::log::PLUGINSYS->error("'{}' is not exporting it's {}", pathstring, prop); freeLibrary(pluginLib); return std::nullopt;; }
```
Jan200101, line 81: I am not a fan of this error message.

How about `'{}' does not include {}`? Though perhaps that is a bit too unclear.

NorthstarDLL/plugins/plugins.cpp
```diff
@@ -54,116 +55,69 @@ std::optional<Plugin> PluginManager::LoadPlugin(fs::path path, PluginInitFuncs*
...
-	rapidjson_document manifestJSON;
-	manifestJSON.Parse(manifest.c_str());
+	uint32_t* abiVersion = (uint32_t*)GetProcAddress(pluginLib, "PLUGIN_ABI_VERSION");
+	char** name = (char**)GetProcAddress(pluginLib, "PLUGIN_NAME");
+	char** description = (char**)GetProcAddress(pluginLib, "PLUGIN_DESCRIPTION");
+	char** abbreviation = (char**)GetProcAddress(pluginLib, "PLUGIN_ABBREVIATION");
+	char** version = (char**)GetProcAddress(pluginLib, "PLUGIN_VERSION");
+	char** dependencyName = (char**)GetProcAddress(pluginLib, "PLUGIN_DEPENDENCY_NAME");
+	uint8_t* context = (uint8_t*)GetProcAddress(pluginLib, "PLUGIN_CONTEXT");
+	PLUGIN_INIT_TYPE plugin_init = (PLUGIN_INIT_TYPE)GetProcAddress(pluginLib, "PLUGIN_INIT");
```
Jan200101, line 78: since these are exported symbols, they are likely read only and should be marked as const so the compiler knows we shouldn't be doing anything funny.

NorthstarDLL/plugins/plugins.cpp
```diff
@@ -54,116 +55,69 @@ std::optional<Plugin> PluginManager::LoadPlugin(fs::path path, PluginInitFuncs*
...
-	int manifestSize = SizeofResource(datafile, manifestResource);
-	std::string manifest = std::string((const char*)LockResource(myResourceData), 0, manifestSize);
-	freeLibrary(datafile);
 
-	rapidjson_document manifestJSON;
-	manifestJSON.Parse(manifest.c_str());
+	uint32_t* abiVersion = (uint32_t*)GetProcAddress(pluginLib, "PLUGIN_ABI_VERSION");
+	char** name = (char**)GetProcAddress(pluginLib, "PLUGIN_NAME");
+	char** description = (char**)GetProcAddress(pluginLib, "PLUGIN_DESCRIPTION");
+	char** abbreviation = (char**)GetProcAddress(pluginLib, "PLUGIN_ABBREVIATION");
```
catornot, line 74: imo `DisplayName` was better, it's not really used for an abbreviation, this string just needs to be 9 characters long to be  like everything else in the log.

catornot (commented):
maybe use a struct instead since that wouldn't require exporting 5 different statics and I like struct. :)

catornot (changes requested):
`plugin_abi.h` should have a copy of `PluginContext` or it should be there instead of `plugins.h`.

NorthstarDLL/plugins/plugins.h
```diff
@@ -1,18 +1,28 @@
...
-	std::string description;
+	char* name;
+	char* displayName;
+	char* dependencyName;
+	char* description;
 
-	std::string api_version;
-	std::string version;
+	uint32_t api_version;
+	char* version;
```
Jan200101, line 26: either turn these const also or keep them as a string so they can be modified

Jan200101 (commented):
ABI_VERSION 4 makes me cry

NorthstarDLL/plugins/plugins.cpp
catornot: yeah `logName` would work

NorthstarDLL/plugins/plugins.cpp
uniboi: What about `{} does not export required symbol {}`?

NorthstarDLL/plugins/plugins.h
```diff
@@ -1,38 +1,76 @@
...
-	std::string version;
+	const uint32_t api_version;
+	const char* version;
 
 	// For now this is just implemented as the index into the plugins array
 	// Maybe a bit shit but it works
-	int handle;
+	const int handle;
 
 	std::shared_ptr<ColoredLogger> logger;
```
uniboi, line 22: I'll change the logging in a subsequent PR so I'm not touching it here

NorthstarDLL/plugins/plugins.cpp
Jan200101: Works for me

GeckoEidechse:
We'll have to update DiscordRPC as well, anyone wanna make a PR?

GeckoEidechse:
> > We'll have to update DiscordRPC as well, anyone wanna make a PR?
> 
> First this will have to get merged and rrplug has to get updated

Ye, that's fair. I just wanna avoid cherry-picking commits for a month due to the lack of update to rrplug/DiscordRPC ^^"

uniboi:
> I sure do love having extra dependencies when all we could have is a git submodule for the main repo for one header.

Nah that ain't it chief.

> Ye, that's fair. I just wanna avoid cherry-picking commits for a month due to the lack of update to rrplug/DiscordRPC ^^"

I'm not done with plugins yet so updating either doesn't make sense right now anyways. I'm just waiting for this PR to get merged

Jan200101:
> I'm not done with plugins yet so updating either doesn't make sense right now anyways.

not a fan of this.
Your work is independent of Northstar, so if you got hit by a bus tomorrow we'd have to either update the plugin shit or revert this.

GeckoEidechse:
@catornot do you think it would be feasible to have a branch in `rrplug` that already implements this? So that we can easily bump DiscordRPC and make a release with this? ^^"

catornot:
> @catornot do you think it would be feasible to have a branch in rrplug that already implements this? So that we can easily bump DiscordRPC and make a release with this? ^^"

yeah I'll make one once I will have time to work on rrplug again

uniboi:
There isn't really a point in this since #615 overhauls plugins completely

PR #595, opened 2023-11-12
Mod download UI integration

Alystrasz:
UI integration of the mod downloading feature merged in https://github.com/R2Northstar/NorthstarLauncher/pull/545

When downloading a mod, this will display download progress, then extraction progress onscreen. Once mod has been downloaded, mods will be reloaded, and server join will pursue.

_**NB: `NSReloadMods` currently doesn't support all mods (weapon skins for instance); some maps are supported.**_

Mods PR: https://github.com/R2Northstar/NorthstarMods/pull/761

### Test instructions

* Install launcher PR
* Install mods PR
* Enable mod downloading ~~in Controls menu~~ **(this is now done by switching the `allow_mod_auto_download` convar to `1`)**
* Join a server ("Space battle" test server should up for you to try this feature, if not ping me please)

### Media

Here's a demo of end-to-end custom map download:

https://github.com/R2Northstar/NorthstarLauncher/assets/11993538/99b61ebd-4dd3-4816-ba31-f6f9137b78fa

### Changes

* Expose download function to squirrel VM;
* `modState` variable is updated regarding mod download state (downloading, extracting etc);
* `NSGetModInstallState` function returns mod download state as a `ModInstallState` struct (defined in squirrel);
* `ModFetchingProgressCallback` function is used to get curl download progress information;
* `GetModArchiveSize` function gives information for extraction progress stats;
* `FetchModSync` isn't run in a thread anymore (since the whole downloading process is already run in a separate thread).

PR #611, opened 2023-12-02
Add slow-speed timeout when downloading mod

Alystrasz:
Adds a timeout if connection is too slow.

***

Previous PR text:

&nbsp;

While currently in a business trip, not having access to a super duper fiber Internet connection, I tried downloading some mods, which failed due to the poor Internet connection: mod request would timeout before mod is downloaded.

```text
[2023-12-02] [11:37:16] [NORTHSTAR] [info] Mods list successfully fetched.
[2023-12-02] [11:37:16] [NORTHSTAR] [info] Loading mods configuration...
[2023-12-02] [11:37:16] [NORTHSTAR] [info] ==> Loaded configuration for mod "Odd.s2space"
[2023-12-02] [11:37:16] [NORTHSTAR] [info] Done loading verified mods list.
[2023-12-02] [11:37:32] [NORTHSTAR] [info] Fetching mod archive from https://gcdn.thunderstore.io/live/repository/packages/odds-s2space-0.0.5.zip
[2023-12-02] [11:37:32] [NORTHSTAR] [info] Downloading archive to C:/users/steamuser/Temp/odds-s2space-0.0.5.zip
[2023-12-02] [11:38:02] [NORTHSTAR] [error] Fetching mod archive failed: Timeout was reached
[2023-12-02] [11:38:02] [NORTHSTAR] [error] Something went wrong while fetching archive, aborting.
```

Looking in the code, I realized I put the same timeout for the mod fetching request than for the mods list fetching request, which is **30 seconds**.

I thought about setting the timeout with a higher value (a few minutes or so), but then I remembered when I was a little boy, I would leave my computer running overnight to download games because my Internet connection was so shitty :>

In this PR, I ended up completely removing the timeout attribute, meaning mod fetching requests will never time out (according to [curl documentation](https://curl.se/libcurl/c/CURLOPT_TIMEOUT.html)).

```text
[2023-12-02] [12:50:45] [NORTHSTAR] [info] Mods list successfully fetched.
[2023-12-02] [12:50:45] [NORTHSTAR] [info] Loading mods configuration...
[2023-12-02] [12:50:45] [NORTHSTAR] [info] ==> Loaded configuration for mod "Odd.s2space"
[2023-12-02] [12:50:45] [NORTHSTAR] [info] Done loading verified mods list.
[2023-12-02] [12:50:56] [NORTHSTAR] [info] Fetching mod archive from https://gcdn.thunderstore.io/live/repository/packages/odds-s2space-0.0.5.zip
[2023-12-02] [12:50:56] [NORTHSTAR] [info] Downloading archive to C:/users/steamuser/Temp/odds-s2space-0.0.5.zip
[2023-12-02] [12:51:48] [NORTHSTAR] [info] Mod archive successfully fetched.
```

/storytelling off


Alystrasz:
This is [not straightforward](https://curl.se/docs/faq.html#Why_does_not_curl_return_an_erro) to implement, but I suppose I could implement such a behavior using [CURLOPT_LOW_SPEED_TIME](https://curl.se/libcurl/c/CURLOPT_LOW_SPEED_TIME.html).
Would you like me to do that in this PR?

NorthstarDLL/mods/autodownload/moddownloader.cpp
```diff
@@ -132,9 +132,13 @@ void FetchModSync(std::promise<std::optional<fs::path>>&& p, std::string_view ur
 	CURL* easyhandle;
 	easyhandle = curl_easy_init();
 
-	curl_easy_setopt(easyhandle, CURLOPT_TIMEOUT, 30L);
 	curl_easy_setopt(easyhandle, CURLOPT_URL, url.data());
 	curl_easy_setopt(easyhandle, CURLOPT_FAILONERROR, 1L);
+
+	/* abort if slower than 30 bytes/sec during 10 seconds */
```
GeckoEidechse, line 138: We only use `//` for single line comments AFAIK

```suggestion
	// abort if slower than 30 bytes/sec during 10 seconds
```

GeckoEidechse:
For testing I guess the easiest way would be to yank the ethernet cable half way through download?

Not sure if we even need testing for it though as the feature as a whole is still considered unreleased as it's gated behind convar.

Alystrasz:
> For testing I guess the easiest way would be to yank the ethernet cable half way through download?

I did the testing by cutting the Wi-Fi signal, but it's the same idea, yeah.

GeckoEidechse:
Given that you already tested it, I'm skip testing on my end and just merge directly instead ^^

GeckoEidechse:
wait, where did the arg removal go, hol' up.

GeckoEidechse:
Oh wait, so the change from c656859bcf8164e6c078e61fa9a95ca9b145004c was already merged in #595. Guess I'm renaming this PR here then so it's a bit better reflecting of the actual change, lol

GeckoEidechse:
Okay, now it should be ready to merge, lol

PR #615, opened 2023-12-14
Plugin interfaces (plugins v4)

uniboi:
Replaces the current plugin api with source interfaces.

- backwards compatible
- no more json in binaries (wtf)
- does not rely on structs from third party libraries (wtf)
- actually initializes variables
- no more basically unused classes

The launcher exposes almost everything required by plugins in interfaces that allow for backwards compatibility.
The only thing that's passed to a plugin directly is the northstar dll HWND and a struct of data that's different for each plugin.

Plugins are required to expose a `void* CreateInterface(const char* name, int* status)` function to share their own interfaces.
The launcher loads the `PluginId` interface from the plugin to query info such as it's name.

Plugins can use the `CreateInterface` function exposed by the northstarDll to use northstar interfaces such as for logging.
An interface is just an abstract class to force all functions into a vftable.

### NSSys001

Exposes some system functionality to plugins

```c
// 32 bit
enum LogLevel {
  INFO,
  WARN,
  ERR,
};

// handle: handle of the plugin. Passed to the plugin on init.
void Log(HMODULE handle, LogLevel level, char* msg); // logs a message with the plugin's log name
void Unload(HMODULE handle); // unloads the plugin
void Reload(HMODULE handle);
```

## Required Plugin Interfaces

Interfaces that *have* to be exposed for the plugin to be loaded.

### PluginId001

```c
// strings of data about the plugin itself. may be extended in the future
// 32 bit
enum PluginString {
  NAME, // the name of the plugin
  LOG_NAME, // the name used for logging
  DEPENDENCY_NAME, // the name used for squirrel dependency constants created. The value returned for this has to be a valid squirrel identifier or the plugin will fail to load
}

// bitfields about the plugin
// 32 bit
enum PluginField {
  CONTEXT // 0x1 if the plugin is allowed to run on dedicated servers and 0x2 if the plugin is allowed to run on clients (is this even needed seems useless to me)
}

char* GetString(PluginString prop);
i64 GetField(PluginField prop);
```

### PluginCallbacks001

```c
struct PluginNorthstarData { HMODULE handle; };

// COPY THE initData IT MAY BE MOVED AT RUNTIME
void Init(HMODULE nsModule, const PluginNorthstarData* initData, bool reloaded); // called after the plugin has been validated. The nsmodule allows northstar plugins to work for the ronin client as well (assuming they update their fork lmao)
void Finalize(); // called after all plugins have been loaded. Useful for dependencies
void Unload(); // called just before the plugin is getting unloaded
void OnSqvmCreated(CSquirrelVM* sqvm); // the context of the sqvm is contained in the instance
void OnSqvmDestroying(CSquirrelVM* sqvm); // callback with the sqvm instance that's about to be destroyed (for UI, CLIENT is destroyed for some reason??)
void OnLibraryLoaded(HMODULE module, const char* libraryName); // called for any library loaded by the game (for example engine.dll)
void RunFrame(); // just runs on every frame of the game I think
```

### What's an interface anyways?

Interfaces are just abstract classes. So make sure the first parameter is always a pointer to the instance of the interface you're using.

an example what `NSSys001` looks like in C:
```c
typedef enum {
  LOG_INFO,
  LOG_WARN,
  LOG_ERR,
};

typedef struct CSys {
  struct {
    void (*log)(struct CSys* self, HMODULE handle, LogLevel level, char* msg);
    void (*unload)(struct CSys* self, HMODULE handle);
  }* vftable;
} CSys;

// use like this
g_c_sys->vftable->log(g_c_sys, g_handle, LOG_INFO, "my balls are itching");
```

Interfaces are created with `CreateInterface` that's exposed in another dll.

## Removed stuff

- plugins no longer get cvar constructs etc. Instead they are required to use the game's interfaces (the shocker)
- no squirrel api functions are exposed. Plugins need to collect the pointers themselves
- plugins are required to implement their own message buffer for the sqvm if necessary. Thank god because the one in NS sucks
- no more `plugin_abi.h` that you can include in plugins for the required typedefs. Not a big con because that barely worked to begin with because of the billion files the `plugin_abi.h` file included.

I've made an [example](https://github.com/uniboi/NSPluginTemplate/) plugin that I try to keep up to date with the new api

That's it not really that complicated imo


catornot (commented):
just first impressions. I still haven't started rrplug rewrite. advent of code takes too much time...

catornot:
eh where review

NorthstarDLL/core/sourceinterface.h
```diff
@@ -1,6 +1,13 @@
 #pragma once
 #include <string>
 
+// interface return status
+enum class InterfaceStatus : int
```
catornot, line 5: this should maybe have a deprecated option

NorthstarDLL/squirrel/squirrel.cpp
```diff
@@ -157,57 +157,6 @@ const char* SQTypeNameFromID(int type)
...
-
-	s->__sq_getentityfrominstance = __sq_getentityfrominstance;
-	s->__sq_GetEntityConstant_CBaseEntity = __sq_GetEntityConstant_CBaseEntity;
-
-	s->__sq_pushnewstructinstance = __sq_pushnewstructinstance;
-	s->__sq_sealstructslot = __sq_sealstructslot;
-}
-
 // Allows for generating squirrelmessages from plugins.
 void AsyncCall_External(ScriptContext context, const char* func_name, SquirrelMessage_External_Pop function, void* userdata)
```
catornot, line 161: this seams to unused?
I like this function so consider keeping it somehow or else I would have to implement it in rrplug :<

uniboi: Honestly I think it's best if the MessageBuffer gets rewritten entirely. Different PR though.
It's pretty easy to implement in plugins though.
```rust
fn RunFrame() {
  for callback in rfCallbacks {
    callback();
  }
}

fn RunNextFrame(callback: fn()) {
  rfCallbacks.push(callback);
}

// ...
RunNextFrame(|| {
  SQObject fn;
  SQRESULT r = sq_getfunction(sqvm, name, &fn, 0);
  if(r == SQRESULT_ERROR) // it's an sqresult iirc
    return; 
  sq_pushobject(sqvm, fn);
  sq_pushroottable(sqvm);
  sq_call(sqvm, 1, false, true);
});
```

NorthstarDLL/core/sourceinterface.h
uniboi: The enum is copied from source. It's honestly pretty useless because you get a nullpointer when it fails anyways.
I don't really see the value of providing a status for deprecated interfaces. What is the plugin going to do when it's deprecated?
Storing if an interface is deprecated also adds a bit more complexity.

NorthstarDLL/squirrel/squirrel.cpp
catornot: well it would require each plugin that needs that to add a atomic operation to runframe and keep track of the sqvm pointer which might not be a lot anyway `¯\_(ツ)_/¯`

NorthstarDLL/core/sourceinterface.h
catornot: the interface status can give more a reasons for why you would get a null pointer or just as a notice for deprecation

NorthstarDLL/squirrel/squirrel.cpp
uniboi: it doesn't have to use an mutex or something like that, the performance overhead is negligible

catornot: it does or else it won't be thread safe. I would use channel thought.

GeckoEidechse:
@uniboi heads-up that there are merge conflicts here rn

NorthstarDLL/core/sourceinterface.h
uniboi: But what are plugins going to do about it?
Old interfaces work the exact same as newer ones, they just have less functionality because new functions get added at the bottom of the vftable.
When an interface is removed, the plugin should realize that it got a null pointer and the status IFACE_FAILED and unload itself / shutdown

catornot: > But what are plugins going to do about it?

better feedback to the end users.

uniboi: there isn't any feedback needed because old interfaces work the same like newer ones, they just have less functions available.
There are no deprecated interfaces. At most, the individual functions of the interface should be deprecated and in that case they should log a warning or something themselves and not the consuming plugin.

ASpoonPlaysGames (changes requested):
Whitespace lol

also quite a few variables could probably be const (and maybe even static?)

GeckoEidechse (approved):
Confirmed working in testing together with https://github.com/R2Northstar/NorthstarDiscordRPC/pull/17

Launched the game and joined a PvP and a Parkour server each while checking that Discord activity would update accordingly.

catornot (approved):
everything looks good now :)
seams to be all working well with rrplug.

although probably later there should be a pr to expose access to some hooked northstar functions.

uniboi:
yeah interfaces aren't done and I wouldn't even consider the ones in this PR final, this is just to replace the plugin sys and expand on interfaces later

uniboi:
what

uniboi:
note to self; read git output before force pushing

primedev/plugins/interfaces/sys/ISys.cpp
```diff
@@ -0,0 +1,66 @@
...
+			break;
+		case LogLevel::ERR:
+			spdLevel = spdlog::level::level_enum::err;
+			break;
+		default:
+			NS::log::PLUGINSYS->warn("Attempted to log with invalid level {}. Defaulting to info", (int)level);
+		case LogLevel::INFO:
+			spdLevel = spdlog::level::level_enum::info;
+			break;
+		}
```
ASpoonPlaysGames, line 26: No `critical` level? ig we dont rly use it in northstar so whatever

primedev/plugins/pluginmanager.h
```diff
@@ -0,0 +1,33 @@
...
+	void InformSqvmCreated(CSquirrelVM* sqvm) const;
+	void InformSqvmDestroying(CSquirrelVM* sqvm) const;
+	void InformDllLoad(HMODULE module, fs::path path) const;
+	void RunFrame() const;
+
+private:
+	void InformAllPluginsInitialized() const;
+
+	std::vector<Plugin> plugins;
+	std::string pluginPath;
```
ASpoonPlaysGames, line 28: I'd prefer if member variables were prefixed with `m_`

primedev/plugins/plugins.cpp
```diff
@@ -1,340 +1,235 @@
...
+Plugin::Plugin(std::string path) : location(path)
 {
-	spdlog::source_loc src {};
-	src.filename = msg->source.file;
-	src.funcname = msg->source.func;
-	src.line = msg->source.line;
-	auto&& logger = g_pPluginManager->m_vLoadedPlugins[msg->pluginHandle].logger;
-	logger->log(src, (spdlog::level::level_enum)msg->level, msg->msg);
-}
+	HMODULE m = GetModuleHandleA(path.c_str());
```
ASpoonPlaysGames, line 27: single letter variable names are cringe.

primedev/plugins/plugins.cpp
```diff
@@ -1,340 +1,235 @@
...
 
-	if (std::find_if(
-			plugin.dependencyName.begin(),
-			plugin.dependencyName.end(),
-			[&](char c) -> bool { return !((c >= 'A' && c <= 'Z') || (c >= 'a' && c <= 'z') || (c >= '0' && c <= '9') || c == '_'); }) !=
-		plugin.dependencyName.end())
+	if (!IsDedicatedServer() && !this->runOnClient)
 	{
-		NS::log::PLUGINSYS->warn("Dependency string \"{}\" in {} is not valid a squirrel constant!", plugin.dependencyName, plugin.name);
+		NS::log::PLUGINSYS->warn("Plugin {} did not request to run on clients", this->name);
```
ASpoonPlaysGames, line 121: isnt this more "non dedi" rather than "clients"

primedev/plugins/plugins.cpp
ASpoonPlaysGames, line 1: Why all of the `this->` in this file?

ASpoonPlaysGames (changes requested):

primedev/plugins/plugins.cpp
uniboi: Honestly maybe the  "context" if plugins are running on a dedi or client should get the boot entirely.

primedev/plugins/plugins.cpp
uniboi: implicit uses of the instance are unreadable imo

primedev/plugins/pluginmanager.h
uniboi: people wouldn't need to prefix member variables if they just used `this`

primedev/plugins/interfaces/sys/ISys.cpp
uniboi: is the critical level even used once in northstar

ASpoonPlaysGames: dont think so, but its still a thing that exists, don't think it really matters if it's an option here

GeckoEidechse (approved):
Re-confirmed working in testing. Tested with:

- loading outdated plugin (it simply gets unloaded as invalid plugin)
- Together with DiscordRPC PR: https://github.com/R2Northstar/NorthstarDiscordRPC/pull/17
  image

PR #617, opened 2023-12-18
Cherry pick logging refactor from primedev

ASpoonPlaysGames:
Honestly I only half like this refactor, I would prefer if we used the `{}` style of formatting stuff among other things.

Anyway this PR isn't about my opinion so whatever


RoyalBlue1:
cant the `{}` formatting be archived with `std::make_format_args` and `std::vformat` we are on cpp 20 anyways

ASpoonPlaysGames:
> cant the `{}` formatting be archived with `std::make_format_args` and `std::vformat` we are on cpp 20 anyways

Yeah, this was just me porting the refactor over without any changes

F1F7Y:
> cant the `{}` formatting be archived with `std::make_format_args` and `std::vformat` we are on cpp 20 anyways

The `{}` formatting comes from fmtlib which spdlog uses so you can very easily port to it if you want.

F1F7Y:
If i was working on refactoring logging now I'd probably have `Msg`, `DevMsg`, `Warning`, `Error`, `FatalError` functions where `DevMsg` would be stubbed in release.

Also having logging levels behind a general convar is a good idea.

ASpoonPlaysGames:
> If i was working on refactoring logging now I'd probably have `Msg`, `DevMsg`, `Warning`, `Error`, `FatalError` functions where `DevMsg` would be stubbed in release.

I definitely agree about the separation between `Error` and `FatalError`, because having to pass `NO_ERROR` to every error log to not have the program terminate is annoying

GeckoEidechse:
> I would prefer if we used the `{}` style of formatting stuff among other things

100% agree on that ^^"

GeckoEidechse:
So how do we approach this? Change `%s` back to `{}` before merge or merge first and then in a second PR revert back to `{}`?

ASpoonPlaysGames:
> So how do we approach this? Change `%s` back to `{}` before merge or merge first and then in a second PR revert back to `{}`?

Id say changing it back would be beyond the scope of the cherry pick.

Also changing it back now would make future primedev integrations more annoying

GeckoEidechse:
I'd so, stick with `%s` and then change it back later once we got primedev updated?
I'm fine with whatever gets us further faster, even if that means having a temporary "regression" [^1]

Could test late tonight. If someone else could give it a try earlier than that, would be perfect so that we can get it merged ASAP. \^\^

[^1]: Obvious remark that this is not a "regression" in the sense of that it breaks anything but it in the sense that it's style/behaviour that we might not necessary be satisfied with.

ASpoonPlaysGames:
> I'd so, stick with `%s` and then change it back later once we got primedev updated? I'm fine with whatever gets us further faster, even if that means having a temporary "regression" [1](#user-content-fn-1-b4c74259f9125641a7be0160b1e66d1f)

Yup works for me, Fifty might even make some more logging refactors between now and then anyway on primedev given the conversation thus far

ASpoonPlaysGames:
Man i wish i could compile locally right now instead of having to do everything inside github

itscynxx:
Game still seems to function fine from the testing I did
Was able to launch the game to the menu fine, as well as play a small amount of gun game

A log from successful menu launch:
[message.txt](https://github.com/R2Northstar/NorthstarLauncher/files/13746267/message.txt)

A log from having an error appear (purposefully, to test it):
[error.txt](https://github.com/R2Northstar/NorthstarLauncher/files/13746247/error.txt)

Things I have feedback on (though I understand this is simply taking from primedev, and not necessarily nitpicking stuff)
It should also be noted that I have the perspective of a helper and explaining to end users for this, and less of a development outlook

Milliseconds for the timer in the logs is most likely unnecessary
Milliseconds for the timer in the folder for the logs is most likely unnecessary
(should be noted that neither of these are _bad_, however it feels like it isn't needed and just takes up more space on screen. in the log, I suppose the milliseconds _could_ matter, but it becomes a "will end users understand why there's more numbers" scenario)

The errors not logging the mods really makes it not useful in a lot of scenarios however. A lot of the compile issues are because of missing dependencies, or conflicting mods. While some you can guess, having the mods in there at all makes the process 100x faster, especially accounting for explaining to the end user. Needing two logs to diagnose one issue is a little annoying. If the errors were still inside the full log, I would say it doesn't matter as much and just ask for the full log, but they seem to be omitted from the full log (or at least, from what I see. I'll attach the log with an error i created for testing, the same as in the previous error log)
[message.txt](https://github.com/R2Northstar/NorthstarLauncher/files/13746255/message.txt)

Apart from the issues I have (which, again, I realize is likely for a later date, easier to write them out here and reference/copy paste later than think of the same things twice), this pr does seem to work completely properly in game. No idea code wise

I believe the only public log reader is Spectre, which I'm already rewriting due to the code being bad, so I'll rewrite with this in mind. Logging the mods in `error.txt` (or logging the errors in `message.txt`) would make this much easier for Spectre and for helpers

ASpoonPlaysGames:
> I believe the only public log reader is Spectre, which I'm already rewriting due to the code being bad, so I'll rewrite with this in mind. Logging the mods in `error.txt` (or logging the errors in `message.txt`) would make this much easier for Spectre and for helpers

I don't remember the old logging printing out the installed mods on a script compile error? Although i guess your point is that people will start sending just the error log, which wont contain any mod information.

I don't see why we couldn't log the installed mods on a compile error, other than it maybe drowning out the actual error in a sea of mod prints.

ASpoonPlaysGames:
@itscynxx could you give it a test with a proper crash? like an access violation or something. There are some changes that i'm not 100% sure on in the crashhandler which would be nice to test

itscynxx:
> @itscynxx could you give it a test with a proper crash? like an access violation or something. There are some changes that i'm not 100% sure on in the crashhandler which would be nice to test

did we fix audio conflict crashes or something? trying to crash it with two audio grapple mods (only thing that came to mind)

itscynxx:
> > I believe the only public log reader is Spectre, which I'm already rewriting due to the code being bad, so I'll rewrite with this in mind. Logging the mods in `error.txt` (or logging the errors in `message.txt`) would make this much easier for Spectre and for helpers
> 
> I don't remember the old logging printing out the installed mods on a script compile error? Although i guess your point is that people will start sending just the error log, which wont contain any mod information.
> 
> I don't see why we couldn't log the installed mods on a compile error, other than it maybe drowning out the actual error in a sea of mod prints.

I mean more so that with the current system, you would get the error in the same log that gives you what mods are installed

If we're worried about mods list drowning out the error in `error.txt`, we could always print it after the error. Enabled and disabled mods should both be printed, as I've seen it where users don't have a dependency enabled way too many times...

GeckoEidechse:
Oh wait, this PR splits logs into mutiple txt files, right? Yeah I remember now why we held up #504 initially.

My concern with splitting up logs is that it's harder for people to upload their logs in tickets. I'd really love to hear the input of some helpers on this whether this is something we want or not.

GeckoEidechse:
Like we should still do this PR but maybe recombine logs again

itscynxx:
ok so found an issue. two grapple mods didn't crash, but downloading a lot of my kill sound mods did.
[error.txt](https://github.com/R2Northstar/NorthstarLauncher/files/13746533/error.txt)
This only writes one minidump to the log, which is fine I think. Currently I recommend helpers to check for multiple crash dumps to find conflicting audio, but I believe ucrtbase only crashes with conflicting audio

Another issue with this is that, for conflicting audio, we generally look at specific audio events to see which mods conflict. I think for this we can simply ask for the full log however. No idea how I'll work that properly into Spectre, but that's my own problem

Because this automatically closes the command prompt for logging, it makes your active application window a sort of "none" to windows (so you aren't in a window, and have to click into one to be in one again)

If you get an access violation you just can't close it for some reason... only on this pr
Completely random guess, might pop up between the window being set to "none" and during the game opening?

Task manager ending Titanfall 2 causes the error to go away as it should, however this shouldn't be the solution, especially with how prevalent (and easy to have) audio conflict errors are

itscynxx:
> Oh wait, this PR splits logs into mutiple txt files, right? Yeah I remember now why we held up #504 initially.
> 
> My concern with splitting up logs is that it's harder for people to upload their logs in tickets. I'd really love to hear the input of some helpers on this whether this is something we want or not ^^

Imo, the only important information from logs is
1. Northstar version
2. Mod version
3. Errors

If the error logs contain all 3 of these, it's likely easier for helpers. Less sifting through random game processes that mean nothing, more to the point. Full log would still be required for diagnosing specific, non obvious audio conflicts however

itscynxx:
My overall thoughts

# On splitting up logs:
Pros:
- More to the point
- Faster to read
- Easier for new people helping to understand logs

Cons:
- Explaining the folders to the end user
- Explaining the different files to the end user

Imo, this feature really comes down to the end user. A smart approach may be to write up a version that has combined logs, and be willing to release a minor version to change back to combined logs if need be. I think the split is better overall, _if_ the end user response is similar to the one we get now (where they don't usually understand, but a quick explanation gives them enough to get it)

# The files themselves:
I think the file names are a little confusing, but don't have any better options that come straight to mind. Not sure if I think a separate file for just warnings is really worth it, generally warnings boil down to "installed package wrong" or ports not being forwarded on clients and attempting to register to ms, but I'm not sure having it entirely hurts either. Would love to hear someone else's thoughts on this, as there may be more warnings that are more important to solely look at warnings for, that I don't know about

For `error.txt`, I think at minimum it should print Northstar version along with mods, and their versions. Having the errors separated themselves is quite nice, however `ctrl + f` exists, and would require also explaining sending the other, full log, a lot of the time. 

The issue I see with this is, for example, reloading mods trying to figure out compile errors and having it spam the console. I really have no good solution for this, however I think not having the mods or Northstar version in the logs really makes the split just not worth it regardless. 

If you're developing something, you already know what your error is from the in game error, and don't need it split into another log. 

If you're an end user, you have no idea what the error in game, or separated into a log is. 

If you're a helper, you have no idea what mods the user does or does not have (e.g. client kill callback issues. Do they have the dependency? If they do, is it disabled? If they don't, they need to download it. Do they have hud revamp? If yes, they need to download the fixing patch or remove hud revamp. Do they not have hud revamp? What else conflicts with it? Do they have an older version?), and no idea what Northstar version the user does or does not have (post merge version obviously). These issues are fixed by also getting the full log, but at that point, it simply makes more sense to just have it all in the full log

# An issue:
## Note: (seemingly) only for borderless window. Regardless of your stance on borderless window, people in the community use it
As of writing, having an access violation on borderless window (such as the extremely common one, audio conflicts) causes the following to happen:
1. Logging starts
2. Game starts launching
3. Logging window hides itself (likely to not confuse the user; a positive)
4. Current active window becomes none (have to click into a window to be in one)
5. Game crashes

Then from here the error acts very strangely. I've had multiple different results at seemingly random.
1. The error message is always on the top of the screen, isn't interactable, and cannot be closed apart from quitting in Steam or task manager
2. The logging window hangs and cannot be closed apart from quitting in Steam or task manager
3. The error message isn't interactable, _until_ you mess around changing windows for a little bit

Asking users to use task manager to close Northstar when this happens should not be the solution. No idea why it happens, but it does

Additionally, having an access violation on either fullscreen or windowed causes the error to go behind the fullscreen game, and causes the game to freeze. I don't know if we can make the error appear on top of the game (from testing on current version it also seems to go behind the game), but on current version, it doesn't freeze and continues to load, and work (just using one of the conflicting audio mods, not sure how it picks though). Which option is better is up for debate (e.g. make user fix their mods vs just letting them play)

# I am biased
Again, this is all coming from the bias of being a helper. Could be a positive or negative thing depending on how you look at it. 

Writing more words here than on my school assignments fr, enjoy the wall of text

ASpoonPlaysGames:
I have moved this back to the old 1 file format, slight change to the actual text formatting though i think? idk.

anyway yeah hopefully this can get merged now as a simple code refactor

ASpoonPlaysGames:
[nslog2024-01-18 21-48-31.txt](https://github.com/R2Northstar/NorthstarLauncher/files/13981959/nslog2024-01-18.21-48-31.txt)
example log file as of now

GeckoEidechse:
Console closing when game is launched is intentional, right?

ASpoonPlaysGames:
> Console window closing when game is launched is intentional, right?

think so, yeah

ASpoonPlaysGames:
`-wconsole` should stop the console from disappearing

GeckoEidechse:
Merge conflicts atm btw ._.

GeckoEidechse:
Looks like either #615 or #626 put merge conflicts on this PR again :c

ASpoonPlaysGames:
Should be dealt with

GeckoEidechse:
Testing checklist:

- [ ] General Linux compat (console window close might be an issue)
- [ ] Docker image
- [ ] General Windows testing
- [ ] Windows dedi testing

(If anyone already did one of these, LMK so I can check them off as done)

Jan200101:
>     * [ ]  General Linux compat (console window close might be an issue)

Tested under Fedora 39 with and without gamescope with no issues.
Since the Steam Deck is just a modified Arch install with gamescope, there will likely be no problems.

GeckoEidechse:
> Tested under Fedora 39 with and without gamescope with no issues.
> Since the Steam Deck is just a modified Arch install with gamescope, there will likely be no problems.

Perfect, thanks <

PR #619, opened 2023-12-18
Set a better `AccessModifierOffset` in `clang-format`

ASpoonPlaysGames:
Default appears to be -2, which causes clang-format to mix tabs and spaces


uniboi:
what code is there to review

ASpoonPlaysGames:
> what code is there to review

More just the formatting changes that this incurs

GeckoEidechse:
@uniboi precisely what @ASpoonPlaysGames said. There's obviously no code review but rather the question whether the changes are wanted or not. The `needs code review label` is just the closest we have to indicate that there's still a need for someone to look at the changes and decide whether they are wanted (next to the obvious `feedback wanted` label)


@ASpoonPlaysGames should we merge this before or after some primedev cherry-picks? primedev doesn't have it so it might increase the div unless someone backports it to there ^^

ASpoonPlaysGames:
> @ASpoonPlaysGames should we merge this before or after some primedev cherry-picks? primedev doesn't have it so it might increase the div unless someone backports it to there ^^

idk, I fix formatting on primedev cherry picks anyway so i dont think it matters if this comes first for what i'm doing

GeckoEidechse:
Merging with 2 in favour (@ASpoonPlaysGames @p0358) and 1 slightly (but not fully) opposed (@Jan200101)

I asked for feedback on the `#development` on the Northstar Discord multiple times and this is all I got so I assume the remaining contributors are indifferent on the change.

PR #621, opened 2023-12-20
Add Action to add label to PR on merge conflict

GeckoEidechse:
Adds a GitHub Action that auto-adds a label to a PR in case there are merge conflicts.

Action in use is https://github.com/marketplace/actions/auto-label-merge-conflicts

I'm not sure if the label also gets auto-removed again but I guess we'll find out :P

### DIFF
--- a/.github/workflows/merge-conflict-auto-label.yml
+++ b/.github/workflows/merge-conflict-auto-label.yml
@@ -1,0 +1,16 @@
+name: Merge Conflict Auto Label
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  triage:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


GeckoEidechse:
For obvious reasons there isn't anything to test or code review here, so instead just react to the PR with a :+1: or :-1: if you want it or not.

ASpoonPlaysGames:
> I'm not sure if the label also gets auto-removed again but I guess we'll find out :P

https://github.com/mschilde/auto-label-merge-conflicts/blob/591722e97f3c4142df3eca156ed0dcf2bcd362bd/lib/main.ts#L145-L150 should do

ASpoonPlaysGames (approved):
I like this, automation is good.

I guess we kinda have to test this in production, so uhhh
fuck it, we ball

PR #624, opened 2023-12-21
Folder restructuring from primedev

ASpoonPlaysGames:
Has to cherry pick by hand because github is dumb but yeah

Cherry picking ccf5772e6f09432d23c4712b03e43d4433d6f08a

PR #625, opened 2023-12-21
Use .def for function exports

ASpoonPlaysGames:
Depends on #624


GeckoEidechse (approved):
Given that this is basically a tiny refactor that doesn't affect logic, no in-game testing should be needed for this.

As for code review, `.def` is an MSVC thing but given that wsock32 stuff also uses it, not an issue IMO.

PR #626, opened 2023-12-22
Port navmesh debug renderer from primedev

ASpoonPlaysGames:
Comes with a bunch of dependencies that I also ported

depends on
- #624


ASpoonPlaysGames:
image
I'm considering this tested

ASpoonPlaysGames:
To turn this on, do `sv_cheats 1; enable_debug_overlays 1` and then `navmesh_debug_hull 1` (or any value 1-4 iirc)

primedev/toolframework/itoolentity.h
```diff
@@ -0,0 +1,46 @@
...
+	virtual void sub_1805E57B0() = 0;
+	virtual void sub_1805E5860() = 0;
+	virtual void sub_1805E55A0() = 0;
+	virtual void sub_1805E49C0() = 0;
+	virtual void sub_1805E7580() = 0;
+	virtual void sub_1805E86A0() = 0;
+	virtual void sub_1805E69B0() = 0;
+	virtual void sub_1805E4D70() = 0; // IClientTools::DrawSprite
+	virtual void sub_1805E57A0() = 0;
+	virtual bool GetLocalPlayerEyePosition(Vector3& org, QAngle& ang, float& fov) = 0;
```
GeckoEidechse, line 27: I assume the remaining fields are just padding?

primedev/core/math/vector.h
```diff
@@ -1,47 +1,331 @@
...
+#include "math_pfns.h"
+
 #include <cmath>
 
-#pragma once
+#define DEG2RAD(a) (a) * (3.14159265358979323846f / 180.0f)
+#define RAD2DEG(a) (a) * (180.0f / 3.14159265358979323846f)
 
-union Vector3
+class Vector3
```
GeckoEidechse, line 11: I'm kinda surprised we're not using `Vector3` anywhere else so that changing it from `union` to a `class` would cause an issue ^^

primedev/client/debugoverlay.h
```diff
@@ -0,0 +1,34 @@
+#pragma once
```
GeckoEidechse, line 1: I'm very much considering doing a quick PR with just that change of moving typedefs to header file ^^

primedev/toolframework/itoolentity.h
GeckoEidechse: cc @F1F7Y

GeckoEidechse (commented):
Confirmed working testing on Forwardbase Kodai using the provided instructions.

primedev/core/math/vector.h
ASpoonPlaysGames: We are using it elsewhere, but not in any way that the struct doesn't support

primedev/toolframework/itoolentity.h
ASpoonPlaysGames: They are all just unknown functions, the names should match with ones that IDA generates

primedev/client/debugoverlay.h
F1F7Y: typedefs in header files just clutter up your intelisense. Using typedefs also means intellisense gives you the type only, not the signature when typing out a function call.

If you really want to just move all function pointers into headers and get rid of all typedefs that aren't used as arguments in a function.

GeckoEidechse: Wait, so why were they moved to header files in the first place then? Or were you just experimenting around at that point? ^^

ASpoonPlaysGames: I mean TBH typedefs shouldn't be used in 99% of cases anyway

F1F7Y: > Wait, so why were they moved to header files in the first place then? Or were you just experimenting around at that point? ^^

I needed to move the function pointers to the header fie so i could use them elsewhere. This meant i had to also move the typedefs.

ASpoonPlaysGames: I have now removed the typedefs and just used decltype where needed

primedev/core/math/vector.h
uniboi: there was genuinely no point for this to exist ngl

primedev/core/tier1.cpp
```diff
@@ -0,0 +1,19 @@
...
+{
+	HMODULE hModule = GetModuleHandleA(svModuleName.c_str());
+
+	if (!hModule)
+	{
+		spdlog::error("Failed to get module handle of '{}'!", svModuleName.c_str());
+		exit(EXIT_FAILURE);
+	}
+
+	CreateInterfaceFn fnCreateInterface = reinterpret_cast<CreateInterfaceFn>(GetProcAddress(hModule, CREATEINTERFACE_PROCNAME));
```
uniboi, line 16: somewhat offtopic but I was considering to just delete sourceinterface.h because it provides nothing useful.

ASpoonPlaysGames: yeah, primedev does that as well, it's on my list of things to cherry-pick.

uniboi (approved):
can't really comment on the structs or mesh renders but the code looks pretty good. I'm going to steal the debug renders for my plugins lmao

GeckoEidechse:
Technically not fully through code (re stuff @uniboi mentioned on structs and mesh renderers) but I'm trusting @F1F7Y on the initial code and @ASpoonPlaysGames on the cherry-pick, so merging regardless ^^

F1F7Y:
> Replying to @GeckoEidechse but don't want to ping others

The renderer itself is ported from recast and adapted for the debug overlay interface with some optimization ( range and plane based culling, outline duplicate culling ). I've tested it thoroughly in primedev.

PR #627, opened 2023-12-24
Simplify vscript stub function

ASpoonPlaysGames:
Depends on #624

Since we can get stack info from the sqvm, we don't need to give the function name in the template and stuff.

Cherry picked from primedev


GeckoEidechse:
What would be a good way to test this?

RoyalBlue1:
> What would be a good way to test this to double check that there's no regression in blocking the command? Not that I'm expecting one, just double checking ^^

Try calling a stubbed function like `script_client DevTextBufferWrite("test")`

primedev/squirrel/squirrel.cpp
```diff
@@ -620,32 +620,25 @@ template <ScriptContext context> void ConCommand_script(const CCommand& args)
...
+	SQStackInfos si;
+	g_pSquirrel<context>->sq_stackinfos(sqvm, 0, si);
 
-	char value[N];
-};
-
-template <ScriptContext context, TemplateStringLiteral funcName> SQRESULT SQ_StubbedFunc(HSquirrelVM* sqvm)
-{
-	spdlog::info("Blocking call to stubbed function {} in {}", funcName.value, GetContextName(context));
+	spdlog::warn("Blocking call to stubbed function {} in {}", si._name, GetContextName(context));
```
GeckoEidechse, line 628: Just noticed that previously we would print as `info` instead of `warn` which tbh was kinda stupid, so good idea on changing that in the process ^^

GeckoEidechse:
> Try calling a stubbed function like `script_client DevTextBufferWrite("test")`

Oh, my dumbass completely forgot about `script_client` and only tried `script`. Thanks <3

GeckoEidechse (approved):
Confirmed working testing by simply running `script_client DevTextBufferWrite("test")` in lobby

image

primedev/squirrel/squirrel.cpp
```diff
@@ -620,32 +620,25 @@ template <ScriptContext context> void ConCommand_script(const CCommand& args)
 }
 
 // literal class type that wraps a constant expression string
```
ASpoonPlaysGames, line 622:
```suggestion
```
Comment no longer relevant

uniboi (approved):
code looks good, doesn't really change anything

PR #630, opened 2023-12-28
Add cmake format check to CI

### LINKED ISSUE #629: Add Action to format cmake files
https://github.com/marketplace/actions/cmake-format-lint-action

We got mixed indentation in cmake files rn for example, an Action that checks format of it would be nice ^^

ASpoonPlaysGames:
Adds format-check for cmake files in CI

Closes #629


ASpoonPlaysGames:
Right, well this is kinda the best I could get. The way we were formatting stuff really just isn't possible to automate (since we don't rly follow lexical rules we just do it visually)

.cmake-format.json
```diff
@@ -0,0 +1,26 @@
...
+		"line_width": 120,
+		"tab_size": 4,
+		"use_tabchars": false,
+		"fractional_tab_policy": "use-space",
+		"max_subgroups_hwrap": 2,
+		"max_pargs_hwrap": 2,
+		"max_rows_cmdline": 2,
+		"separate_ctrl_name_with_space": false,
+		"separate_fn_name_with_space": false,
+		"dangle_parens": false,
```
GeckoEidechse, line 12:
```suggestion
		"dangle_parens": true,
```

GeckoEidechse (commented):
cmake formatting changes have already been merged into main and the CI has been tested to work.

No in-game testing needed as this only affects config file stuff.

GeckoEidechse (approved):
(forgot to press _approve_)

PR #636, opened 2024-01-12
Disable MAD on vanilla

itscynxx:
Simply adds a check for the `ON_DLL_LOAD` that runs the mod auto download code for if vanilla compatibility is enabled. 

Although it would never run on vanilla, some people are confused by the `Custom verified mods URL not found in command line arguments, using default URL.` and `Mod downloader initialized` added in logs when loading mod auto download, and, from testing, doesn't break anything to disable in this way. I was able to play a match of vanilla and a match of northstar without issues

No idea if this can be done better, just looked at what Spoon did for the master server auth and put it here

### DIFF
--- a/primedev/mods/autodownload/moddownloader.cpp
+++ b/primedev/mods/autodownload/moddownloader.cpp
@@ -1,5 +1,5 @@
 #include "moddownloader.h"
-#include "core/vanilla.h
+#include "core/vanilla.h"
 #include <rapidjson/fwd.h>
 #include <mz_strm_mem.h>
 #include <mz.h>
--- a/primedev/mods/autodownload/moddownloader.cpp
+++ b/primedev/mods/autodownload/moddownloader.cpp
@@ -1,4 +1,5 @@
 #include "moddownloader.h"
+#include "core/vanilla.h
 #include <rapidjson/fwd.h>
 #include <mz_strm_mem.h>
 #include <mz.h>
@@ -587,6 +588,9 @@ void ModDownloader::DownloadMod(std::string modName, std::string modVersion)
 
 ON_DLL_LOAD_RELIESON("engine.dll", ModDownloader, (ConCommand), (CModule module))
 {
+	if (g_pVanillaCompatibility->GetVanillaCompatibility())
+		return;
+	
 	g_pModDownloader = new ModDownloader();
 	g_pModDownloader->FetchModsListFromAPI();
 }


itscynxx:
Could someone explaining why the format check fails?

as far as I'm aware I do the same thing as done here: https://github.com/R2Northstar/NorthstarLauncher/blob/main/primedev/masterserver/masterserver.cpp#L605

unless this is some tabs vs spaces thing ig

ASpoonPlaysGames:
> > Could someone explaining why the format check fails?
> 
> You have some trailing whitespace in an empty line

Or at least that's what format check is saying

Alystrasz:
What's the exact procedure to launch vanilla through Northstar and have access to logs?

ASpoonPlaysGames:
> What's the exact procedure to launch vanilla through Northstar and have access to logs?

launch northstar with the `-vanilla` launch arg

Alystrasz (approved):
PR confirmed working in testing, `g_pModDownloader` isn't initialized on vanilla plus, effectively removing related messages.

##### Without PR

```text
[2024-01-14] [00:08:31] [NORTHSTAR] [info] Enabling hook Status_ConMsg
[2024-01-14] [00:08:31] [NORTHSTAR] [info] Enabling hook CClientState_ProcessPrint
[2024-01-14] [00:08:31] [NORTHSTAR] [info] Registering Convar spewlog_enable
[2024-01-14] [00:08:31] [NORTHSTAR] [info] Mod downloader initialized
[2024-01-14] [00:08:31] [NORTHSTAR] [info] Custom verified mods URL not found in command line arguments, using default URL.
[2024-01-14] [00:08:31] [NORTHSTAR] [info] Enabling hook OpenExternalWebBrowser
[2024-01-14] [00:08:31] [NORTHSTAR] [info] Registering Convar ns_prefer_datatable_from_disk
```

##### With PR

```text
[2024-01-14] [00:12:52] [NORTHSTAR] [info] Enabling hook Status_ConMsg
[2024-01-14] [00:12:52] [NORTHSTAR] [info] Enabling hook CClientState_ProcessPrint
[2024-01-14] [00:12:52] [NORTHSTAR] [info] Registering Convar spewlog_enable
[2024-01-14] [00:12:52] [NORTHSTAR] [info] Enabling hook OpenExternalWebBrowser
[2024-01-14] [00:12:52] [NORTHSTAR] [info] Registering Convar ns_prefer_datatable_from_disk
```

GeckoEidechse:
I'm kinda curious what would happen if MAD were to be called in vanilla (e.g. via some mod). Just null pointer deref and crash ig?

itscynxx:
> I'm kinda curious what would happen if MAD were to be called in vanilla (e.g. via some mod). Just null pointer deref and crash ig?

Me too, got any easy way of testing it?
Not sure if just calling `DownloadMod` does all of it or not and it's kinda late to be thinking here lol

ASpoonPlaysGames:
> I'm kinda curious what would happen if MAD were to be called in vanilla (e.g. via some mod). Just null pointer deref and crash ig?

The various SQ functions should have checks added to ensure that they don't do this.

catornot:
> The various SQ functions should have checks added to ensure that they don't do this.

It would be better if they wouldn't even get registered at all if MAD is disabled.

ASpoonPlaysGames:
> > The various SQ functions should have checks added to ensure that they don't do this.
> 
> It would be better if they wouldn't even get registered at all if MAD is disabled.

True, but that's more complex, and ADD_SQFUNC doesn't support it

Alystrasz:
> Me too, got any easy way of testing it? Not sure if just calling `DownloadMod` does all of it or not and it's kinda late to be thinking here lol

I believe calling `DownloadMod` would straight up crash the client, which is better than downloading anything I guess

GeckoEidechse:
With #751 merged, is this still relevant?

ASpoonPlaysGames:
The error message is probably gone now, but if we are in vanilla mode it's probably better to just never instantiate the mod downloader? 

Would be worth checking the codebase for things that grab the static instance and make sure that they check for nullptr though

GeckoEidechse:
Yeah, if we avoid loading MAD on vanilla then we also need to check that the uninstantiated object is never loaded cause otherwise it's gonna crash on reading random uninitialised memory.

ASpoonPlaysGames:
> Yeah, if we avoid loading MAD on vanilla then we also need to check that the uninstantiated object is never loaded cause otherwise it's gonna crash on reading random uninitialised memory.

This is why just grabbing static things blindly is bad

GeckoEidechse:
Putting into draft so that I know what to focus on and as per discussion this PR is blocked by the lack of checks that we do not access the uninitialised object. @itscynxx if you (or anyone else) disagrees, feel free to undraft it as is or leave a comment for me to undraft and we can take another look.

PR #637, opened 2024-01-13
Fix debug overlay not rendering script function stuff

ASpoonPlaysGames:
Basically when i integrated some debug overlay stuff from `primedev`, I mistakenly called the original version of the hooked function, instead of the new version. The reason for this was because `primedev` doesn't use `AUTOHOOK`, but we do, and the name of the new function on `primedev` is the same as the original function on main. fun.

Offending commit (i think) https://github.com/R2Northstar/NorthstarLauncher/commit/6e32a584804158daec2f57cfe3fa058f1ce3d0a7

### DIFF
--- a/primedev/client/debugoverlay.cpp
+++ b/primedev/client/debugoverlay.cpp
@@ -170,7 +170,7 @@ int* g_nRenderTickCount;
 int* g_nOverlayTickCount;
 
 // clang-format off
-AUTOHOOK(DrawOverlay, engine.dll + 0xABCB0,
+AUTOHOOK(DrawOverlay, engine.dll + 0xABCB0, 
 void, __fastcall, (OverlayBase_t * pOverlay))
 // clang-format on
 {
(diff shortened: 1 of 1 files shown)


GeckoEidechse:
What would be the quickest way to test this? ^^"

ASpoonPlaysGames:
> What would be the quickest way to test this? ^^"

`sv_cheats 1; enable_debug_overlays 1; script DebugDrawSphere( GetPlayerArray()[0].GetOrigin(), 50, 255, 100, 0, true, 20 )` when in a map

Alystrasz (approved):
Confirmed working in testing.
Calling `DebugDrawSphere` and `DebugDrawBox`:

##### Without PR

image

##### With PR

image

GeckoEidechse (approved):
As far as I understand the code, the changes appear to be correct. Given that it works in testing and simply calls our version instead of native directly, I'd say this is enough for code review and good to merge :D

PR #638, opened 2024-01-13
Add `RunOnVanilla` to `mod.json` files

itscynxx:
again, just need to do this because my computer gives me now 1400 errors when trying to build northstar


itscynxx:
This was meant to be on my fork just to build because of the errors my pc causes, thanks for automatically putting it here github

itscynxx:
ngl i just put like an hour into this, got it working and went "lets fucking go it works" and then realized it isnt even useful for the problem i was hoping to solve

PR #639, opened 2024-01-14
Add deprecation notice for `run_northstar.txt`

Jan200101:
related to #508 
followed up by #512 

`run_northstar.txt` is a relic of times gone by and is no longer needed.

### DIFF
--- a/primedev/primelauncher/main.cpp
+++ b/primedev/primelauncher/main.cpp
@@ -254,20 +254,41 @@ void PrependPath()
 
 bool ShouldLoadNorthstar(int argc, char* argv[])
 {
+	bool loadNorthstar = true;
 	for (int i = 0; i < argc; i++)
+	{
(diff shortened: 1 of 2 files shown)


itscynxx:
Code seems to make sense, error code makes sense and gives a proper solution, and works in testing
image

HOWEVER, it should be noted that it only worked _after_ deleting `Northstar.dll`, after having tried just replacing the files, which didn't make the error appear. While this (deleting the files and not simply replacing) is realistically the recommended solution, it should be noted that there _could_ be people that miss the error message

Jan200101:
When does it occur twice?
when launching with just `-northstar` or via NorthstarLauncher?

itscynxx:
> When does it occur twice?
> when launching with just `-northstar` or via NorthstarLauncher?

launching Titanfall 2 via Steam using `-windowed -noborder`, with `run_northstar.txt` in the `Titanfall2` directory with a `1` inside it, on windows 10

PR #643, opened 2024-01-20
Add and use ScopeGuard

ASpoonPlaysGames:
These things are kinda cool, basically you use them to run a callback (generally a lamba) when they go out of scope.

It has a *very minor* performance loss compared to the goto stuff we were doing before, but it makes the code more readable since you wont be skipping declarations of variables and stuff.


GeckoEidechse (commented):
Based on quick code review I can confirm that basically all the goto jumps are moved into the scope guard which is called on deconstruction.

Would like someone else to double check the code though <3

ASpoonPlaysGames:
btw I tested this locally, by adding a log call inside the scope guard to ensure that the code is ran and works fine. seems to be good

GeckoEidechse:
Eh, time to test in prod I guess :trollface:

GeckoEidechse (approved):

PR #645, opened 2024-01-21
Warnings

ASpoonPlaysGames:
this build will absolutely fail all the time because warnings are fatal :)

but yeah


primedev/Launcher.cmake
```diff
@@ -26,3 +26,15 @@ set_target_properties(
...
+
+if(MSVC)
+    target_compile_options(NorthstarLauncher PRIVATE /W4 /WX)
+else()
+    target_compile_options(
+        NorthstarLauncher
+        PRIVATE -Wall
+                -Wextra
+                -Wpedantic
+                -Werror
```
Jan200101, line 38: Please do not make warnings fatal by default.
what is and isn't a warning can change a lot between compiler versions causing phantom errors that only appear in the CI but not locally or the other way around.

ASpoonPlaysGames: honestly i dont think this PR is ever gonna get merged, since it should really get split up into a bunch more PRs, so I think that it's ok to have fatal warnings on this branch specifically, so that it sort of "proves" that i didnt miss any.

PR #647, opened 2024-01-21
Address C4267 compiler warnings

ASpoonPlaysGames:
[C4267](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-3-c4267?view=msvc-170): implicit conversion from `size_t` to a smaller type

If you want to test and make sure i got them, you can compile with `/W4 /we4267` (msvc)

### DIFF
--- a/primedev/client/audio.cpp
+++ b/primedev/client/audio.cpp
@@ -445,7 +445,7 @@ bool, __fastcall, (void* sample, void* audioBuffer, unsigned int audioBufferLeng
 		else
 		{
 			data = dat->second.get();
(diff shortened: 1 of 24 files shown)


pg9182 (approved):
Looks good. Not tested locally.

nitpick: c++ generally uses static_cast/reinterpret_cast rather than c-style casts.

GeckoEidechse:
Trusting @pg9182 on the code review. Given that this is only casting and type declaration related fixes I'm fine with merging without testing.

PR #648, opened 2024-01-21
Address C4100 compiler warnings (unused var)

ASpoonPlaysGames:
Uses a macro to avoid the warning. I could have gone for just commenting out the parameter name in most cases, but for things like `ADD_SQFUNC` the parameter names aren't available as they are part of the macro.

### DIFF
--- a/primedev/server/serverpresence.h
+++ b/primedev/server/serverpresence.h
@@ -45,10 +45,10 @@ public:
 class ServerPresenceReporter
 {
 public:
-	virtual void CreatePresence(const ServerPresence* pServerPresence) {}
-	virtual void ReportPresence(const ServerPresence* pServerPresence) {}
-	virtual void DestroyPresence(const ServerPresence* pServerPresence) {}
-	virtual void RunFrame(double flCurrentTime, const ServerPresence* pServerPresence) {}
+	virtual void CreatePresence(const ServerPresence* /*pServerPresence*/) {}
+	virtual void ReportPresence(const ServerPresence* /*pServerPresence*/) {}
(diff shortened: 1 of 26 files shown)


primedev/server/serverpresence.h
```diff
@@ -45,10 +45,10 @@ struct ServerPresence
...
 {
 public:
-	virtual void CreatePresence(const ServerPresence* pServerPresence) {}
-	virtual void ReportPresence(const ServerPresence* pServerPresence) {}
-	virtual void DestroyPresence(const ServerPresence* pServerPresence) {}
-	virtual void RunFrame(double flCurrentTime, const ServerPresence* pServerPresence) {}
+	virtual void CreatePresence(const ServerPresence* /*pServerPresence*/) {}
+	virtual void ReportPresence(const ServerPresence* /*pServerPresence*/) {}
+	virtual void DestroyPresence(const ServerPresence* /*pServerPresence*/) {}
+	virtual void RunFrame(double /*flCurrentTime*/, const ServerPresence* /*pServerPresence*/) {}
```
GeckoEidechse, line 51: Huh?

ASpoonPlaysGames: I figured it would be dumb to have a function body of literally only `NOTE_UNUSED`. honestly these should probably be pure virtual?

GeckoEidechse (approved):
Confirmed working in testing.

GeckoEidechse (commented):
Code review wise, I still find the `do { } while(false)` to be a massive hack but apparently it's a common practice so ¯\\\_(ツ)\_/¯

PR #650, opened 2024-01-24
Pull zlib from custom fork

ASpoonPlaysGames:
Fixes builds failing due to https://github.com/madler/zlib/commit/2e3d86c4e12c4c23db4229eea7d0c1cd4571273e


GeckoEidechse (approved):
Pulls zlib from our custom fork https://github.com/R2Northstar/zlib to revert https://github.com/madler/zlib/commit/2e3d86c4e12c4c23db4229eea7d0c1cd4571273e which broke our builds.

No testing needed as it more or less reverts zlib back to a build that worked before.

PR #651, opened 2024-01-27
ScopeGuard improvements

ASpoonPlaysGames:
`std::function` introduces a layer of indirection that can be removed through templating the class.


Jan200101:
replaces the use of `std::function` with templates and a flag to check when we are done.
Code looks good, though this will result in multple versions of the class (which isn't a problem unless we are going ham with it)

GeckoEidechse (approved):
Confirmed working (i.e. not breaking) in quick testing ^^

PR #652, opened 2024-01-27
Remove unused variable in mod concommands

ASpoonPlaysGames:
yeah...

im not adding "needs testing" because uhhh, good luck testing this


GeckoEidechse (approved):
Simply removes an unused variable. Not much to review or test here.

Merging as soon as CI passes.

PR #662, opened 2024-02-05
Update default mods list URL

GeckoEidechse:
The default branch for the VerifiedMods repo was renamed from `master` to `main`. See https://github.com/R2Northstar/VerifiedMods/issues/28 for more info.

**Testing:**

Simply enable MAD by setting `allow_mod_auto_download` to `1` and join any server that requires some verified mod.


ASpoonPlaysGames (approved):
Imma be real this doesnt need testing, code is good (nothing really changed)

PR #664, opened 2024-02-11
Update silver-bun to `72c74b4`

ASpoonPlaysGames:
https://github.com/IcePixelx/silver-bun/tree/72c74b455bf4d02b424096ad2f30cd65535f814c

I don't think this can be submoduled btw, because of how silver-bun is set up.

A good few primedev refactoring PRs will depend on this


primedev/core/tier0.cpp
```diff
@@ -24,7 +24,7 @@ ON_DLL_LOAD("tier0.dll", Tier0GameFuncs, (CModule module))
 	TryCreateGlobalMemAlloc();
 
 	// setup tier0 funcs
-	CommandLine = module.GetExport("CommandLine").RCast<CommandLineType>();
-	Plat_FloatTime = module.GetExport("Plat_FloatTime").RCast<Plat_FloatTimeType>();
-	ThreadInServerFrameThread = module.GetExport("ThreadInServerFrameThread").RCast<ThreadInServerFrameThreadType>();
+	CommandLine = module.GetExportedFunction("CommandLine").RCast<CommandLineType>();
+	Plat_FloatTime = module.GetExportedFunction("Plat_FloatTime").RCast<Plat_FloatTimeType>();
+	ThreadInServerFrameThread = module.GetExportedFunction("ThreadInServerFrameThread").RCast<ThreadInServerFrameThreadType>();
```
GeckoEidechse, line 29: I'm assuming these kinda changes are related to the bump of silver-bun. Cause they don't appear so on a quick 5 second skim ^^"

ASpoonPlaysGames: Yup, function name changes and such

GeckoEidechse:
> I don't think this can be submoduled btw, because of how silver-bun is set up.

Out of scope for this PR odc but for long it might be worth to get in touch with @IcePixelx to potential adjust `silver-bun` so that we could submodule it or alternatively we create a fork with minimal changes that would allow for submoduling ^^

(Just thinking out loud here bascially)

GeckoEidechse:
Note to self to also add @IcePixelx as a co-author on merge, given that it's their code that is vendored here ^^

GeckoEidechse (approved):
Played a few rounds with it on client and didn't notice any issues so I'm marking this as working in testing ^^

GeckoEidechse:
Code reviews, I have serious doubt this will ever get properly reviewed. As such I will just trust @Icepixelx on the initial implementation of the lib, @F1F7Y on the integration in primedev, and @ASpoonPlaysGames on cherry-picking it correctly.

PR #665, opened 2024-02-11
Refactor DLL callbacks

ASpoonPlaysGames:
cherry-picked and slightly modified from primedev, due to missing earlier commits from primedev


GeckoEidechse:
Merged conflicts atm @ASpoonPlaysGames

GeckoEidechse:
@ASpoonPlaysGames just launching the game with this PR should be enough to verify that the PR works[^1], right?



[^1]: With the exemption of the Xinput dll of course. That also needs to be verified manually but cannot really do that on Linux I think ^^"

ASpoonPlaysGames:
As far as I know, yeah. Id go into a match just to make sure that everything is loaded and working though personally. Not sure why the xinput dll needs to be checked manually?

GeckoEidechse:
> Not sure why the xinput dll needs to be checked manually?

Last time we had a PR touch it, it accidentally broke the DLL override, meaning the Xinput DLL without ASLR was loaded again. Hence I'd prefer someone double checks that the right DLL is being loaded (don't remember what tool to use for that anymore) just so that we don't accidentally introduce a vulnerability into Northstar **again** :,)

barnabwhy:
> Not sure why the xinput dll needs to be checked manually?

it needs to be checked that `XInput1_3.dll` isn't being loaded and that instead `XInput9_1_0.dll` is instead just in case that behaviour was broken by this change

GeckoEidechse:
Related [^1]


> Last week, CyberpunkSaveEditor creator PixelRick disclosed that a buffer overflow vulnerability chained with a **non-ASLR DLL (xinput1_3.dll)** allows specially crafted save games or modifications to perform code execution on a PC.
<https://www.bleepingcomputer.com/news/security/cyberpunk-2077-bug-fixed-that-let-malicious-mods-take-over-pcs/>

(Not the best source but quickest I could find)

&nbsp;

- https://learn.microsoft.com/en-us/windows/win32/xinput/xinput-versions


[^1]: mostly putting it here so that I can reference it again in the future

GeckoEidechse:
Useful links

- https://learn.microsoft.com/en-us/sysinternals/downloads/listdlls
- https://github.com/R2Northstar/NorthstarLauncher/pull/583#pullrequestreview-1743399979

barnabwhy (approved):
works for me, loads xinput 9.1.0 and not 1.3
image

GeckoEidechse (changes requested):
As per @F1F7Y's comment on Discord (https://discord.com/channels/920776187884732556/936105663942246410/1273620659167297610)

> do not merge it
> new code forgets to call `g_pPluginManager->InformDllLoad`
> which in the old only some funcs call which is funky imo

barnabwhy:
worth noting this pr makes a change to the behaviour when xinput could not be loaded
previously an error message box would show and the game would exit
now there is just a error log and the game is allowed to run as normal
allowing the game to run without xinput will cause the game to silently accept no controller input on computers which for some reason don't have `XInput9_1_0.dll` in their System32 folder as they normally should

primedev/windows/libsys.cpp
```diff
@@ -0,0 +1,122 @@
...
+	if (lpLibFileName <= lpLibName && !strncmp(lpLibName, XINPUT1_3_DLL, strlen(XINPUT1_3_DLL) + 1))
+	{
+		hModule = o_LoadLibraryExA("XInput9_1_0.dll", hFile, dwFlags);
+
+		if (!hModule)
+		{
+			spdlog::error("Could not find XInput9_1_0.dll");
+
+			return nullptr;
+		}
```
barnabwhy, line 66: We can't handle it like this. it causes the game to quietly close with only a single error log in northstar

There's two ways we can work around this issue:
1. Restore the previous behaviour, where a message box is shown to the user
2. The following, which allows the game to run with no controller input:
```suggestion
		if (!hModule)
		{
			spdlog::error("Could not find XInput9_1_0.dll");

			// return -1 so that the game doesn't close (it has a check for if the hModule is 0 which then closes the game)
			return (HMODULE)(-1);
		}
```

barnabwhy (changes requested):

barnabwhy (approved):
everything now looks good to me
load success, game opens fine
image
load fail, error message shows
image

primedev/windows/libsys.cpp
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	// Get module base name in ASCII as noone wants to deal with unicode
+	CHAR szModuleName[MAX_PATH];
+	GetModuleBaseNameA(GetCurrentProcess(), hModule, szModuleName, MAX_PATH);
+
+	// DevMsg(eLog::NONE, "%s\n", szModuleName);
+
+	// Call callbacks
+	CallLoadLibraryACallbacks(szModuleName, hModule);
```
F1F7Y, line 33: Since we currently only use `CallLoadLibraryACallbacks` we can remove `CallLoadLibraryWCallbacks`. That is if the above code is kept.

primedev/windows/libsys.cpp
```diff
@@ -0,0 +1,123 @@
...
+void LibSys_RunModuleCallbacks(HMODULE hModule)
+{
+	if (!hModule)
+	{
+		return;
+	}
+
+	// Get module base name in ASCII as noone wants to deal with unicode
+	CHAR szModuleName[MAX_PATH];
+	GetModuleBaseNameA(GetCurrentProcess(), hModule, szModuleName, MAX_PATH);
```
F1F7Y, line 28: I don't think i tested this when i originally wrote this. Unsure if this is going to work as it doesn't convert from UTF-16 to ASCII. Considering the implementation of `CallLoadLibrary*Callbacks` `GetModuleFileName` might be better suited for the job.

F1F7Y (approved):
Did not test. I don't think this should break anything, but still left some comments.

Ship it :100:

GeckoEidechse:
Merging based on reviews.

PR #671, opened 2024-02-18
Clean up wsock proxy code and move wsock build system logic

Jan200101:
- moves `WSockProxy` to `wsockproxy/CmakeLists`
- remove exepath stuff from dllmain
  + its still done in loader.cpp because its used when reporting failure
- Disabled any Thread Library calls
  + we don't need to know about threads at all in the proxy
- yoink `wsock32.asm` into outer space
  + turns out, we can just call the function in a void shim since that wont touch the registers
- stop copying `wsock32.dll` to the game directory
  + this should improve the state of things when using the EA App

### DIFF
--- a/primedev/wsockproxy/dllmain.cpp
+++ b/primedev/wsockproxy/dllmain.cpp
@@ -5,7 +5,7 @@
 FARPROC p[73];
 HMODULE hL = 0;
 
-static wchar_t buffer[4096];
+static wchar_t wsockPath[4096];
 
 BOOL WINAPI DllMain(HINSTANCE hInst, DWORD reason, LPVOID)
 {
@@ -17,13 +17,13 @@ BOOL WINAPI DllMain(HINSTANCE hInst, DWORD reason, LPVOID)
 		if (!ProvisionNorthstar()) // does not call InitialiseNorthstar yet, will do it on LauncherMain hook
 			return true;
 
-        GetSystemDirectoryW(buffer, 4096);
-		swprintf_s(buffer, 4096, L"%s\\wsock32.dll", buffer);
+        GetSystemDirectoryW(wsockPath, 4096);
+		swprintf_s(wsockPath, 4096, L"%s\\wsock32.dll", wsockPath);
 
-		hL = LoadLibraryExW(buffer, 0, LOAD_WITH_ALTERED_SEARCH_PATH);
+		hL = LoadLibraryExW(wsockPath, 0, LOAD_WITH_ALTERED_SEARCH_PATH);
 		if (!hL)
 		{
-			LibraryLoadError(GetLastError(), L"wsock32.dll", buffer);
+			LibraryLoadError(GetLastError(), L"wsock32.dll", wsockPath);
 			return false;
 		}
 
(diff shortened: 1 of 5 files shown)


Jan200101:
I tested it under NorthstarProton
image


Both Klemmbaustein and elad tested it on Windows
image


Build was done via msvc-wine, I encourage a third party to test with their own compile.

Jan200101:
(ignore the force push, I didn't push the correct commit)

GeckoEidechse:
> I tested it under NorthstarProton [...] Both Klemmbaustein and elad tested it on Windows

 Awesome.  Fast tracking this then and marking testing part as done ^^

EladNLG (approved):
looked at: all of it
overall, looks good!

much better than... whatever was there before...

and big red make me go happy inside :]

ASpoonPlaysGames (commented):
When this gets merged, can someone ping me or something so that i can update my  warnings PR (it might not conflict but there are cmake changes that wont be needed anymore since no .asm)

primedev/wsockproxy/dllmain.cpp
```diff
@@ -1,90 +1,29 @@
...
-		return NULL;
-
-	DWORD length = GetModuleFileNameW(NULL, dest, destSize);
-	return length && PathRemoveFileSpecW(dest);
-}
-
-wchar_t exePath[4096];
-wchar_t buffer1[8192];
-wchar_t buffer2[12288];
+static wchar_t buffer[4096];
```
ASpoonPlaysGames, line 8: I'd prefer if this name was a bit better, not a huge deal since I know that's not rly the point of the PR and the variable existed already. Feel free to just mark as resolved if you want

primedev/wsockproxy/loader.cpp
```diff
@@ -8,6 +9,21 @@
 
 namespace fs = std::filesystem;
 
+static wchar_t buffer[8192];
```
ASpoonPlaysGames, line 12: Same as above

ASpoonPlaysGames (dismissed):
minor changes, code looks good otherwise

primedev/wsockproxy/dllmain.cpp
Jan200101: okay, what name do you propose?
How about `wsock_path`?

primedev/wsockproxy/loader.cpp
Jan200101: maybe `northstardll_path`?

primedev/wsockproxy/dllmain.cpp
ASpoonPlaysGames: Sure (but camel case, so `wsockPath`)

primedev/wsockproxy/loader.cpp
ASpoonPlaysGames: Honestly, just `northstarPath` would probably do? idk

GeckoEidechse (approved):
Simple sign-off review based on other reviews so that I can merge the PR ^^

PR #672, opened 2024-02-25
Fix bad offsets of `sq_getfunction`

catornot:
everything was seemingly working before but the offset for the server's `sq_getfunction` is still wrong by a bit and the client one was being set twice for some reason.

[prime dev reference](https://github.com/F1F7Y/NorthstarPrime/blob/e49d88f785372edf0821ed938b93dd28b6e36a9e/primedev/vscript/languages/squirrel_re/squirrel/sqapi.cpp#L491)

### DIFF
--- a/primedev/squirrel/squirrel.cpp
+++ b/primedev/squirrel/squirrel.cpp
@@ -715,7 +715,7 @@ ON_DLL_LOAD_RELIESON("client.dll", ClientSquirrel, ConCommand, (CModule module))
 
 	// Message buffer stuff
 	g_pSquirrel<ScriptContext::UI>->messageBuffer = g_pSquirrel<ScriptContext::CLIENT>->messageBuffer;
-	g_pSquirrel<ScriptContext::CLIENT>->__sq_getfunction = module.Offset(0x572FB0).RCast<sq_getfunctionType>();
+	g_pSquirrel<ScriptContext::CLIENT>->__sq_getfunction = module.Offset(0x6CB0).RCast<sq_getfunctionType>();
 	g_pSquirrel<ScriptContext::UI>->__sq_getfunction = g_pSquirrel<ScriptContext::CLIENT>->__sq_getfunction;
 	g_pSquirrel<ScriptContext::CLIENT>->__sq_stackinfos = module.Offset(0x35970).RCast<sq_stackinfosType>();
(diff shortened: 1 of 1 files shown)


primedev/squirrel/squirrel.cpp
```diff
@@ -756,9 +756,6 @@ ON_DLL_LOAD_RELIESON("client.dll", ClientSquirrel, ConCommand, (CModule module))
 
 	StubUnsafeSQFuncs<ScriptContext::CLIENT>();
 	StubUnsafeSQFuncs<ScriptContext::UI>();
-
-	g_pSquirrel<ScriptContext::CLIENT>->__sq_getfunction = module.Offset(0x6CB0).RCast<sq_getfunctionType>();
-	g_pSquirrel<ScriptContext::UI>->__sq_getfunction = g_pSquirrel<ScriptContext::CLIENT>->__sq_getfunction;
```
ASpoonPlaysGames, line 761: Why is the UI one not getting set anymore

catornot: it still [is](https://github.com/R2Northstar/NorthstarLauncher/blob/150f008871642daeb24591566e288865ecdc82bf/primedev/squirrel/squirrel.cpp#L718C1-L719C106).
I just removed this duplicate assignment.

GeckoEidechse (approved):
The changes simply update the offset values and also remove remove a duplicate assignment.

I trust @catornot to have double checked the new offset and not entered any wrong a values so I'm skipping verifying them.

Similarly I'm taking the liberty of skipping testing.

F1F7Y:
> The changes simply update the offset values and also remove remove a duplicate assignment.
> 
> I trust @catornot to have double checked the new offset and not entered any wrong a values so I'm skipping verifying them.
> 
> Similarly I'm taking the liberty of skipping testing.

The offsets are taken from primedev. The bad offsets caused crashes in my squirrel rewrite so i had to correct them and i made sure to double check both :).

GeckoEidechse:
>The offsets are taken from primedev. The bad offsets caused crashes in my squirrel rewrite so i had to correct them and i made sure to double check both :).

Even better, thanks for clarifying and letting me know <3
And sorry I didn't notice, would've added you as co-author on merge as well otherwise ^^"

PR #673, opened 2024-03-02
Don't allow too many arguments to be sent to the map command

ASpoonPlaysGames:
The map command should only take one argument, so any more than that is superfluous.


Alystrasz (approved):
Confirmed working in testing, passing more than one argument to the `map` command prints a warning message and does nothing, while currently the same command would succeed.

image

Not completely sure how you decided to focus on this specifically, but additional checks are always welcome I guess 😄

PR #674, opened 2024-03-02
Check OpenExternalWebBrowser custom flag better

ASpoonPlaysGames:


primedev/scripts/client/scriptbrowserhooks.cpp
```diff
@@ -9,7 +9,8 @@ void, __fastcall, (char* pUrl, char flags))
 // clang-format on
 {
 	bool bIsOriginOverlayEnabledOriginal = *bIsOriginOverlayEnabled;
-	if (flags & 2 && !strncmp(pUrl, "http", 4)) // custom force external browser flag
+	bool isHttp = !strncmp(pUrl, "http://", 7) || !strncmp(pUrl, "https://", 8);
```
Alystrasz, line 12: Does this mean https links currently do not open in external browser?
Because I'm pretty sure it was the case with the parkour mode :-?

GeckoEidechse (approved):
Confirmed working in testing. Code changes to check prefix also make sense.

PR #675, opened 2024-03-02
forcefully add .txt to all safe IO files

ASpoonPlaysGames:
This is a breaking change, all existing safe IO files wont be able to be read anymore since they wont have the .txt extension. I guess just get people to add it manually or something idk.

There is a absolutely a better solution than this, but yeah

### DIFF
--- a/primedev/mods/modsavefiles.cpp
+++ b/primedev/mods/modsavefiles.cpp
@@ -78,6 +78,9 @@ template <ScriptContext context> void SaveFileManager::SaveFileAsync(fs::path fi
 			{
 				mutex.get().lock();
 
+				// forcefully append .txt to the path
+				file = std::filesystem::path(file.string() + ".txt");
+
 				fs::path dir = file.parent_path();
 				// this actually allows mods to go over the limit, but not by much
 				// the limit is to prevent mods from taking gigabytes of space,
@@ -129,6 +132,9 @@ template <ScriptContext context> int SaveFileManager::LoadFileAsync(fs::path fil
 			{
 				mutex.get().lock();
 
+				// forcefully append .txt to the path
+				file = std::filesystem::path(file.string() + ".txt");
+
 				std::ifstream fileStr(file);
 				if (fileStr.fail())
 				{
@@ -175,6 +181,9 @@ template <ScriptContext context> void SaveFileManager::DeleteFileAsync(fs::path 
 			try
 			{
 				m.get().lock();
+
+				// forcefully append .txt to the path
+				file = std::filesystem::path(file.string() + ".txt");
 
 				fs::remove(file);
 
@@ -266,7 +275,8 @@ ADD_SQFUNC("void", NSSaveFile, "string file, string data", "", ScriptContext::SE
 	// this actually allows mods to go over the limit, but not by much
 	// the limit is to prevent mods from taking gigabytes of space,
 	// this ain't a cloud service.
-	if (GetSizeOfFolderContentsMinusFile(dir, fileName) + content.length() > MAX_FOLDER_SIZE)
+	// ".txt" is appended to the file name when saving/loading/deleting, so mimic that
+	if (GetSizeOfFolderContentsMinusFile(dir, fileName + ".txt") + content.length() > MAX_FOLDER_SIZE)
 	{
 		g_pSquirrel<context>->raiseerror(
 			sqvm,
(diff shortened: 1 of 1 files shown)


ASpoonPlaysGames:
> What's the reason for imposing the `txt` file extension?

To stop people from writing executable scripts like .bat files

Alystrasz:
@ASpoonPlaysGames build fails

EladNLG:
Add `.json` as an exception.

Safe I/O supports JSON and there is prractically no reason for `.json` to not be an available file extension.

Jan200101:
> Add `.json` as an exception.

I don't think it matters what extension the file has at the end, as long as it can't result in code execution by PEBKAC.

I do take issue with `.txt` being the default, the data being stored doesn't necessarily have to be plaintext. 
`.dat` would probably be way better (though perhaps a whitelist of extensions with a default appended `.dat` would be even more better for developers)

GeckoEidechse:
> The user has to MANUALLY open their titanfall 2 dir and go to R2Northstar/save_files/MOD.NAME/...

Not really. Before #674 `LaunchExternalWebBrowser()` had a bad check that (combined with Windows allowing to cd into and out of non-existing dirs), meant that `LaunchExternalWebBrowser("http\\..\\R2Northstar\\save_data\\Gecko.AwesomeMod\\my.bat", WEBBROWSER_FLAG_FORCEEXTERNAL)` would attempt to run `http\\..\\R2Northstar\\save_data\\Gecko.AwesomeMod\\my.bat` which given that `.bat` files are treated as executable on Windows, it would then run that batch script.

The same cannot happen with `.txt` as it's not treated as executable on Windows.

We patched `LaunchExternalWebBrowser()` ofc, so the exploit chain is already broken but we still want to consider preventing writing files that are treated as "executable" by default in Windows ^^

EladNLG:
> > The user has to MANUALLY open their titanfall 2 dir and go to R2Northstar/save_files/MOD.NAME/...
> 
> Not really. Before #674 `LaunchExternalWebBrowser()` had a bad check that (combined with Windows allowing to cd into and out of non-existing dirs), meant that `LaunchExternalWebBrowser("http\\..\\R2Northstar\\save_data\\Gecko.AwesomeMod\\my.bat", WEBBROWSER_FLAG_FORCEEXTERNAL)` would attempt to run `http\\..\\R2Northstar\\save_data\\Gecko.AwesomeMod\\my.bat` which given that `.bat` files are treated as executable on Windows, it would then run that batch script.
> 
> The same cannot happen with `.txt` as it's not treated as executable on Windows.
> 
> We patched `LaunchExternalWebBrowser()` ofc, so the exploit chain is already broken but we still want to consider preventing writing files that are treated as "executable" by default in Windows ^^

I think its still a problem that a mod can easily can do `LaunchExternalWebBrowser( "https://shadywebsite.com/NotAVirus.exe" )` and start a download whilst prompting the user to open the file.

GeckoEidechse:
> I think its still a problem that a mod can easily can do `LaunchExternalWebBrowser( "https://shadywebsite.com/NotAVirus.exe" )` and start a download whilst prompting the user to open the file.

Probably but at least that requires an interaction from the user. The other one is zero-click.

Nevertheless best to make an issue for this IMO so that we can keep track of it as it's unrelated to this PR at this point ^^

EladNLG:
> > Add `.json` as an exception.
> 
> I don't think it matters what extension the file has at the end, as long as it can't result in code execution by PEBKAC.
> 
> I do take issue with `.txt` being the default, the data being stored doesn't necessarily have to be plaintext. 
> `.dat` would probably be way better (though perhaps a whitelist of extensions with a default appended `.dat` would be even more better for developers)

The data, necessarily, has to be plaintext.

Jan200101:
> Literally how, safe IO works in strings from squirrel, so writing a binary format is just not gonna go well

despite that its not uncommon to use a string as a container for binary data.
I'm not sure exactly how Squrirel, let alone the squirrel in the game, handles this, but this is a common pattern in similar languages most notably Python and Lua.

ASpoonPlaysGames:
> despite that its not uncommon to use a string as a container for binary data. I'm not sure exactly how Squrirel, let alone the squirrel in the game, handles this, but this is a common pattern in similar languages most notably Python and Lua.

strings in squirrel are null terminated, so good luck making a decent binary format with that. There is also not really any need for binary formats, just encode things as json and parse once on load, you arent gonna be doing things like texture streaming through this come on now

EladNLG:
> > despite that its not uncommon to use a string as a container for binary data. I'm not sure exactly how Squrirel, let alone the squirrel in the game, handles this, but this is a common pattern in similar languages most notably Python and Lua.
> 
> strings in squirrel are null terminated, so good luck making a decent binary format with that. There is also not really any need for binary formats, just encode things as json and parse once on load, you arent gonna be doing things like texture streaming through this come on now

Fun fact:

They are [not null terminated](https://developer.electricimp.com/squirrel/squirrel-guide/variables-strings-blobs).

Jan200101:
> Despite that, you cannot store binary data through safe i/o since we restrict data to only consist of ASCII characters.

Do we?
The only code I can find that restricts a string to ascii is related to the path, which I think makes sense
But file content only has a null byte check (which may be wrong here since the function includes a comment that says its suppose to check the path?)

ASpoonPlaysGames:
There should be better ways of achieving the same goal.

GeckoEidechse:
> There should be better ways of achieving the same goal.

Like #682 :>

PR #679, opened 2024-03-08
Fix crash in silver-bun

### LINKED ISSUE #678: Silverbun crashes some users
Users are noticing a crash on 1.24.3 likely related to things loading too quickly. When using VanillaPlus (which loads a lot less than regular Northstar does), some users are noticing crashes. It was confirmed this was not the fault of VanillaPlus by using launch options `-northstar -profile=rfhius`, a garbage profile to load no mods, as to load faster and check that it was the speed of loading causing the issue. Using these launch options, users who crashed on VanillaPlus still crashed.

After some testing, they only experienced the crash when using the `Northstar.dll` including changes from the [Silverbun pull request](https://github.com/R2Northstar/NorthstarLauncher/pull/664) (tested by using the artifact of the silverbun pr, and the compiler warnings pr. Compilers warnings worked fine on its own, silverbun build artifact crashed)

Downgrading to `v1.24.2` is the current solution for users, but should be looked into having an actual solution in launcher

The specific crash message is here:
image

It should be noted only some users get this crash. I was unable to replicate it and am unsure why, no obvious differences in hardware between me and a few people who have tried, but I'd guess it's somehow related

Relevant Discord ticket where this was discovered (requires ticket viewing permission): https://discord.com/channels/920776187884732556/1215670795708399616

@lexvi on the Northstar Discord is a helper who notified me of this issue (and has the crash), so asking them to try things could also be helpful

ASpoonPlaysGames:
Closes #678 

Marking as draft until I can get some confirmation of the fix.

### DIFF
--- a/primedev/thirdparty/silver-bun/module.cpp
+++ b/primedev/thirdparty/silver-bun/module.cpp
@@ -15,7 +15,16 @@
 //-----------------------------------------------------------------------------
 CModule::CModule(HMODULE hModule)
 {
-	m_pModuleBase = reinterpret_cast<uintptr_t>(hModule);
+	MODULEINFO mInfo {0};
+
+	if (hModule && hModule != INVALID_HANDLE_VALUE)
+		GetModuleInformation(GetCurrentProcess(), hModule, &mInfo, sizeof(MODULEINFO));
+
+	m_nModuleSize = static_cast<size_t>(mInfo.SizeOfImage);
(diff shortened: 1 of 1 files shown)


ASpoonPlaysGames:
image
seems to work now

itscynxx (approved):
makes game not explode 👍 
(2 cases of users crashing are reporting this build working)

GeckoEidechse (approved):
Doesn't crash in quick testing with default profile and standard Northstar.

PR #408, opened 2023-01-29
Properly handle invalid cvar replications without blocking netmessage

BobTheBob9:
Reimplements `ns_server_name` and `ns_server_desc` replication without causing issues on clients that do not have the change locally, and prevents bad replicated cvars from breaking client connect

PR #680, opened 2024-03-12
Resolves the merge conflicts for #408

catornot:
hopefully everything still works.
it would be really cool if this pr and #408 would be merge soon :)


catornot:
maybe there is a better way than this 🤔

GeckoEidechse (approved):
Merges cleanly locally and even though the way GitHub shows it is a bit weird, the only real change is 19599b6c39a55d2fac10497be91af3bf4e332038

PR #681, opened 2024-03-14
I reverted things in #680 accidently when resolving conflicts so this is here to fix it

catornot:
yeah so when I was resolving conflicts I couldn't really tell what had to remain so some stuff got removed ( resolving conflicts sucks ). Then no one looked at it when merging😅
hopefully now everything is good


GeckoEidechse:
Yeah, the diff was so huge, it was basically impossible to notice the issue, haha.

Thanks for catching and fixing (for presumably good this time) now :D

GeckoEidechse (approved):
As far as I can tell the changes just fix some accidentally reverted changes in #408 

I did NOT test this.

PR #682, opened 2024-03-22
mods(Safe I/O): Only allow creating files with whitelisted filetypes

GeckoEidechse:
Only allow creating files with whitelisted filetypes to prevent writing `.bat` or `.vb` files with Safe I/O that could then be called using some exploit (e.g. #674), we should prevent writing file types that by default are interpreted as executable.

Whitelisting was chosen over blacklisting as we can always extend the list and don't have to worry about potentially forgetting to blacklist a certain filetype.

Of course this doesn't prevent writing a bash script to a `.txt` file and then somehow getting Windows to interpret it as a batch file.

Reading files is not restricted to filetypes as we are primarily concerned with creating files here.

Supersedes #675

~~Completely untested atm.~~

### Testing instructions:

1. Use the following mod: [GeckoEidechse.Safe-IO-Testing.zip](https://github.com/user-attachments/files/17837293/GeckoEidechse.Safe-IO-Testing.zip)
2. Go into private match and start a match
3. Enable cheats (`sv_cheats 1`)
4. Open console and run
    ```
    script GeckoSaveFile()
    ```
5. Go to `<profile folder>\save_data\GeckoEidechse.Safe-IO-Testing\`
6. Check that only `test1.txt` exists but no `test2.bat` or `test3`.


GeckoEidechse:
I wrote this on a system that is unable to compile Northstar and without having written C++ for like more than half a year, would be surprised if it even compiles xD

ASpoonPlaysGames:
> @ASpoonPlaysGames is also working on that (#675), are you aware of that? 😄

ngl that PR is kinda dead, i've just not had the time to do much lately. Honestly this PR might be preferable to mine.

primedev/mods/modsavefiles.cpp
```diff
@@ -74,6 +74,24 @@ template <ScriptContext context> void SaveFileManager::SaveFileAsync(fs::path fi
...
 		{
+			// Check if has extension and return early if not
+			if (!file.has_extension())
+			{
+				spdlog::error("SAVE FAILED!");
+				spdlog::error("No file extension specified");
+			}
+
+			// TODO: move into list of global consts?
+			std::set<std::string> whitelist = {".txt", ".json"};
```
catornot, line 85: Is there a reason for this being a `set` instead of like an `vector` or `array` or is it just an arbitrary choice?

Alystrasz:
@GeckoEidechse you created the branch on the main repo so we cannot update it :cry:

primedev/mods/modsavefiles.cpp
GeckoEidechse: Mostly arbitrary choice

catornot: Then it should probably be a vector.

barnabwhy: can probably also be made `static const`

GeckoEidechse: @barnabwhy done via 314b86aead1f3bd2e20840d493178b07113571eb

@catornot going back to this, wouldn't a set make more sense as soon as the list of extensions gets quite long? Like rn it's only 2 cause that's the two most important plaintext filetypes I could think of but the idea is that the list would get extended quite a bit later down the line.

catornot: how many filetypes will you need?

barnabwhy:
This isn't directly related to the PR but I noticed while reviewing that the code immediately below directly locks and unlocks the mutex, but ideally it should be using `std::lock_guard` at the start of the try catch block so that we don't need to explcitly unlock on every scope exit.

primedev/mods/modsavefiles.cpp
```diff
@@ -74,6 +74,24 @@ template <ScriptContext context> void SaveFileManager::SaveFileAsync(fs::path fi
 	std::thread writeThread(
 		[mutex, file, contents]()
 		{
+			// Check if has extension and return early if not
+			if (!file.has_extension())
+			{
+				spdlog::error("SAVE FAILED!");
```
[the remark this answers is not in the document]
GeckoEidechse: I just copied the existing error message string. That being said, fully agreed ^^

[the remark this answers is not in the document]
GeckoEidechse: Done in 4c791663ffc8443bc80a851efbcad824cd7a4e89

primedev/mods/modsavefiles.cpp
```diff
@@ -74,6 +74,24 @@ template <ScriptContext context> void SaveFileManager::SaveFileAsync(fs::path fi
...
+			}
+
+			// TODO: move into list of global consts?
+			static const std::set<std::string> whitelist = {".txt", ".json"};
+
+			// Check if file extension is whitelisted
+			std::string extension = file.extension().string();
+			if (whitelist.find(extension) == whitelist.end())
+			{
+				spdlog::error("SAVE FAILED!");
```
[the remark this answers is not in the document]
GeckoEidechse: Done in 4c791663ffc8443bc80a851efbcad824cd7a4e89

primedev/mods/modsavefiles.cpp
GeckoEidechse: idk

RoyalBlue1 (approved):
Code looks good I didn't test though.
If we plan on staying C++20 using `!whitelist.contains(extention)` might be more readable than `whitelist.find(extention) == whitelist.end()` but that is up to preference and shouldn't be blocking really

primedev/mods/modsavefiles.cpp
GeckoEidechse: Resolving as per https://github.com/R2Northstar/NorthstarLauncher/pull/682#pullrequestreview-2449702477

GeckoEidechse:
mfw when 3 people reviewed the PR and no one noticed that the return statements are missing in the check :clueless:

Alystrasz:
Code looks good to me, test mod works as expected in game:
image

Alystrasz (approved):
Code looks good, test mod works as expected:

image

PR #685, opened 2024-04-01
Depedency Constant Version Tables

uniboi:
```
#if HAS_UEBERBLICK
// This compiles only if the dependency is loaded. No matter the version of the dependency
#endif

#if HAS_UEBERBLICK && HAS_UEBERBLICK.major == 1
// This compiles ONLY if the dependency version is 1.x.x
#endif

#if HAS_UEBEBLICK && HAS_UEBERBLICK.major == 1 && HAS_UEBERBLICK.minor >= 2 && HAS_UEBERBLICK.minor <= 5
// This compiles only if the dependency's version is between 1.2.x - 1.5.x
#endif
```

- if a dependency is not loaded, the constant is `0`
- otherwise the dependency is a table that holds `major`, `minor` and `patch` keys for the versions of the mod
- only valid versions are inserted in the table. e.g. `1.x.x` would only expose the `major` key
- build & pre-release info is ignored

### DIFF
--- a/primedev/squirrel/squirrel.cpp
+++ b/primedev/squirrel/squirrel.cpp
@@ -10,6 +10,8 @@
 #include "plugins/pluginmanager.h"
 #include "ns_version.h"
 #include "core/vanilla.h"
+#include "squirrelclasstypes.h"
+#include "squirreldatatypes.h"
 
 #include <any>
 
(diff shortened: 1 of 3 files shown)


catornot (changes requested):
your semver parser crashed :|


patch is empty and minor has a `.`
image

no stack trace in logs 

mods that made it crash : [live holo](https://thunderstore.io/c/northstar/p/cat_or_not/LiveHolo/) & [holo mimic](https://thunderstore.io/c/northstar/p/cat_or_not/HoloMimic/)

it crashed at `CreateSlot("patch", stoi(patch));`  btw

GeckoEidechse:
I guess this needs a preliminary check if mod version is even properly formatted and fail out if not?

GeckoEidechse:
@uniboi bump
(especially cause merge conflicts atm)

GeckoEidechse:
Converting to draft until merge conflicts are addressed. cc @uniboi

PR #687, opened 2024-04-06
fix: Add vanilla compatibility for setplaylistvar byte-patch

### LINKED ISSUE #623: Private game settings don't apply (vanilla)
Reproducing:
Launch Northstar in vanilla compatibility mode (VanillaPlus is not required. Launching the game via steam with the launch options `-northstar -vanilla -profile=rhsdiudfkbjsd` is enough. The profile is gibberish to load no mods)
Launch a private match
Change the settings a bit
Launch the game
Settings are default

I found out about this due to an [issue on the VanillaPlus repo](https://github.com/Zayveeo5e/NP.VanillaPlus/issues/13) about this same thing. Thought I'd messed something up, but discovered launching with no mods still caused it to happen

My knowledge of C++ simply isn't enough to work on Northstar yet to figure out why this happens. I don't think it's related to the client starting a listen server over joining a Respawn server, as you can't run `sv_cheats` (or at least, to my knowledge. Running `sv_cheats 1`, then typing `sv_che` shows it autofill to `sv_cheats 0`, which afaik means it didn't work)

VITALISED:
This PR fixes #623 by adding a vanilla compatibility check for some NOPs in playlist.cpp

### DIFF
--- a/primedev/shared/playlist.cpp
+++ b/primedev/shared/playlist.cpp
@@ -122,6 +122,7 @@ ON_DLL_LOAD_RELIESON("engine.dll", PlaylistHooks, (ConCommand, ConVar), (CModule
 	module.Offset(0x18ED8D).Patch("C3");
 
 	// patch to allow setplaylistvaroverride to be called before map init on dedicated and private match launched through the game
+	// only run on non-vanilla as this patch makes users unable to change private match settings on vanilla
 	if (!g_pVanillaCompatibility->GetVanillaCompatibility())
 		module.Offset(0x18ED17).NOP(6);
 }
--- a/primedev/shared/playlist.cpp
+++ b/primedev/shared/playlist.cpp
@@ -122,6 +122,6 @@ ON_DLL_LOAD_RELIESON("engine.dll", PlaylistHooks, (ConCommand, ConVar), (CModule
 	module.Offset(0x18ED8D).Patch("C3");
 
 	// patch to allow setplaylistvaroverride to be called before map init on dedicated and private match launched through the game
-	if(!g_pVanillaCompatibility->GetVanillaCompatibility())
+	if (!g_pVanillaCompatibility->GetVanillaCompatibility())
 		module.Offset(0x18ED17).NOP(6);
 }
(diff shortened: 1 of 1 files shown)


r3muxd (approved):

r3muxd:
+1 lgtm

primedev/shared/playlist.cpp
```diff
@@ -121,5 +122,6 @@ ON_DLL_LOAD_RELIESON("engine.dll", PlaylistHooks, (ConCommand, ConVar), (CModule
 	module.Offset(0x18ED8D).Patch("C3");
 
 	// patch to allow setplaylistvaroverride to be called before map init on dedicated and private match launched through the game
```
itscynxx, line 124:
```suggestion
	// patch to allow setplaylistvaroverride to be called before map init on dedicated and private match launched through the game
	// only run on non-vanilla as this patch makes users unable to change private match settings on vanilla
```

itscynxx (changes requested):
This pr works as intended, and solves the issue I've been trying to with Northstar (where vanilla private match settings wont apply). 

No idea if it introduces any kind of security concern, someone else can confirm (e.g. does this allow vanilla users to change these variables when they aren't meant to be able to via console)

However, in contrary to your other prs this one does only work if vanilla and northstar are split. I am personally all for them being split still, but it's something to consider if you really wish to make them compatible

I just think the comment should be changed to describe what's happening better

GeckoEidechse:
> However, in contrary to your other prs this one does only work if vanilla and northstar are split.

I'm confused, you mean this (somehow) breaks behaviour in Northstar?

primedev/shared/playlist.cpp
```diff
@@ -121,5 +122,7 @@ ON_DLL_LOAD_RELIESON("engine.dll", PlaylistHooks, (ConCommand, ConVar), (CModule
...
-	module.Offset(0x18ED17).NOP(6);
+	// only run on non-vanilla as this patch makes users unable to change private match settings on vanilla
+	if (!g_pVanillaCompatibility->GetVanillaCompatibility())
```
ASpoonPlaysGames, line 126: would rather this byte patch was replaced with a function hook, so that in the case where vanilla compatibility becomes toggleable at runtime this will behave properly.

ASpoonPlaysGames (changes requested):

PR #688, opened 2024-04-06
feat(?): Allow Vanilla Compatibility Mode to Auth

VITALISED:
This PR makes some small changes to allow `-vanilla` users to authenticate to MS, because it is really stupid that they cannot frankly. (This is a counterpart to: https://github.com/R2Northstar/NorthstarMods/pull/800)

### DIFF
--- a/primedev/client/clientauthhooks.cpp
+++ b/primedev/client/clientauthhooks.cpp
@@ -22,7 +22,6 @@ void, __fastcall, (void* a1))
 	if (g_pVanillaCompatibility->GetVanillaCompatibility())
 	{
 		AuthWithStryder(a1);
-		return;
 	}
 
 	// game will call this forever, until it gets a valid auth key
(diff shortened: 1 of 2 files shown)


itscynxx:
For the record, as the "main" person pushing vanilla support at the moment, I don't think we have the dev strength to really keep up with genuinely adding vanilla and northstar compat side by side. I do technically have a local set of core mods that does this, however this is confined to my machine and I'm unsure if it actually brings about any issues

There are a lot of risks this introduces, and there are about 4 people in the world that care about fixing the issues it could cause

This pr also has unrelated changes in it. Why remove key invalidation? Does it break vanilla? If it does, then make it only run while not on vanilla?

VITALISED:
I think the thing with the key invalidation stuff was because I was faffing around with switching from Northstar->Vanilla and it wasn't getting any new tickets from Origin, probably need to tidy that up

GeckoEidechse:
Pretty sure not allowing vanilla mode to auth with Atlas was on purpose for a reason. cc @ASpoonPlaysGames ig ^^

ASpoonPlaysGames:
The entire reason behind not allowing them to auth is because in vanilla compatibility mode we also disable various things that we wouldn't want disabled on untrustworthy servers. Very intentional to not allow people using -vanilla to play on northstar servers.

Issue #634, opened 2024-01-07
-nojoy in dedi


p0358:
`-nojoy` should be added here to prevent loading XInput DLL on dedi (where it's not needed at all):
https://github.com/R2Northstar/NorthstarLauncher/blob/f98513d71f4d39bd8286ccf816d2603fb09035e6/primedev/dedicated/dedicated.cpp#L215-L222

I can't force myself to make a PR currently, so if someone feels like doing that, they can and do Co-authored-by me there xd But leaving it as issue regardless or it'll be forgotten.

PR #689, opened 2024-04-07
Adds -nojoy and -nosteamcontroller to dedicated cmdline args

Zonespace27:
Closes https://github.com/R2Northstar/NorthstarLauncher/issues/634

Adds `-nojoy` (joystick inputs) and `-nosteamcontroller` (steam controller inputs) to the dedicated server's default args. This is supposed to stop the loading of the XInput DLLs, but at the moment does not. This PR is drafted while I try to figure out if there's something more that I need to do.

image

### DIFF
--- a/primedev/dedicated/dedicated.cpp
+++ b/primedev/dedicated/dedicated.cpp
@@ -222,6 +222,8 @@ ON_DLL_LOAD_DEDI_RELIESON("engine.dll", DedicatedServer, ServerPresence, (CModul
 	CommandLine()->AppendParm("-nosound", 0);
 	CommandLine()->AppendParm("-windowed", 0);
 	CommandLine()->AppendParm("-nomessagebox", 0);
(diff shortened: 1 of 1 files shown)


r3muxd:
a.) "nosteamcontroller" does not exist as a string so ????????????
b.) "nojoy" does not stop xinput dll from being loaded, it just skips initializing joysticks as the name implies; the proper way to fix this is to hook and rebuild CInputSystem::Init or bytepatch

Zonespace27:
> a.) "nosteamcontroller" does not exist as a string so ????????????

Was going off of Valve documentation, I didn't consider that the steam controller was released after TF|2 stopped being updated


> b.) "nojoy" does not stop xinput dll from being loaded, it just skips initializing joysticks as the name implies; the proper way to fix this is to hook and rebuild `CInputSystem::Init` or bytepatch

Will look into

Alystrasz:
Closing as stale.

PR #691, opened 2024-04-24
Add "Mod Weapon Vars"

EladNLG:
## What?

Creates an alternative to weapon mods by allowing overrides of any _numeric_ eWeaponVar through script.

### Example

`ScriptWeaponVars_SetInt(weapon, eWeaponVar.ammo_clip_size, 50)`

## Why?

1. Weapon files need to be downloaded client side to work, and MAD isn't finished yet.
2. Weapon mods have a 200 "modifications" limit. This has been raised to ~240 through admirable work of others, but will not be enough for larger-scale mods/mods that want to apply a variety of modifiers on a variety of weapons.
3. Allows setting a variable to a range of values instead of being limited to 32 options (e.g. if you'd want to recreate the nemesis with the current weapon system, you wouldn't be able to because the amount of possible values for its burst delay is effectively infinite, however, this allows such things.)

WIP. Relies on

- #692

### Required before merging
- [x] Manager for script weapon modifiers
- [x] System to reapply overrides when weapon vars are recalculated (e.g. after `AddMod`) (Script PR not existing yet)

### Optional
~~- [ ] Auto-sync server changes to client through a remote function~~
~~- [ ] Add "predicted" versions of functions, that do not transmit to client~~
Decided against autotransmitting changes for a few reasons:
1) MAD
2) would be a pain and overcomplicate things imo

\[Screenshots/Examples TBD\]


EM4Volts:
while im not able to "review" this because i dont know anything in c++ or such at all i just want to say that his would be an excelent addition, the mod limit is big but still something that can happen. and having the evars modifyable directly via script enables alot more then the current need of switching mods on stuff.

primedev/client/weaponx.cpp
ASpoonPlaysGames, line 1: If there isn't anything in here, not much point in it existing

primedev/mods/modweaponvars.cpp
```diff
@@ -0,0 +1,103 @@
...
+template <ScriptContext context> WeaponVarInfo* weaponVarArray;
+
+AUTOHOOK_INIT()
+
+// TODO: _actually_ reverse this.
+// Seemingly recalculates weapon variables every tick, figure out what this actually does
+// rn we are "disabling" this function because it seems to make stuff
+// work, figure out why.
+// also ""slight"" performance increase
+AUTOHOOK(Cl_UpdateWeaponVars, client.dll + 0x5B2670, char, __fastcall, (void* a2))
```
ASpoonPlaysGames, line 18: Bit scary but sure

primedev/mods/modweaponvars.h
```diff
@@ -0,0 +1,22 @@
+enum WeaponVarType : __int8
+{
+	// support for other types isnt really needed
+	// and would probably allow for some fucking insane
+	// exploit chain somehow
+	INTEGER = (char)1,
+	FLOAT32 = (char)2
```
ASpoonPlaysGames, line 7: Do you actually need to cast these? wont the compiler implicitly convert them?

primedev/server/weaponx.h
```diff
@@ -0,0 +1,14 @@
...
+
+#pragma pack(push, 1)
+struct CWeaponX
+{
+	void* vftable;
+	const char gap_8[5128];
+	// this is used by the game as the initial offset
+	// for getting eWeaponVar values. No idea about
+	// the actual datatype. Start of another struct?
+	const char weaponVars[0x1000];
```
ASpoonPlaysGames, line 12: Does native do any checks to ensure that this is within some bounds? if so, you could probably figure out the exact size

primedev/mods/modweaponvars.cpp
```diff
@@ -0,0 +1,103 @@
...
+	{
+		// invalid type used
+		g_pSquirrel<context>->raiseerror(sqvm, "weaponVar type is not float!");
+		return SQRESULT_ERROR;
+	}
+
+	if (context == ScriptContext::SERVER)
+	{
+		CWeaponX* weapon = (CWeaponX*)ent;
+		*(float*)(&weapon->weaponVars[varInfo->offset]) = value;
```
catornot, line 81: should add a check here for if it's a weapon entity

primedev/mods/modweaponvars.cpp
```diff
@@ -0,0 +1,103 @@
...
+	}
+	else // if (context == ScriptContext::CLIENT)
+	{
+		C_WeaponX* weapon = (C_WeaponX*)ent;
+		*(float*)(&weapon->weaponVars[varInfo->offset]) = value;
```
catornot, line 86: same here

catornot (changes requested):

primedev/mods/modweaponvars.cpp
EladNLG: A better solution is to probably hook the function that actually recalculates mod values and do a codecallback after it

primedev/server/weaponx.h
EladNLG: could probably get the highest offset and then add 8 bytes to that, should be enough for our use case?

primedev/client/weaponx.cpp
EladNLG: kinda dont like having a header without a source counterpart, but sure

EladNLG:
PR finally works with prediction - gonna try cleaning up code tomorrow.

_Unfortunately_, i cannot figure out proper naming for the hooked functions - im just going off of frequency of calls :[

Please help if you can.

EladNLG:
Also, do we want to support strings/assets/vectors?

ASpoonPlaysGames:
> Also, do we want to support strings/assets/vectors?

If it's not that bad, yeah would be nice. Vectors probably the most useful

primedev/mods/modweaponvars.cpp
```diff
@@ -0,0 +1,255 @@
...
+	if (context == ScriptContext::SERVER)
+		return *ent == CWeaponX_vftable;
+	else
+		return *ent == C_WeaponX_vftable;
+}
+
+AUTOHOOK_INIT()
+
+// RENAME THIS BEFORE MERGE (REVIEWERS IF YOU STILL SEE THIS BLOCK IT NOWNOWNOWNOWNOW)
+AUTOHOOK(Cl_WeaponTick, client.dll + 0x59D1E0, bool, __fastcall, (C_WeaponX * weapon))
```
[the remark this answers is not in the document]
EladNLG: yeah, since we got rid of the other hook (we no longer need it) i can just call this Cl_WeaponTick...

the name might be wrong tho? ill be adding a comment for now

primedev/mods/modweaponvars.cpp
```diff
@@ -0,0 +1,311 @@
...
+	else // if (context == ScriptContext::CLIENT)
+	{
+		C_WeaponX* weapon = (C_WeaponX*)ent;
+		*(float*)(&weapon->weaponVars[varInfo->offset]) = value;
+	}
+
+	return SQRESULT_NULL;
+}
+
+ADD_SQFUNC("void", ModWeaponVars_GetType, "int eWeaponVar", "", ScriptContext::SERVER | ScriptContext::CLIENT)
```
catornot, line 158: this sqfunction should be marked with a int return type probably.

primedev/mods/modweaponvars.cpp
```diff
@@ -0,0 +1,311 @@
...
+	return result;
+}
+
+AUTOHOOK(Sv_CalcWeaponMods, server.dll + 0x6C8B80, bool, __fastcall, (int unk_0, char* unk_1, char* unk_2, bool unk_3, int unk_4))
+{
+	bool result = Sv_CalcWeaponMods(unk_0, unk_1, unk_2, unk_3, unk_4);
+
+	if (result && IsWeapon<ScriptContext::SERVER>((void**)(unk_2 - 0x1410)))
+	{
+		SQObject* entInstance = g_pSquirrel<ScriptContext::SERVER>->__sq_createscriptinstance((void**)(unk_2 - 0x1410));
```
catornot, line 73: these offsets into `unk_2` seam weird but they should still be a struct somehow imo.

primedev/mods/modweaponvars.cpp
EladNLG: oop my mistake, will fix tomorrow

primedev/mods/modweaponvars.cpp
catornot: atleast put the offset into a named constant.
or add the field to the `CWeaponX` class and get it's offset from that.

EladNLG: Its already in the weapon structs, will convert into an `offsetof` tomorrow

GeckoEidechse:
Uh, for some reason format check is failing now @EladNLG :c

EladNLG:
Need to change hook, currently fucks up stuff (causes weird MGL-related crash, probably fucked up params?)

ASpoonPlaysGames:
@EladNLG Ik it's been a while, has the stuff you've been working on moved beyond this PR? Is it still worth merging?

EladNLG:
@ASpoonPlaysGames if people want this feature ill try to get back and work on it, but otherwise yeah im focused on other stuff now

catornot:
as discussed a while back this does have a performance hit for larger player counts so this can't be merged without a feature gate at the very least. Also this feature is seemingly only needed in roguelike, and this could be just re-implemented in a plugin if someone else needs it.

PR #692, opened 2024-04-25
Add CreateScriptInstance

EladNLG:
Adds a function that converts entities to SQObjects to the SquirrelManagers.


primedev/squirrel/squirrel.h
```diff
@@ -447,7 +448,7 @@ inline VoidFunction SQMessageBufferPushArg(Vector3& arg) {
 // Vectors
 template <ScriptContext context>
 inline VoidFunction SQMessageBufferPushArg(SQObject* arg) {
-	return [arg]{ g_pSquirrel<context>->pushSQObject(g_pSquirrel<context>->m_pSQVM->sqvm, arg); };
+	return [arg]{ g_pSquirrel<context>->pushobject(g_pSquirrel<context>->m_pSQVM->sqvm, arg); };
```
catornot, line 451: `SQMessageBufferPushArg` should probably not be overloaded for `SQObject*` since some sqobjects are managed by the sqvm

catornot (approved):
looks good, checked the offsets in ghidra 👍 
didn't test in game.

EladNLG:
Ok, genuinely, how can this be even tested?

primedev/squirrel/squirrel.h
GeckoEidechse, line 451: I'm stupid, can you ELI5 this change to me? ^^"

GeckoEidechse:
> Ok, genuinely, how can this be even tested?

Usually the approach is to have a small PoC script mod that highlights the change. I don't understand enough atm to tell you whether that would make sense in this case ^^"

primedev/squirrel/squirrel.h
EladNLG: Somebody failed their spelling bee and the compiler for some reason didnt figure it out until I tried to call the function, and THEN it started complaining about `pushSQObject` not existing (because it is supposed to be `pushobject` instead.)

Basically, compiler is dumb.

EladNLG:
> > Ok, genuinely, how can this be even tested?
> 
> Usually the approach is to have a small PoC script mod that highlights the change. I don't understand enough atm to tell you whether that would make sense in this case ^^"

Unfortunately, no known part of the codebase (outside of #691) uses pushobject/createscriptinstance. My guess is that #691 will test this by proxy?

catornot:
I did test the server offsets in a plugin :p

primedev/squirrel/squirrel.h
GeckoEidechse: Copy that, fast tracking that fix o7

GeckoEidechse:
Maybe we just merge without further testing ig?

catornot:
> Maybe we just merge without further testing ig?

Yeah, just merge it.

GeckoEidechse:
Aight consider it merged. I might just gonna put a few more fixes in front to reduce the amount of cherry-picking I have to do ^^

catornot:
Wow this got finally merged

PR #695, opened 2024-05-04
Fix plugin with NULL strings causing UB

Jan200101:
string members were instances with the char* value before they were sanity checked for NULL.

`std::string(NULL)` is UB, but usually just results in a crash.

@uniboi because he is the author of plugins v4


ASpoonPlaysGames (approved):
Code looks good, not sure why we were using the pointer before checking it tbh

Jan200101:
`bugfix 👍` from uniboi

Jan200101:
I'd say no testing is required, this is a pretty straight forward fix that shouldn't impact any active Northstar users.

GeckoEidechse:
> I'd say no testing is required, this is a pretty straight forward fix that shouldn't impact any active Northstar users.

Agreed

PR #698, opened 2024-05-12
Make Script Errors from Northstar Callbacks Fatal

EladNLG:
Read Title.

## Why

Because:
- Errors are incredibly confusing when non-fatal - there is no indicator that an error is happening
- The error is unknown without a try/catch block
- Errors cannot be tracked to a line, or file, and its location has to be figured out manually
- They can throw a game into an unexpected state with no indicator, confusing both players AND developers


Jan200101:
here is the function definition for other reviewers:
```
SQRESULT sq_call(HSQUIRRELVM v, SQInteger params, SQBool retval, SQ-
Bool raiseerror);
```

code looks good, unsure if this will affect any existing mods (though if they are affected they already have an error so perhaps breaking already broken mods even more is best)

Jan200101:
played 30 minute of Attrition with a lot of mods installed that make use of scripts and thus far I've seen no problems.

I think its fine to merge as is but it would be cool if there an example mod that would silently error before this for the sake of testing.

GeckoEidechse:
100% agreed with this PR. I'll put it into `v1.26` release o7

GeckoEidechse (approved):
Signing off based on other reviews.

PR #699, opened 2024-05-21
Fix engine error crash caused by language detection hooks

wolf109909:
During my debugging around user's game constantly crashes on startup with file corruption detected, I found this code which does not satisfy the guarding logic and need further detection to guard against files named `general_stream_xxx.mstr`. After applying this code change in my NorthstarCN fork, the crashes are resolved.

The initial debugging procedures:
1. Console logs `Origin detected language "stream", but we do not have audio for it installed, falling back to the next option`
2. Games crashes with engine error: file corruption detected.
3. Found problematic code, applied the code change.
4. The problem is fixed without modifying the environment.

### DIFF
--- a/primedev/client/languagehooks.cpp
+++ b/primedev/client/languagehooks.cpp
@@ -41,7 +41,7 @@ std::vector<std::string> file_list(fs::path dir, std::regex ext_pattern)
 std::string GetAnyInstalledAudioLanguage()
 {
 	for (const auto& lang : file_list("r2\\sound\\", std::regex(".*?general_([a-z]+)_patch_1\\.mstr")))
-		if (lang != "general" || lang != "")
+		if (lang != "general" && lang != "" && lang != "stream")
 			return lang;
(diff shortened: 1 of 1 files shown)


GeckoEidechse (approved):
omfg, how did this stay unnoticed for so long x_x

Was [introduced](https://github.com/R2Northstar/NorthstarLauncher/pull/19/files#diff-b67efcd504e292e54c256a4ec7ec84317f76f78090987ce9b54856259ba7f5ccR47-R48) in #19 by accident

(For those that didn't have their morning coffee yet, previous version would always return `true` :P)

GeckoEidechse:
(Skipping testing cause the change is simple boolean logic and @wolf109909 mentioned they tested it (albeit on NorthstarCN so I might do this as a small patch release just to double check in prod)

wolf109909:
lol that's ancient!

wolf109909:
I'd say that this won't be noticed simply because `english` is before `stream` xD

PR #701, opened 2024-06-12
Set to specific MSVC version

GeckoEidechse:
This is due to some changes in `14.40` causing issues for some players

Closes #700 (hopefully)

Issue #700, opened 2024-06-12
Game fails to launch for some players


GeckoEidechse:
Gets stuck at

![image](https://github.com/R2Northstar/NorthstarLauncher/assets/40122905/4ae0969c-e679-48e4-b805-fe9bdee41a77)

GeckoEidechse:
After some investigating, it seems that builds compiled with MSVC at `14.39` work while `14.40` does not for affected players.

GeckoEidechse:
Possibly related https://github.com/actions/runner-images/issues/10020

GeckoEidechse:
Compiled binaries with two different podman containers that use either MSVC `14.39` or `14.40`
- Working binaries: [14.39.zip](https://github.com/user-attachments/files/15810214/14.39.zip)
- Broken binaries: [14.40.zip](https://github.com/user-attachments/files/15810221/14.40.zip)

GeckoEidechse:
Fixed by #704

PR #702, opened 2024-06-12
Disable dynamic runtime to prevent dependency issues

Jan200101:
possible alternative to #701


GeckoEidechse:
Will merge the moment someone that can reproduce #700 confirms it works for them o7

GeckoEidechse:
Didn't seem to fix it based on the results from a ticket on Northstar Discord

image

GeckoEidechse:
Another case of this patch unfortunately not working

screenshot

PR #705, opened 2024-06-14
Assign `CreateInterface` member instead of creating a new variable

Jan200101:
Before this change `m_pCreateInterface` was never assigned and thus undefined (and on wine Null initialized) so `Plugin::isValid` would always fail.


Jan200101:
Tested against CI build, works as expected.

For reference
Before:
image

After:
image

GeckoEidechse (approved):
Looks like #615 forgot to set this, whoopsie

Skipping testing cause the original bug is clear to me and I trust @Jan200101 to have done some basic testing ^^

PR #706, opened 2024-06-15
Don't instantiate templates before declaration, add missing instantiation

Jan200101:
In C++ explicit template instantiations need to be done AFTER template declarations are made.

Does not work:
```
template class SquirrelManager<ScriptContext::SERVER>;
template <ScriptContext context> class SquirrelManager;
```

Works:
```
template <ScriptContext context> class SquirrelManager;
template class SquirrelManager<ScriptContext::SERVER>;
```

### DIFF
--- a/primedev/squirrel/squirrel.cpp
+++ b/primedev/squirrel/squirrel.cpp
@@ -157,11 +157,6 @@ const char* SQTypeNameFromID(int type)
 	return "";
 }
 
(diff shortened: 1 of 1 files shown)


Klemmbaustein (approved):
This is just moving a few lines of code, I don't see any problems with this. And the code is more correct this way.
And it compiles and works for me.

GeckoEidechse (approved):
Signing off so we can merge based on the review by @Klemmbaustein

PR #707, opened 2024-06-15
Use double brackets to let compiler know assignment is intentional

Jan200101:
some compilers (most notably clang) gets a bit worried that we may be assigning unintentionally.
Using double brackets silences it and lets it know that this is correct.


GeckoEidechse (approved):
Compiler warnings related change that has no impact on actual logic. Given that assignment is intentional here, signalling it to the compiler is preferred.

Skipping testing due to nature of the change.

PR #708, opened 2024-06-15
Add overrides to `custom_sink_it_` methods

Jan200101:
the compiler knows we want to override here, since the original `custom_sink_it_` is virtual but we should be explicit to prevent any mistakes.


GeckoEidechse (approved):
Code compilation related change. If CI passes testing shouldn't be necessary.

I trust @Jan200101 with the implemented change.

PR #709, opened 2024-06-15
Remove invalid memset in ServerPresence

Jan200101:
The use of sizeof is incorrect here since `this` is a pointer and sizeof is used on the pointer directly, instead of what the pointer points to. It seems to work without issue due to padding but could cause issues if an instanced class like std::string gets nulled.


ASpoonPlaysGames (approved):
Not much to say, this code only worked before due to lucky padding. Memset isnt needed for this either

PR #710, opened 2024-06-15
Remove useless `this` pointer check

Jan200101:
The standard states that this must always be a valid pointer so these checks are optimized out anyway. Sane compilers, such as clang, also complain about this and state that this is pointless.


GeckoEidechse (approved):
Just gonna trust you on this one and merge without testing.

PR #711, opened 2024-06-15
Check if hook target is non `null` before calling MinHook

Jan200101:
MinHook doesn't check this either so it would also be found out when a Winapi function fails.

Lets fail earlier with an error message.


GeckoEidechse (approved):
Adds two checks if address is `null` and prints an error if they are. Not much to review here IMO and I trust @Jan200101 in the change. Skipping testing.

PR #712, opened 2024-06-15
Add missing case for switch cases in JSON deserialization function

Jan200101:
Add missing cases for switch cases.

RapidJSON switch gets a new case for null, where we do nothing. Future enum values would result in a new warning. (Might be good to add default case with error logging too?)


GeckoEidechse (approved):
Merging without testing due to the nature of the change.

PR #714, opened 2024-06-15
Remove duplicate wsock32 export

Jan200101:
Newer versions of lld check for duplicate export and error accordingly

```
lld-link: error: duplicate export ordinal: select
```


GeckoEidechse (approved):
Removes single duplicate line that was introduced in 33255f13d87be83d3d57fbecb73a54d38d70a5da011e5a71c482bd10a75392fb. Not much to review here. Skipping testing.

PR #716, opened 2024-06-16
Add default case for SquirrelContext switch

Jan200101:
logs errors so we can catch them if they do ever happen.

follow-up to #712


GeckoEidechse (approved):
The fact that C++ allows for omitting default case instead of forcing developer to specify if the default case should just be to do nothing, genuinely scares me (I miss Rust).

Anyway, change is IMO simple enough in nature that testing can be omitted.

Issue #715, opened 2024-06-15
Choice of JSON Library | Migrating from RapidJSON


Jan200101:
While working on #713 I found out that newer toolchains fail to build RapidJSON due to const-qualified types being kept const https://github.com/Tencent/rapidjson/issues/2277

With RapidJSON not having had a release in almost a decade, maybe we should switch to a better maintained library?

Needs to be discussed.

Jan200101:
Personally I think https://github.com/nlohmann/json would be a good replacement.
Its already considered the go-to C++ JSON library and would allow cmake integration.

Jan200101:
Its been mentioned on discord
RapidJSON is still in development however they stopped tagging commits.
I really hate this because it doesn't give us anchors to work with where one can say "This fix was included with Version XYZ".

We could continue cherry picking fixes or fixing stuff ourselves or move over to a library that respects the ecosystem and didn't just release it for luls like Tencent did.

Jan200101:
Digging through the chat log about previous discussions
```
[1:53 PM]Gecko: Does the nlohmann support [JSON5](https://json5.org/) ? ^^"

Cause rapidjson allows for comments in JSON which is part of JSON5 spec (but it also doesn't fully support JSON5 cause that would be too easy :dread:)

Basically, if planning to replace rapidjson, the replacement needs to be able to deal with comments (and maybe trailing commas) cause otherwise a bunch of mods will start to fail parsing due to comments in their mod.json...
```

Jan200101:
It was mentioned that one reason RapidJSON was chosen is performance.
Here are some alternatives which are high performance JSON Libraries which we can use in C++:
- [simdjson](https://github.com/simdjson/simdjson)
- [yyjson](https://github.com/ibireme/yyjson) 

Both of which consider themselves fasters than RapidJSON by a large margin.

GeckoEidechse:
Depending on how substantial changing JSON library would be, I guess we could treat RapidJSON as a rolling release and use something like [dependabot](https://docs.github.com/en/code-security/getting-started/dependabot-quickstart-guide) for informing us about new commits.
Of course I understand that this is not optimal so I'd love input from anyone who might have more experience on the matter ^^

ASpoonPlaysGames:
primedev moved over to nlohmann (or however it is spelt) and that seemed to go fine. Might be relatively trivial to port as well

Jan200101:
[...]

PR #717, opened 2024-06-16
Fix const-qualified assignment in RapidJSON

Jan200101:
pulls in https://github.com/Tencent/rapidjson/pull/719
resolves https://github.com/Tencent/rapidjson/issues/2277

Its  worth noting that the fix existed since 2016 and had not been pulled in yet.
Worth considering for #715


GeckoEidechse (approved):
Addresses a compiler issue so I'm taking the freedom in skipping testing.

PR #718, opened 2024-06-16
Use static MSVC Runtime Library

Jan200101:
We tried this already for when MSVC broke Northstar before but didn't merge it since it didn't fix the issue.

I think its best if we use the static runtime anyways to remove more dynamic dependencies that we can't anticipate.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,9 @@
 cmake_minimum_required(VERSION 3.15)
+cmake_policy(
+    SET
+    CMP0091
+    NEW
+    )
 
 project(
     Northstar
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -1,4 +1,9 @@
 cmake_minimum_required(VERSION 3.15)
+cmake_policy(
+    SET
+    CMP0091
+    NEW
+    )
```
GeckoEidechse, line 6: What does this do? ^^"

Jan200101: Cmake is backwards compatible even through breaking changes by seeing the requires cmake version and setting certain policies.

This policy is for how the runtime library flag is being resolved and defaults to OLD
we explicitly want the NEW behavior because the OLD behavior caches the values and could create potentially incorrect results.

See https://cmake.org/cmake/help/latest/policy/CMP0009.html

GeckoEidechse (approved):
Gonna test this one in prod, lol

GeckoEidechse:
Does increase binary size from ~3MB to ~10MB but given the size of the rest of our files (~90MB) this is acceptable IMO.

PR #720, opened 2024-06-17
Don't build when format check failed

Jan200101:
This will prevent needless CI running when the format is all wack

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,7 @@ env:
 jobs:
   build:
     runs-on: windows-2022
+    needs: [format-check, format-check-cmake-files]
     steps:
       - name: Checkout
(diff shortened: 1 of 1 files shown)


GeckoEidechse:
This is kinda on purpose so that we can easily test PR builds even if PR author managed to mess up formatting.

I'm fine with changing it though if desired ^^

Jan200101:
I guess?
but considering the nature of a pull request, it should have the right format already.
Idk, leaving this up to someone else to decide.

Alystrasz:
I agree with this change, let's not encourage authors not to format their code!

GeckoEidechse:
Just to be clear, PRs are already blocked from being merged until they are properly formatted. This change would just block PRs from being tested while they are still improperly formatted ^^

PR #721, opened 2024-06-17
Add clang CI

Jan200101:
THIS ONLY APPLIES TO THE CI, NOT RELEASES

Builds Northstar with the Microsoft provided clang-cl build.

Anything that downloads GH Actions artifacts like FlightCore (@GeckoEidechse) will need to be updated.

Alternative I could check for a way of only having a `LLVM` suffix for clang builds.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,6 @@ env:
 jobs:
   build:
     runs-on: windows-2022
-    needs: [format-check, format-check-cmake-files]
     strategy:
       matrix:
         config:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,15 @@ env:
 jobs:
   build:
     runs-on: windows-2022
+    needs: [format-check, format-check-cmake-files]
+    strategy:
+      matrix:
+        config:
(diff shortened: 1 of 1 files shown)


Jan200101:
This PR would also need the required build to be changed

GeckoEidechse:
> Anything that downloads GH Actions artifacts like FlightCore (@GeckoEidechse) will need to be updated.

FlightCore is the only application I'm aware of that downloads builds from CI. Will keep in mind o7

.github/workflows/ci.yml
```diff
@@ -8,6 +8,15 @@ env:
 jobs:
   build:
     runs-on: windows-2022
+    needs: [format-check, format-check-cmake-files]
+    strategy:
```
GeckoEidechse, line 12:
```suggestion
    runs-on: windows-2022
    strategy:
```

Trying to sneak in #720? Hehe :>

GeckoEidechse (approved):
This just adds an additional CI run that compiles the project with clang next to MSVC

This does not affect release builds. The only change is that MSVC CI builds from push have a name change in the produced zip folder. FlightCore has [already been updated to handle this](https://github.com/R2NorthstarTools/FlightCore/pull/949) so this PR is good to merge ^^

PR #722, opened 2024-06-17
Update all actions to supported versions

Jan200101:
checkout v3 has been out of date for a while
artifact v1, v2 and v3 will be deprecated this year, so we have to switch.


GeckoEidechse:
I split up the v4 update of `actions/checkout` and `actions/upload-artifact` into separate PRs (#724 and #725). In regards to the add-to-project workflow I was thinking to just do an org wide action and then put it there instead of having that workflow added to each repo separately ^^

PR #725, opened 2024-06-19
Update `actions/upload-artifact` to v4

GeckoEidechse:
v3 is outdated and will soon be deprecated

Spliced from #722


GeckoEidechse (commented):
Artifacts are uploaded correctly and FlightCore can download and install them as well so I consider this tested and good to merge ^^

PR #726, opened 2024-06-20
Lowercase all linked libraries for building on Linux

Jan200101:
Microsoft, in their infinite wisdom, decided to suffix some libraries with `.Lib` instead of `.lib` This causes issues with cmake on Linux because it only looks for `.lib` which it won't be able to resolve because the file system is case sensitive. Microsoft does this for backwards compatibility, in cmake this is a limitation so the best solution is to bite the bullet and lowercase all libraries which setups such as wine-msvc and xwin already do.

### DIFF
--- a/primedev/Launcher.cmake
+++ b/primedev/Launcher.cmake
@@ -19,7 +19,7 @@ target_link_libraries(
             uuid.lib
             odbc32.lib
             odbccp32.lib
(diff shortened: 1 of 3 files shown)


GeckoEidechse (approved):
Change only affects building Northstar. Given that CI built Northstar just fine still, this is good to merge.

PR #727, opened 2024-06-22
Restructure primelauncher cmake logic

Jan200101:
- moves the primelauncher cmake logic into its subdirectory allowing everything related to it to be self contained


Jan200101:
If the cmake format check is about to complain, I will scream

Jan200101:
reverted GUI subsystem stuff since it would require more work that I didn't think about.
It will be made into a new PR

GeckoEidechse (approved):
Change in build system, can skip testing as CI passes.

PR #728, opened 2024-06-22
Fix memalloc problems

Jan200101:
#718 stopped Northstar from working for two reasons:
- the way the *alloc family was declared was not compatible with how the CRT worked
- _recalloc_base was declared but never defined

Microsofts documentation on _recalloc is a bit bad so I followed the chromium implementation
https://chromium.googlesource.com/chromium/src/+/84db13d2045c3bc2753c4048ec1816373de0c3cc/base/allocator/allocator_shim_override_ucrt_symbols_win.h#102

If the chromium implementation is right and only the new bits are zeroed then the r5sdk implementation is wrong.

Since we need to get the old size for recalloc I also implemented _msize

Tested this myself on Windows but I would love to get another tester for this

### DIFF
--- a/primedev/core/memalloc.cpp
+++ b/primedev/core/memalloc.cpp
@@ -3,7 +3,7 @@
 
 // TODO: rename to malloc and free after removing statically compiled .libs
 
-extern "C" void* _malloc_base(size_t n)
+void* _malloc_base(size_t n)
 {
 	// allocate into static buffer if g_pMemAllocSingleton isn't initialised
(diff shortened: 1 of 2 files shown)


Jan200101:
> seems to work fine

Klemmbaustein

Klemmbaustein (approved):
The code looks good and this works fine for me

Klemmbaustein (approved):

GeckoEidechse (approved):
Signing off based on other reviews

PR #729, opened 2024-06-22
Convert primelauncher to GUI Subsystem

Jan200101:
Converts primelauncher to GUI subsystem.
Since the GUI subsystem NEEDS the entrypoint to be WinMain I made it call main, since large portions of the code rely on argc and argv handling.

This is practically how Microsoft implements this
`__argc` and `__argv` are available globals
https://learn.microsoft.com/en-us/cpp/c-runtime-library/argc-argv-wargv?view=msvc-170&redirectedfrom=MSDN

### DIFF
--- a/primedev/primelauncher/CMakeLists.txt
+++ b/primedev/primelauncher/CMakeLists.txt
@@ -1,6 +1,6 @@
 # NorthstarLauncher
 
-add_executable(NorthstarLauncher "main.cpp" "resources.rc")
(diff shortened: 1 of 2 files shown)


GeckoEidechse (approved):
Works in testing when launching from FlightCore.

However, when launching from the `r2ds.bat` in its current state it spawns two windows.
Would need to update it to `start NorthstarLauncher.exe --dedicated --multiple`

Jan200101:
Would cause issues with the Server Docker since it it expects Northstar to be a GUI app

PR #730, opened 2024-06-23
add debug alloc definitions

Jan200101:
Making debug builds currently fails HARD because it attempts to do a lot of bad stuff.

Northstar won't run with this PR, but I'm working on fixing that.


Jan200101:
building works and we are no longer trapped by asserts
however now the game freezes upon entering the actual engine loop

PR #731, opened 2024-06-23
Add missing object check in mod manager

Jan200101:
`HasMember` asserts `IsObject()` internally.
In release builds this is not an issue but this does not work on Debug builds.

`m_bHasEnabledModsCfg` is the output of `m_EnabledModsCfg.IsObject()` and this pattern exists in other places, so I think this is a "good enough" solution at the moment (though I consider this file to smell a bit)


GeckoEidechse (approved):
Based on the PR description I think it's fine to skip testing here.

PR #732, opened 2024-06-25
correct includes everywhere

Jan200101:
lots of headers have dependencies that are fulfilled through the precompiled header either directly or indirectly.

For the sake of code clarity translation unit dependencies should be properly marked.


GeckoEidechse (changes requested):
So there's a few changes in there as well such as the removal of a few cases of `NOTE_UNUSED( );`, changes of `assert_msg` to `assert()`, as well as introduction of `[[maybe_unused]]` and given my limited C++ knowledge I'm trying to figure out how that relates to the core change of the PR which is changing the `includes` ^^"

Jan200101:
> So there's a few changes in there as well such as the removal of a few cases of `NOTE_UNUSED( );`, changes of `assert_msg` to `assert()`, as well as introduction of `[[maybe_unused]]` and given my limited C++ knowledge I'm trying to figure out how that relates to the core change of the PR which is changing the `includes` ^^"

those were extra changes that I forgot to remove
Will do so ASAP

primedev/thirdparty/silver-bun/module.cpp
```diff
@@ -204,7 +204,7 @@ CMemory CModule::FindString(const char* szString, const ptrdiff_t nOccurrence, b
 
 	for (size_t i = 0ull; i < m_ExecutableCode.m_nSectionSize - 0x5; i++)
 	{
-		byte byte = pTextStart[i];
+		BYTE byte = pTextStart[i];
```
GeckoEidechse, line 207: Unrelated change. Probably left over by accident?

primedev/core/convar/cvar.h
```diff
@@ -1,10 +1,11 @@
 #pragma once
 #include "convar.h"
+#include "concommand.h"
+#include "core/macros.h"
 
 //-----------------------------------------------------------------------------
 // Forward declarations
 //-----------------------------------------------------------------------------
-class ConCommandBase;
```
GeckoEidechse, line 7: I assume that class is unused and hence removed? Separate PR maybe? ^^

primedev/shared/keyvalues.cpp
```diff
@@ -1,5 +1,7 @@
 #include "keyvalues.h"
-#include <winnt.h>
```
GeckoEidechse, line 2: I assumed `winnt.h` was actually unused and hence removed?

primedev/client/audio.cpp
```diff
@@ -49,7 +54,7 @@ EventOverrideData::EventOverrideData(const std::string& data, const fs::path& pa
 	}
 
 	rapidjson_document dataJson;
-	dataJson.Parse<rapidjson::ParseFlag::kParseCommentsFlag | rapidjson::ParseFlag::kParseTrailingCommasFlag>(data);
+	dataJson.Parse<(rapidjson::ParseFlag::kParseCommentsFlag | rapidjson::ParseFlag::kParseTrailingCommasFlag)>(data);
```
GeckoEidechse, line 57: Extra brackets added. Is there a reason to add them? Genuinely curious ^^

primedev/thirdparty/silver-bun/module.cpp
Jan200101: nope.
I tested this PR by emptying the precompiled header and without it the definition of `byte` is missing here and the compilation would fail.
Upstream was changed a lot so and BYTE makes sense here so I believe this is the right solution.

primedev/core/convar/cvar.h
Jan200101: the issue I had at the time is that `ConCommandBase` was used directly but I can't spot that from a glance.
Will have to test if this is actually needed.

primedev/shared/keyvalues.cpp
Jan200101: opposite actually
`winnt.h` cannot work without including some other stuff first.
In this case switching to `windows.h` worked just fine.

primedev/client/audio.cpp
Jan200101: At the time the compiler threw an error about this, but that could have been a premature error due to a faulty type definition somewhere.
Will test if this can be reverted.

primedev/thirdparty/silver-bun/module.cpp
GeckoEidechse: Huh, very curious.

Alystrasz (approved):
I guess clang would have raised warning if added imports were useless?
Anyway, beside Gecko's comments, code looks fine, and works in testing.

primedev/client/audio.cpp
Jan200101: turns out, this was caused by `RAPIDJSON_HAS_STDSTRING` being unset
will revert

primedev/core/convar/cvar.h
Jan200101: reverted

primedev/thirdparty/silver-bun/module.cpp
Jan200101: found it
its an internal windows typedef that i was not previously aware of
`typedef byte BYTE`

reverted

PR #733, opened 2024-06-25
remove needless parts out of precompiled header

Jan200101:
needs #732 
lots of things in the precompiled header really do not benefit from being in there since they are rarely used or not that large.

locally this has sped up the compiled by a minimum of 5 seconds


Jan200101:
will be split up into other PRs

PR #735, opened 2024-06-26
Replace `NOTE_UNUSED` with standard `[[maybe_unused]]` attribute

Jan200101:
Replaces the `NOTE_UNUSED` macro with the standardized `[[maybe_unused]]` attribute.


GeckoEidechse:
Tagging @ASpoonPlaysGames who introduced `NOTE_UNUSED()`. This might be of interest to you. I only learned about this with this PR too :eyes:

ASpoonPlaysGames (commented):
A lot of these variables are always unused. Using  `maybe_unused` for them seems incorrect. 

`maybe_unused` afaik is meant for variables that are used under some build configurations, but not all.


I think I maybe prefer `int /*name*/` for always unused parameters, and `maybe_unused` for conditionally used ones? Open to suggestions though

Jan200101:
this is nothing more than `NOTE_UNUSED` but as an attribute instead of within the code to give the compiler a hint.

ASpoonPlaysGames:
Functionally, yes. But using the standard attribute means that we should really follow best practice with it

PR #736, opened 2024-06-26
replace assert_msg macro with plain assert

Jan200101:
the `assert_msg` macro really didn't do a lot and was only used in one place so..... yeah.
Less code smell!


Jan200101:
Format check fails without #735

PR #738, opened 2024-06-26
Remove EXPORT macro from precompiled header

Jan200101:
its used once and does not warrant being in the precompiled header


GeckoEidechse (approved):
Single compilation related change. Skipping testing as CI passed.

PR #739, opened 2024-06-26
Remove precompiled header from wsock proxy

Jan200101:
The wsock proxy is so small and self contained that it really does not benefit from this.


GeckoEidechse (approved):
Compilation related change. Passes CI so good to merge for me. Skipping testing.

PR #741, opened 2024-06-26
Check for Console Window before allocating one

Jan200101:
gets rid of that annoying
`[*] Failed to create a console window, maybe a console already exists?`
when launching via NorthstarLauncher.exe


catornot (approved):
very simple change :+1:
no more `failed to create a console window` :))))

GeckoEidechse:
Merging based on review

PR #742, opened 2024-06-26
Convert not too important logging to debug

Jan200101:
Hey

you know how Northstar spews out over 500 log messages on startup?
you know how most of those just log the internal state?
Why do we display them as info?
With this PR I propose to turn them into debug logs and enable the output of debug logs by default for the time being.

Afterwards we can add flags or cvars to control what information is output where so debug output isn't spammed into the console for example.


Alystrasz (approved):
Works (obviously) in testing.
We'll have to be careful when hiding debug messages though, as they are valuable when debugging someone's client.

Jan200101:
Agreed, this is just step 1.

The idea here is to turn everthing not useful to a user into a debug message.
From there we can filter them when we output e.g. the in game console but still log. everthing in a file

PR #744, opened 2024-06-27
Add ccache integration

Jan200101:
I mean, why not, right?

ccache supports msvc-like compilers such as msvc or clang-cl and it beats needless recompilation.

Pretty poggers if I may say so.

:+1:


Jan200101:
I want to *kill* whoever decided on this fucking cmake format checker because it is very terrible and makes everything butt ugly.

PR #747, opened 2024-07-06
Define `fs` alias in source file where used

GeckoEidechse:
Define `fs` alias for filesystem namespace in source file where used instead of relying on implicit includes.

Spliced from #732


GeckoEidechse (commented):
Compilation related change so no testing needed IMO.

PR #642, opened 2024-01-19
Adjust for restructured `verified-mods.json`

GeckoEidechse:
Updates the launcher code to deal with adjusted verified mods JSON structure defined in
- https://github.com/R2Northstar/VerifiedMods/pull/23

### Testing:

To test, run Northstar with:
```
.\NorthstarLauncher.exe -customverifiedurl=https://raw.githubusercontent.com/R2Northstar/VerifiedMods/f48c43b980490fd80761d967866e16cfd0c793ab/verified-mods.json
```

and join any server with a MAD mod.

PR #748, opened 2024-07-08
Adjust for restructured `verified-mods.json`

Alystrasz:
Updates the launcher code to deal with adjusted verified mods JSON structure defined in https://github.com/R2Northstar/VerifiedMods/pull/41.

The idea here is to allow installing mods from other sources than Thunderstore.

(Somehow) supercedes https://github.com/R2Northstar/NorthstarLauncher/pull/642.

**(should be merged with https://github.com/R2Northstar/VerifiedMods/pull/41!)**

#### TODOs

- [x] Update manifesto format checking
- [x] Wait until manifesto fetching is effectively protected by the MAD flag (https://github.com/R2Northstar/NorthstarLauncher/pull/751)

#### Testing

1. To make your client use the verified mods list's new format, launch Northstar like this: `.\NorthstarLauncher.exe -customverifiedurl=https://raw.githubusercontent.com/R2Northstar/VerifiedMods/refactor/include-mod-source-in-version-links/verified-mods.json`
2. Join mp lobby;
3. Set `allow_mod_auto_download` to 1 through console (`allow_mod_auto_download 1`);
4. `sv_cheats 1`;
5. Fetch manifesto: `script NSFetchVerifiedModsManifesto()`;
6. Download mod: `script NSDownloadMod("nyami.mp_brick", "1.0.2")`;

Looking at your logs, mod should be downloaded normally.


GeckoEidechse:
With https://github.com/R2Northstar/VerifiedMods/pull/39 merged, pulling verified mods from custom source is no longer necessary to test this PR ^^

GeckoEidechse:
bumping so it shows up in the webhook channel and hopefully someone tests it :>

GeckoEidechse:
Quick question until I get around testing this, what happens when we merge this PR but not https://github.com/R2Northstar/VerifiedMods/pull/41 ?

Does client joining a server that requires verified mods crash or not?

Cause with #749 and #751 merged and released only recently, there's still clients that run older versions as the fetch happens before the version gate check, so even if a client would fail version gate, they'd crash before that already due to trying to fetch invalid manifest.
Hence I'm wondering if we should give https://github.com/R2Northstar/VerifiedMods/pull/41 some more time but at the same time already prepare client for new structure by merging this PR as soon as it's tested/reviewed ^^

Alystrasz:
> Quick question until I get around testing this, what happens when we merge this PR but not [R2Northstar/VerifiedMods#41](https://github.com/R2Northstar/VerifiedMods/pull/41) ?
> 
> Does client joining a server that requires verified mods crash or not?
> 
> Cause with #749 and #751 merged and released only recently, there's still clients that run older versions as the fetch happens before the version gate check, so even if a client would fail version gate, they'd crash before that already due to trying to fetch invalid manifest. Hence I'm wondering if we should give [R2Northstar/VerifiedMods#41](https://github.com/R2Northstar/VerifiedMods/pull/41) some more time but at the same time already prepare client for new structure by merging this PR as soon as it's tested/reviewed ^^

If we merge the launcher part but not the VerifiedMods counterpart, client won't crash but manifesto parsing will fail (without crashing with #749), thus preventing all mods from being downloaded.

GeckoEidechse:
> If we merge the launcher part but not the VerifiedMods counterpart, client won't crash but manifesto parsing will fail (without crashing with #749), thus preventing all mods from being downloaded.

Ok that's what I was hoping for and would be temporarily be acceptable as MAD is behind a feature flag anyway. This way we can push out this PR even as part of a patch release as the effect is not noticeable by clients under normal circumstances and for dev stuff we can just pass the CLI flag to change the manifest location to that of the PR.

GeckoEidechse (changes requested):
Following the instructions, I get "Tried to download a mod that is not verified, aborting" in console.

Do note that this happens with every combination of using the old and new manifesto and using this PR and most recent release.
So most likely the issue is not in the PR itself but maybe the instructions or something else.

Alystrasz:
> Following the instructions, I get "Tried to download a mod that is not verified, aborting" in console.

So this is due to the local verified mods manifest which is empty (due to `NSDownloadMod` not fetching manifest from GitHub, woopsies); `NSFetchVerifiedModsManifesto` must be invoked before installing mods.
Testing instructions were updated accordingly.

GeckoEidechse (approved):
Confirmed working in testing :D

image

ASpoonPlaysGames (approved):
Code looks good :)

PR #750, opened 2024-07-10
Fix typo in class member function

GeckoEidechse:
`pushSQObject` does not exist. Spliced from #692 as per https://github.com/R2Northstar/NorthstarLauncher/pull/692#discussion_r1591169750


GeckoEidechse (commented):
CI passes, compilation related change so this can be merged without further testing.

PR #759, opened 2024-08-08
Add comment explaining time choice for cron job

GeckoEidechse:
Just adds a comment to CI file explaining the reasoning for choosing a specific time for the cron job.


GeckoEidechse:
No usual type of review needed as just adds a comment to the CI config file and is not something NorthstarLauncher related.

PR #760, opened 2024-08-08
Add option to manually run merge conflict label action

GeckoEidechse:
so that labels can easily be updated if need be.


GeckoEidechse:
No usual type of review needed as this modifies logic about the repo itself and not something NorthstarMods related.

Issue #690, opened 2024-04-15
Northstar dll crash when not supplying atleast semi valid rpak.json


EM4Volts:
To reproduce:
put rpaks in rpaks folder of mod, don't put json file in.

can be circumvented with a `rpak.json` file that isn't even valid like:

```
{
    "Preload":
    {
        "example.rpak": false
    }
}
```

(yes, it doesn't care about if that rpak even exists, putting this in resolves the crash)

EM4Volts:
henlo frens! i would like to request to reserve this issue for myself, id like to try and fix this if possible :) 

was looking at the code rn because this issue once again happened to me today on accident and it seems like the fix could be simple.

GeckoEidechse:
Closed by #762

PR #762, opened 2024-08-09
Fix postload crash from missing `rpak.json`

EM4Volts:
Hello frens!

This pr fixes issue #690 

To reproduce the crash on a version without the fix:

1. make a mod with a paks folder
2. put atleast an rpak into it
3. dont make the rpaks.json or make it empty
4. launch the game

this should now crash at line 878 in modmanager.cpp as the original logic tries to set a postload string even tho none exists as the variable bUseRpakJson is being inverted in the if statement and doest ask for an && wether the other statements are also true, thus leading to the string being set to a value that doesnt exist making launcher go brrrt

To Test wether it works:
follow the reproduce steps, and notice that it wont crash and still load rpaks properly.

### DIFF
--- a/primedev/mods/modmanager.cpp
+++ b/primedev/mods/modmanager.cpp
@@ -872,7 +872,7 @@ void ModManager::LoadMods()
 
 					if (!bUseRpakJson)
 					{
-						spdlog::warn("Mod {} contains rpaks without valid rpaks.json, rpaks might not be loaded", mod.Name);
+						spdlog::warn("Mod {} contains rpaks without valid rpak.json, rpaks might not be loaded", mod.Name);
 					}
 					else
 					{
--- a/primedev/mods/modmanager.cpp
+++ b/primedev/mods/modmanager.cpp
@@ -863,19 +863,27 @@ void ModManager::LoadMods()
 			for (fs::directory_entry file : fs::directory_iterator(mod.m_ModDirectory / "paks"))
 			{
 				// ensure we're only loading rpaks
-				if (fs::is_regular_file(file) && file.path().extension() == ".rpak" && bUseRpakJson)
+				if (fs::is_regular_file(file) && file.path().extension() == ".rpak")
 				{
+
+
 					std::string pakName(file.path().filename().string());
-
 					ModRpakEntry& modPak = mod.Rpaks.emplace_back();
-					modPak.m_bAutoLoad =
-						!bUseRpakJson || (dRpakJson.HasMember("Preload") && dRpakJson["Preload"].IsObject() &&
-										  dRpakJson["Preload"].HasMember(pakName) && dRpakJson["Preload"][pakName].IsTrue());
 
-					// postload things
-					if (bUseRpakJson &&
-						(dRpakJson.HasMember("Postload") && dRpakJson["Postload"].IsObject() && dRpakJson["Postload"].HasMember(pakName)))
-						modPak.m_sLoadAfterPak = dRpakJson["Postload"][pakName].GetString();
+					if (!bUseRpakJson)
+					{
+						spdlog::warn("Mod {} contains rpak(s) without a valid rpaks.json, rpak(s) might not be loaded", mod.Name);
+					}
+					else
+					{
+						modPak.m_bAutoLoad = (dRpakJson.HasMember("Preload") && dRpakJson["Preload"].IsObject() &&
+							dRpakJson["Preload"].HasMember(pakName) && dRpakJson["Preload"][pakName].IsTrue());
+
+						// postload things
+						if (dRpakJson.HasMember("Postload") && dRpakJson["Postload"].IsObject() &&
+							dRpakJson["Postload"].HasMember(pakName))
+							modPak.m_sLoadAfterPak = dRpakJson["Postload"][pakName].GetString();
+					}
 
 					modPak.m_sPakName = pakName;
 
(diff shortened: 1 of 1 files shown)


ASpoonPlaysGames (approved):
Can confirm that the logic change makes sense, in fact I believe that I wrote the original broken logic ages ago :(

GeckoEidechse:
Any chance you got a mod on hand to test this with? Or link to Thunderstore? Then I could test rn and merge

EM4Volts:
> Any chance you got a mod on hand to test this with? Or link to Thunderstore? Then I could test rn and merge

https://thunderstore.io/c/northstar/p/S2Mods/ResonanceRifle/

should work. just go ingame with it and check if its fine ingame.
then restart but with a deleted paks/rpak.json and ingame it should load but not have materials (source error texture)

GeckoEidechse:
Ok so on Linux (Fedora 40) I get an infinite loading screen with missing `rpak.json` (the one before main menu with the bt loading icon in the bottom right) while with the `rpak.json`  the game loads fine (mod enabled in both cases).

Maybe someone can check on Windows rn?

EM4Volts:
> Ok so on Linux (Fedora 40) I get an infinite loading screen with missing `rpak.json` (the one before main menu with the bt loading icon in the bottom right) while with the `rpak.json` the game loads fine (mod enabled in both cases).
> 
> Maybe someone can check on Windows rn?

just rechecked myself and it seems so :/ couldve sworn it worked before. i however think i know how to fix this, will push update in a second

EM4Volts:
yeah i think i know the problem now. dont merge this even with the new v2 fix thing please. this currently now works the way i thought it would but no longer works on stuff like maps with rpaks.

for example, above isntructions work and goes ingame but no longer actually registers the paks if no json exists, as such maps cant just deliver the rpaks without an json.

i need to look a bit deeper into how the rpak stuff works in launcher ^^

EM4Volts:
ok, published the final fix now. this should work the exact way it should work.
while also allowing maps to not need the json while still registering starpaks if they exist so they can work by the game loading thoose paks.

testing is still the same

primedev/mods/modmanager.cpp
```diff
@@ -865,17 +865,25 @@ void ModManager::LoadMods()
...
-										  dRpakJson["Preload"].HasMember(pakName) && dRpakJson["Preload"][pakName].IsTrue());
-
-					// postload things
-					if (!bUseRpakJson ||
-						(dRpakJson.HasMember("Postload") && dRpakJson["Postload"].IsObject() && dRpakJson["Postload"].HasMember(pakName)))
-						modPak.m_sLoadAfterPak = dRpakJson["Postload"][pakName].GetString();
+
+					if (!bUseRpakJson)
+					{
+						spdlog::warn("Mod {} contains rpaks without valid rpaks.json, rpaks might not be loaded", mod.Name);
```
ASpoonPlaysGames, line 875:
```suggestion
						spdlog::warn("Mod {} contains rpaks without valid rpak.json, rpaks might not be loaded", mod.Name);
```

primedev/mods/modmanager.cpp
```diff
@@ -865,17 +865,25 @@ void ModManager::LoadMods()
 				// ensure we're only loading rpaks
 				if (fs::is_regular_file(file) && file.path().extension() == ".rpak")
 				{
-					std::string pakName(file.path().filename().string());
 
+
+					std::string pakName(file.path().filename().string());
```
ASpoonPlaysGames, line 870: Accidental moving of this line?

primedev/mods/modmanager.cpp
```diff
@@ -865,17 +865,25 @@ void ModManager::LoadMods()
...
+					}
+					else
+					{
+						modPak.m_bAutoLoad = (dRpakJson.HasMember("Preload") && dRpakJson["Preload"].IsObject() &&
+							dRpakJson["Preload"].HasMember(pakName) && dRpakJson["Preload"][pakName].IsTrue());
+
+						// postload things
+						if (dRpakJson.HasMember("Postload") && dRpakJson["Postload"].IsObject() &&
+							dRpakJson["Postload"].HasMember(pakName))
+							modPak.m_sLoadAfterPak = dRpakJson["Postload"][pakName].GetString();
```
ASpoonPlaysGames, line 885:
```suggestion
						if (dRpakJson.HasMember("Postload") && dRpakJson["Postload"].IsObject() &&
							dRpakJson["Postload"].HasMember(pakName))
						{
							modPak.m_sLoadAfterPak = dRpakJson["Postload"][pakName].GetString();
						}
```
I know it was already like this before, but i think this is better for readability

ASpoonPlaysGames (changes requested):

primedev/mods/modmanager.cpp
EM4Volts: yeah that mustve happened when i was adding a test logger for something. whoopsies!

EM4Volts: should be fixed (if vs didnt betray me)

ASpoonPlaysGames (approved):
Code looks good to me

EM4Volts:
Forgot to mention, fix also adds a warn in console for mods when they dont have the json.
the warn can be helpful to people making mods that wonder why a rpak doesnt load if they dont have the json.

F1F7Y (approved):
Changes look good

GeckoEidechse (approved):
Confirmed working in testing.

Without this PR, removing `rpak.json` from `S2Mods-ResonanceRifle-1.1.1` results in the game crashing on start-up.

With this PR the game no longer crashes on the missing JSON but instead simply fails to load some textures

image


And naturally with this PR an existing valid `rpak.json` gets handled as before.

PR #768, opened 2024-08-18
Fix casing of `KERNEL32.DLL`

ASpoonPlaysGames:
This is a certified windows moment. Case insensitive but also not sometimes I guess


GeckoEidechse (approved):
Launches fine in testing.

PR #769, opened 2024-08-19
Try loading XInput 1.4 before trying to load 9.1.0

barnabwhy:
XInput 9.1.0 has [some limitations to do with `XInputGetCapabilities`](https://learn.microsoft.com/en-us/windows/win32/xinput/xinput-versions) that may make the game behave differently when we use it as a replacement for XInput 1.3
XInput 1.4 doesn't have this issue, which means that if it works correctly we should always prioritise using this version over 9.1.0

XInputGetCapabilities is called by the game in at least two places: `inputsystem.dll + 0xCE4C` & `inputsystem.dll + 0xE2DF` (function pointer set at `inputsystem.dll + 0xCF0B`)

I have checked and 1.4 does have the `/DYNAMICBASE` option enabled so it supports ASLR just like 9.1.0.
Since `XInput1_4.dll` is only included stock on Windows 8 and newer, this PR makes it fall back to `XInput9_1_0.dll` if the former couldn't be loaded, as to maintain the current state of compatibility with Windows 7.

I have personally tested this, and saw no changes in behaviour compared to using XInput 9.1.0, however with this change the game should now handle non-gamepad controllers as it originally was intended to.

**Note: I tested that gamepad input worked fine using an Xbox One S controller. I do not own any non-gamepad controllers so it may be worth investigating what difference is actually made, if any, when something which does not list itself as a gamepad is used in game.**

XInput 1.4 loaded successfully:
image

XInput 1.4 failed to load but 9.1.0 loaded fine:
image

Neither XInput 1.4 or XInput 9.1.0 succeeded in loading:
image-1

### DIFF
--- a/primedev/windows/libsys.cpp
+++ b/primedev/windows/libsys.cpp
@@ -56,15 +56,30 @@ HMODULE WINAPI WLoadLibraryExA(LPCSTR lpLibFileName, HANDLE hFile, DWORD dwFlags
 	// replace xinput dll with one that has ASLR
 	if (lpLibFileName <= lpLibName && !strncmp(lpLibName, XINPUT1_3_DLL, strlen(XINPUT1_3_DLL) + 1))
 	{
-		hModule = o_LoadLibraryExA("XInput9_1_0.dll", hFile, dwFlags);
+		const char* replacementDll = "XInput1_4.dll";
+		hModule = o_LoadLibraryExA(replacementDll, hFile, dwFlags);
 
 		if (!hModule)
 		{
-			MessageBoxA(0, "Could not find XInput9_1_0.dll", "Northstar", MB_ICONERROR);
+			replacementDll = "XInput9_1_0.dll";
+			spdlog::warn("Couldn't load XInput1_4.dll. Will try XInput9_1_0.dll. If on Windows 7 this is expected");
+			hModule = o_LoadLibraryExA(replacementDll, hFile, dwFlags);
+		}
+
+		if (!hModule)
+		{
+			spdlog::error("Couldn't load XInput9_1_0.dll");
+			MessageBoxA(
+				0,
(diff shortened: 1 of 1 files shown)


GeckoEidechse (approved):
I personally find the pattern for selecting the DLL to load a bit weird but it makes sense so ig ¯\_(ツ)_/¯

(Did NOT test, so someone still needs to do that and check that at least 1.4 is getting loaded)

primedev/windows/libsys.cpp
```diff
@@ -56,15 +56,27 @@ HMODULE WINAPI WLoadLibraryExA(LPCSTR lpLibFileName, HANDLE hFile, DWORD dwFlags
 	// replace xinput dll with one that has ASLR
 	if (lpLibFileName <= lpLibName && !strncmp(lpLibName, XINPUT1_3_DLL, strlen(XINPUT1_3_DLL) + 1))
 	{
-		hModule = o_LoadLibraryExA("XInput9_1_0.dll", hFile, dwFlags);
+		const char* replacementDll = "XInput1_4.dll";
```
F1F7Y, line 59: There's no need for `replacementDll`, just pass the string directly to the function.

F1F7Y (changes requested):
Ideally this would be done whereever xinput is being loaded but this also works for now :)

primedev/windows/libsys.cpp
barnabwhy: It's used for the log at the end

F1F7Y: Missed that, would be cool if you renamed it to `pszReplacementDll` <3

barnabwhy: i did it :3

F1F7Y (approved):
Changes from code review have been addressed.

F1F7Y (approved):
Tested on Windows 10, correctly loaded XInput1_4.dll. Didn't notice any issues on mnk just running the gauntlet.

Ship it!

PR #771, opened 2024-08-23
Formatting: Set short inline functions to be on one line

ASpoonPlaysGames:
I'm losing my mind at clang-format making all my short getters and setters into 4 lines for no good reason


GeckoEidechse (approved):
Change to clang-format, no impact on actual logic so no testing needed.

Looked at the format changes in separate branch and they look good. Actual formatting will be applied in separate commit.

PR #772, opened 2024-08-25
project: remove clang-format for good

F1F7Y:
Removes clang-format and replaces it with a code standards document. Doesn't format any files.

First commit (8a4948e69a3f0e9de54a699e14fafa7a9737bde3) is the important change.

Second commit(2f02a6978d7f93267e1c83369ced66e6357777e7) just removes all the clang-format comments.


GeckoEidechse (commented):
My take on this is that we should definitely have the code standard documentation merged ASAP but I disagree with with removing clang-format.

If the formatting is shit (it is), the clang-format config should be adjusted instead. Maybe we could find the settings that are close / the same as valve source engine so that comparing diffs is easier?

ASpoonPlaysGames:
Closing as stale (and so many conflicts)

PR #773, opened 2024-08-25
project: Use pragma pack and remove static_assert spam

F1F7Y:
Only removes the `static_assert` spam, doesn't fix any wonky classes.

### DIFF
--- a/primedev/engine/r2engine.h
+++ b/primedev/engine/r2engine.h
@@ -170,6 +170,7 @@ enum class eSignonState : int
 	CHANGELEVEL = 9, // server is changing level; please wait
 };
 
+#pragma pack(push, 1)
 class CBaseClient
 {
 public:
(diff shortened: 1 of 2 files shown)


GeckoEidechse:
Talking with @RoyalBlue1 in vc about this and we kinda agree that in regards to the `static_assert` while it does hurt readability a bit, it's safer to keep them, especially given the fact that we have a size mismatch as pointed out by your code comment.

That being said we should still do the pragma pack part ofc. Maybe splice from the PR so that we can merge it early? ^^

F1F7Y:
> Talking with @RoyalBlue1 in vc about this and we kinda agree that in regards to the `static_assert` while it does hurt readability a bit, it's safer to keep them, especially given the fact that we have a size mismatch as pointed out by your code comment.

The size mismatch is because the northstar class is wrong  ( see [CPlayer](https://github.com/F1F7Y/NorthstarPrime/blob/primedev/primedev/game/server/player.h) ) . While for basic classes the `static_assert` does hurt readability only a bit for larger classes ( once again see [CPlayer](https://github.com/F1F7Y/NorthstarPrime/blob/primedev/primedev/game/server/player.h) and notice it inherits from other large classes ) it would hurt a LOT. There's also the argument that we have classes without `static_assert` and pragma packing and they haven't broken and unless we change the build system they wont break :).

> That being said we should still do the pragma pack part ofc.

The `static_assert` is fine imo but ig we can yeah

> Maybe splice from the PR so that we can merge it early? ^^

No :P, this is already split as i didnt want it to be too large.

GeckoEidechse:
I think the reasoning @RoyalBlue1 and I had in regards to the `static_assert` was to catch any case in which someone would accidentally swap class members.

I don't really hold too strong of an opinion there other than that I prefer to be overly cautious with checks to assert assumptions early to prevent weird behaviour caused by broken assumptions.

I'll leave the discussion up between the two of you ^^

ASpoonPlaysGames:
I'd argue that they can be removed for cases where we have mapped out all (or the vast majority) of the members. Member offsets accidentally being changed is much more likely to happen when we have huge unk gaps between members

F1F7Y:
Imo we have code review to catch these kinds of errors. If we enforce the `// Offset ( Size )` comments its going to be even easier to catch.

Also anything from datamaps doesnt need offset checks at all as it was given to us by the game (eg CPlayer).

p0358:
I am against removing static_asserts as someone who did a good deal of effort to convince people to add them in the first place xD there's no reason not to have them honestly, and they'll save you in times you might not even expect. Even for fully mapped classes I wouldn't be comfortable removing those, various unexpected things can happen, too many to just cross one's fingers that everything is always gonna be as we expect. Rule of thumb should be, if you need any offset of any member to match up with something in existing game's code, have a static assert for it...

catornot:
closing as stale

PR #774, opened 2024-08-25
Fix manual hooks storing their name badly

ASpoonPlaysGames:
They weren't storing the null terminator since `strlen` doesn't include it. This meant that any `ManualHook`s were logging their name blindly and just reading random memory until they hit a null terminator.

I also moved over to just using `std::string` instead of doing manual allocations etc. since we are doing C++ here we have these tools why not use them.

### Testing:
1. Launch the game without the PR and observe `OnCommandSubmittedHook` doesn't log it's own name properly when enabled (there are also others)
2. Launch the game with this PR and observe that this is now fixed.

### DIFF
--- a/primedev/core/hooks.cpp
+++ b/primedev/core/hooks.cpp
@@ -85,19 +85,9 @@ void __fileAutohook::DispatchForModule(const char* pModuleName)
 			hook->Dispatch();
 }
 
-ManualHook::ManualHook(const char* funcName, LPVOID func) : pHookFunc(func), ppOrigFunc(nullptr)
-{
-	const size_t iFuncNameStrlen = strlen(funcName) + 1;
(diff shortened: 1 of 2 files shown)


ASpoonPlaysGames:
Broken log: (line 924)
[nslog2024-08-25 20-15-11.txt](https://github.com/user-attachments/files/16741903/nslog2024-08-25.20-15-11.txt)

EladNLG (approved):
Looked at everything, changes are obvious, and make sense.

GeckoEidechse (approved):
Cannot reproduce in testing with the CI builds but I trust @ASpoonPlaysGames to have done the testing with the isolated change and so I'm signing it off together with the review by @EladNLG.

RoyalBlue1 (approved):
Changes look good and now its more in line how c++ should be written

PR #775, opened 2024-08-25
Format constructor initializers in a better way

ASpoonPlaysGames:
Old:
image

New:
image


GeckoEidechse (approved):
Formatting related change, no testing needed.

Contributors were poked about the change and overall sentiment was neutral / in favour.

Therefore merging PR.

PR #776, opened 2024-08-25
Fix memory leak and (kinda) cleanup code?

ASpoonPlaysGames:
To try and preserve behaviour I used `do while` which I don't think ended up making the code much more readable... ah well, fixed a memory leak at least.

### Testing:
- Not really certain, but I think a good number of hooks would break if this code was broken, so uhh launch the game and play a match or something I guess

### DIFF
--- a/primedev/Northstar.cmake
+++ b/primedev/Northstar.cmake
@@ -65,6 +65,7 @@ add_library(
     "engine/host.cpp"
     "engine/hoststate.cpp"
     "engine/hoststate.h"
+    "engine/models.cpp"
     "engine/r2engine.cpp"
     "engine/r2engine.h"
     "engine/runframe.cpp"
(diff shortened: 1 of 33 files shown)


F1F7Y:
Considering we want to nuke AUTOHOOK is there any use in potentially breaking it just for it to look slightly nicer?

GeckoEidechse:
Tried testing on Linux but it failed to launch and in the process my Wine prefix got polluted to the point where EA would crash. I'm not sure if this is a Linux skill issue or an effect of this PR.

Either way, if someone could test this on Windows in the meantime to LMK whether it works or not, that'd be great <3

GeckoEidechse:
@RoyalBlue1 said she agrees with @F1F7Y on this.

NachosChipeados (approved):
Works fine for me. Played 2 full matches: one in Northstar and one in Vanilla+ with no crashes. Tested on Windows

Alystrasz:
Merging based on reviews.

PR #777, opened 2024-08-25
Remove uses of Autohook from `audio.cpp`

ASpoonPlaysGames:
Trying to move towards manual hooks over macro hell, one file at a time.

Requires:
- #780

### Code review:

### Testing:
- Load an audio mod and see if it still works I guess?

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -35,5 +35,5 @@ IndentExternBlock: Indent
 PointerAlignment: Left
 SortIncludes: false
 NamespaceIndentation: All
-PackConstructorInitializers: NextLine
-BreakConstructorInitializersBeforeComma: true
+PackConstructorInitializers: Never
+BreakConstructorInitializers: BeforeComma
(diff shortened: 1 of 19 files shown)


GeckoEidechse:
Part of #778

F1F7Y (approved):
Changes look ok, however this isnt going to work in testing as `mileswin64.dll` is imported by `client.dll` and we only run callbacks for `LoadLibrary*` calls. ( See this change in primedev: https://github.com/F1F7Y/NorthstarPrime/commit/368f66043c6b908dc93ec29f3e856beaaabf6df5#diff-54ccaa9ecaf9f49dd1e7d51bfe382cb5025349aee6cd4a3003282314a99d19ad )

F1F7Y:
(Approved because this'll fail when testing, meh)

GeckoEidechse (changes requested):
As @F1F7Y predicted, this breaks in testing.

Tested with https://thunderstore.io/c/northstar/p/FUZE/Replace_Background_Music_for_The_Lobby_with_Helldivers2_OST/

Without the PR the lobby music was overridden, with the PR it no longer was.

ASpoonPlaysGames:
Now depends on #780

GeckoEidechse (approved):
Confirmed working testing. My death sound audio override mod still worked.

F1F7Y (approved):
No changes to code, fix is already on main. LGTM

PR #782, opened 2024-08-26
Remove uses of Autohook from `clientauthhooks.cpp`

ASpoonPlaysGames:
### Code review:

### Testing:
- Make sure that authing with Atlas (`AuthenticateOriginWithMasterServer`) still works and all that

### DIFF
--- a/primedev/client/clientauthhooks.cpp
+++ b/primedev/client/clientauthhooks.cpp
@@ -12,16 +12,14 @@ const int NOT_DECIDED_TO_SEND_TOKEN = 0;
 const int AGREED_TO_SEND_TOKEN = 1;
 const int DISAGREED_TO_SEND_TOKEN = 2;
 
-// clang-format off
-AUTOHOOK(AuthWithStryder, engine.dll + 0x1843A0,
-void, __fastcall, (void* a1))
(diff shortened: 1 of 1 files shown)


F1F7Y (approved):
Changes look okay :3

primedev/client/clientauthhooks.cpp
```diff
@@ -3,25 +3,21 @@
...
-// clang-format on
+static void (*__fastcall o_pAuthWithStryder)(void* a1) = nullptr;
+static void __fastcall h_AuthWithStryder(void* a1)
 {
 	// don't attempt to do Atlas auth if we are in vanilla compatibility mode
 	// this prevents users from joining untrustworthy servers (unless they use a concommand or something)
 	if (g_pVanillaCompatibility->GetVanillaCompatibility())
 	{
-		AuthWithStryder(a1);
+		o_pAuthWithStryder(a1);
```
GeckoEidechse, line 20: variable rename in the same PR, aaaaah

(it's ok, I'll just ignore it to not stall it :P)

GeckoEidechse (approved):
Authed with master server just fine in testing and joined a random server without issues.

PR #783, opened 2024-08-26
Remove uses of Autohook from `clientruihooks.cpp`

ASpoonPlaysGames:
### Code review:

### Testing:
- Test that `rui_drawEnable` still functions properly (`0` stops drawing all RUI, `1` draw RUI as per usual)


F1F7Y (approved):
Changes look okay :3

GeckoEidechse (approved):
`rui_drawEnable` still works as expected in testing

image

PR #780, opened 2024-08-26
Call module callbacks for nested imported modules (i.e `mileswin64.dll`)

ASpoonPlaysGames:
Mostly ported from 368f66043c6b908dc93ec29f3e856beaaabf6df5 with some modifications to the recursion

PR #784, opened 2024-08-26
Remove uses of Autohook from `clientvideooverrides.cpp`

ASpoonPlaysGames:
Depends on #780 

### Code review:

### Testing:
- Make sure that `.bik` files can still be loaded from mod directories (custom main menus)


F1F7Y (approved):
Changes look good :>

GeckoEidechse (approved):
Confirmed working in testing. My custom main menu video still played :D

image

PR #786, opened 2024-08-27
Remove uses of Autohook from `debugoverlay.cpp`

ASpoonPlaysGames:
### Code review:

### Testing:
- Make sure that `enable_debug_overlays` still works


GeckoEidechse (approved):
Confirmed working in testing

image

Issue #778, opened 2024-08-25
Move away from AUTOHOOK and then do hook queueing for performance


GeckoEidechse:
From @ASpoonPlaysGames on Northstar Discord via https://discord.com/channels/920776187884732556/950322078945538058/1277403606617686147

> The Plan™️ is that once things are moved away from AUTOHOOK I want to do the hook queueing stuff as well for performance

GeckoEidechse:
Current PRs working towards this:

- #796
- #795
- #793
- #792
- #789
- #787
- #786
- #784
- #783
- #782
- #777

z9ctn4fn:
seems good to close unless there's some other files that need autohook removing in which case do that first

PR #787, opened 2024-08-27
Remove uses of Autohook from `languagehooks.cpp`

ASpoonPlaysGames:
### Code review:

### Testing:
- Make sure that the logging in the language hooks still triggers and behaves normally.

### DIFF
--- a/primedev/client/languagehooks.cpp
+++ b/primedev/client/languagehooks.cpp
@@ -48,10 +48,8 @@ std::string GetAnyInstalledAudioLanguage()
 	return "NO LANGUAGE DETECTED";
 }
 
-// clang-format off
-AUTOHOOK(GetGameLanguage, tier0.dll + 0xF560,
-char*, __fastcall, ())
-// clang-format on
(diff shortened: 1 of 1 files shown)


GeckoEidechse:
Not sure how to test this cause I cannot even get current main to switch the language when setting the game to German in Steam and then using either `-lanuage=english` or `-lanuage english`...

What do you think would be the best way to proceed?

F1F7Y (approved):
Changes to code look good :rabbit:

F1F7Y:
> Not sure how to test this cause I cannot even get current main to switch the language when setting the game to German in Steam and then using either `-lanuage=english` or `-lanuage english`...
> 
> What do you think would be the best way to proceed?

Merge it and make a tracking issue or bother someone who can step through the code :)

GeckoEidechse:
Made issue via #778 

I guess we just YOLO this one otherwise. Like ideally we'd fix the hook first and then replace the macro as it _should_ be easier to bisect the change that broke it in this case. That being said, I also don't wanna stall the process of removing the autohook macro so I'm merging despite not having a way to verify in testing that the change is correct.

PR #788, opened 2024-08-27
vscript: Move squirrel types to their respective files

F1F7Y:
Mostly taken from primedev, just moves some structs around.

Things to test:
- Does the video game run
- Do script commands work (ei `script_client`, `script`, ...)
- Idk


ASpoonPlaysGames (approved):
Looks good at a glance, comes sorta free with some better struct/class definitions as well.

Disclaimer: I did not go and verify that the memory layout of all of the structs matches up with the previous ones. That would take too long and gets verified anyway with testing later for the most part.

primedev/squirrel/squirreldatatypes.h
```diff
@@ -1,501 +0,0 @@
...
-	unsigned char gap_43F1[23];
-};
-
-/* 165 */
-struct tableNode
-{
-	SQObject val;
-	SQObject key;
-	tableNode* next;
-};
```
GeckoEidechse, line 410: Where does this one go to?

GeckoEidechse (commented):
Reproduced the change locally to better understand it. Got it to almost match. Noticed that there a few more changes than just moving code around like adjusting existing structs with new fields that have been reversed since.
Would have been nice to have that as a separate change but oh well.

Otherwise I didn't notice any obvious issues but I also didn't go super in-depth.

primedev/squirrel/squirreldatatypes.h
F1F7Y: Renamed to `_HashNode` in `SQTable`

GeckoEidechse (approved):
Launched private match with this PR and ran

```
sv_cheats 1
script AddPlayerScore(GetPlayerArray()[0], "KillPilot")
```

and everything still worked as expected.

GeckoEidechse:
This one is gonna require a lot more extensive testing to actually verify correctness. That's not gonna happen. So the best we can do is just test in prod and hope that any issues get reported back to us (no remote logging to do that automatically ;-;)

PR #789, opened 2024-08-27
Remove uses of Autohook from `latencyflex.cpp`

ASpoonPlaysGames:
### Code review:

### Testing:
- Make sure that the convar `r_latencyflex` works as usual

### DIFF
--- a/primedev/client/latencyflex.cpp
+++ b/primedev/client/latencyflex.cpp
@@ -6,15 +6,13 @@ ConVar* Cvar_r_latencyflex;
 
 void (*m_winelfx_WaitAndBeginFrame)();
 
-// clang-format off
-AUTOHOOK(OnRenderStart, client.dll + 0x1952C0, 
-void, __fastcall, ())
-// clang-format on
+void(__fastcall* o_pOnRenderStart)() = nullptr;
+void __fastcall h_OnRenderStart()
 {
 	if (Cvar_r_latencyflex->GetBool() && m_winelfx_WaitAndBeginFrame)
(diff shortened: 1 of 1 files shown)


GeckoEidechse:
`r_latencyflex` doesn't show up for me on `main`. I assume it only does when actually running LatencyFlex? Honestly this is gonna be a pain to test so if @F1F7Y says the code is fine, I'd propose to merge and immediately release afterwards to test in prod instead.

F1F7Y:
`r_latencyflex` should always show up on client. I'll look into it when I get home ( in ~4h )

F1F7Y:
I forgor 💀

F1F7Y (approved):
Changes look good :fish:

F1F7Y:
not seeing `r_latencyflex` means the dll load callback just isnt getting called or its returning here https://github.com/R2Northstar/NorthstarLauncher/blob/c8b7bb0f41f04119492a46c41d55e27d7eeb6f5a/primedev/client/latencyflex.cpp#L31-L32 which also logs so would be cool if you could check logs @GeckoEidechse if you have time.

Reading the check the convar only exists when either `latencyflex_layer.dll` or `latencyflex_wine.dll` exist.

GeckoEidechse (approved):
The line @F1F7Y mentioned still shows up using this PR so I guess it works. Unfortunately I don't have a LatencyFlex setup to test this further so this must do ig.

```
[2024-08-31] [11:32:02] [NORTHSTAR] [info] Unable to load LatencyFleX library, LatencyFleX disabled.
```

PR #793, opened 2024-08-29
Remove uses of Autohook from `rejectconnectionfixes.cpp`

ASpoonPlaysGames:
### Code review:

### Testing:
- Honestly, this is a difficult one to test functionality for... I guess the hook being created will have to do?


xFrannM (commented):
PR tested and working fine.

F1F7Y (approved):
Changes look good 🐱

PR #795, opened 2024-08-30
Remove uses of Autohook from `filesystem.cpp`

ASpoonPlaysGames:
### Code review:

### Testing:
- Make sure that the loading mod files (scripts etc.) works as normal


F1F7Y (approved):
Code👍

GeckoEidechse (commented):
Confirmed working in testing.

As can be seen from the screenshot mod file (bik video in this case) was still loaded
image

GeckoEidechse (approved):
(forgot to press approve in the review above)

PR #798, opened 2024-08-31
Remove uses of Autohook from `sourceinterface.cpp`

ASpoonPlaysGames:
### Code review:

### Testing:
- Make sure that `CreateInterface <SERVER/CLIENT/ENGINE>` still get logged
- Make sure that the in-game console still exists (since `InitialiseConsoleOnInterfaceCreation` gets called from `ClientCreateInterface`)

### DIFF
--- a/primedev/core/sourceinterface.cpp
+++ b/primedev/core/sourceinterface.cpp
@@ -1,8 +1,6 @@
 #include "sourceinterface.h"
 #include "logging/sourceconsole.h"
 
-AUTOHOOK_INIT()
-
 // really wanted to do a modular callback system here but honestly couldn't be bothered so hardcoding stuff for now: todo later
 
 static void*(__fastcall* o_pClientCreateInterface)(const char* pName, const int* pReturnCode) = nullptr;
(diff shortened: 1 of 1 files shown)


GeckoEidechse (approved):
Confirmed working in testing:

```
[2024-08-31] [15:24:55] [NORTHSTAR] [info] CreateInterface ENGINE VCvarQuery001
[2024-08-31] [15:24:55] [NORTHSTAR] [info] CreateInterface ENGINE VAvi001
[2024-08-31] [15:24:55] [NORTHSTAR] [info] CreateInterface ENGINE VBik001
[2024-08-31] [15:24:55] [NORTHSTAR] [info] CreateInterface ENGINE VENGINE_LAUNCHER_API_VERSION004
[2024-08-31] [15:24:55] [NORTHSTAR] [info] CreateInterface ENGINE VDataCache003
[2024-08-31] [15:24:55] [NORTHSTAR] [info] CreateInterface ENGINE VPrecacheSystem001
```
and
image

F1F7Y:
Gotta say this file has no right to still be in this code base. Same with the `SourceInterface` class (copied from icepick btw), might PR later :)

F1F7Y (approved):
Changes look good

PR #799, opened 2024-08-31
Remove uses of Autohook from `dedicated.cpp`

ASpoonPlaysGames:
### Code review:

### Testing:
- Make sure that dedicated servers still gracefully close on script error instead of hard crashing.

### DIFF
--- a/primedev/dedicated/dedicated.cpp
+++ b/primedev/dedicated/dedicated.cpp
@@ -8,8 +8,6 @@
 #include "masterserver/masterserver.h"
 #include "util/printcommands.h"
 
-AUTOHOOK_INIT()
-
 bool IsDedicatedServer()
 {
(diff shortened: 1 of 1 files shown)


F1F7Y (approved):
Changes look good

GeckoEidechse:
> Make sure that dedicated servers still gracefully close on script error instead of hard crashing.

Do we have an easy way to script crash via some script function? Or do we need to write like a broken mod to test this >.<

ASpoonPlaysGames:
> Do we have an easy way to script crash via some script function? Or do we need to write like a broken mod to test this >.<

The latter I think. Iirc any error induced by a simple `script` command is non-fatal, so youd have to introduce a logic (not compile) error into the scripts and trigger it

F1F7Y (approved):
Works in testing

PR #800, opened 2024-08-31
Remove uses of Autohook from `dedicatedmaterialsystem.cpp`

ASpoonPlaysGames:
### Code review:

### Testing:
- Make sure that the `-softwared3d11` launch arg still works as before? (tbh we could remove this now? Don't think I've seen anyone use it in the last 2 years)

### DIFF
--- a/primedev/dedicated/dedicatedmaterialsystem.cpp
+++ b/primedev/dedicated/dedicatedmaterialsystem.cpp
@@ -3,9 +3,7 @@
 
 AUTOHOOK_INIT()
 
-// clang-format off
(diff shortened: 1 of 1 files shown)


F1F7Y (approved):
Code looks a-okay

GeckoEidechse:
> Make sure that the `-softwared3d11` launch arg still works as before? (tbh we could remove this now? Don't think I've seen anyone use it in the last 2 years)

Docker image doesn't use it, right?

F1F7Y (approved):
Works in testing

PR #802, opened 2024-08-31
Remove uses of Autohook from `host.cpp`

ASpoonPlaysGames:
### Code review:

### Testing:
- Check logs to make sure that all of the changed hooks are being created properly (this one logs `Host_Init()` in the console when it runs so check for that when going into the lobby)
- Make sure that `autoexec_ns_server` and `autoexec_ns_client` are being executed still etc.

### DIFF
--- a/primedev/engine/host.cpp
+++ b/primedev/engine/host.cpp
@@ -6,8 +6,6 @@
 #include "r2engine.h"
 #include "core/tier0.h"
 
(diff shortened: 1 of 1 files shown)


F1F7Y (approved):
Changes look good :giraffe:

F1F7Y (approved):
Works in testing.

PR #803, opened 2024-08-31
logging: Remove use of `SourceInterface` for `CGameConsole`

F1F7Y:
`SourceInteface` class goes back to icepick and is not good. We have a replacement, let's use it.

When testing make sure the console can stil be opened and is being logged to.


GeckoEidechse (approved):
Confirmed working in testing as per instructions

image

ASpoonPlaysGames (approved):
Code looks good, nothing uses the `extern` in the header anyway so that's fine to remove

RoyalBlue1 (approved):
Makes code look way cleaner and mimics what original source does more closely

PR #804, opened 2024-08-31
core: Remove use of `SourceInterface` for `CCvar`

F1F7Y:
### Code review:

Look at code

### Testing:

Run commands like `help` or `find` that use `g_pCVar`.


GeckoEidechse (approved):
Confirmed working in testing as per instructions

ASpoonPlaysGames (approved):
Code looks good, `g_pCVarInterface` was basically never used anyway since we just used `g_pCVar` instead.

PR #806, opened 2024-09-02
Remove uses of Autohook from `hoststate.cpp`

ASpoonPlaysGames:
### Code review:

### Testing:
- Check logs to make sure that all of the changed hooks are being created properly (these ones for the most part log something when they are triggered, so check for that)
- Make sure that server presence is tracking the correct port (honestly easiest to just stick some breakpoints in the code and make sure that it is reading the ConVar properly


F1F7Y (approved):
Works, changes look good

PR #807, opened 2024-09-02
game: Add more ways to crash to the `crash_test` concommand

F1F7Y:
### Code review:

Look at code.

### Testing:

Test by running `crash_test` with and without an argument.
`crash_test` is the same as `crash_test 0`

All (0 - 4) options should crash, not all triger the crash handler!


.gitignore
```diff
@@ -18,7 +18,7 @@ mono_crash.*
 
 # CMake output
 out/
-game/
```
ASpoonPlaysGames, line 21: Unrelated change?

F1F7Y: related change. `clientmode_shared.cop` is in `primedev/game/client/` which was getting filtered by this. The `game/` in the gitignore was always meant for the build folder.

GeckoEidechse: We should probably update our build instructions then cause following current build instructions in `BUILD.md` results in the build output being put into `/game` as opposed to `/build/game` [^1] \^\^

[^1]: `/` is repo root in this case, not filesystem root

GeckoEidechse (approved):
Successfully crashed Northstar with `crash_test` with and without arguments (`1`-`5`)

Not all of them created a `.dmp`  or wrote the stacktrace to logs which I suppose is the reason we adding more cases so that we can properly handle it \^\^

.gitignore
F1F7Y: I designed current cmake to be used with visual studio (which should put it in `/build/game/` still). When manually running cmake you can put it wherever you want by design. I plan to redo cmake in the near future anyways

GeckoEidechse: Yeah so the building with VS as you outlined should put it in `/build/game/` but the [compiling from command line section in our BUILD.md](https://github.com/R2Northstar/NorthstarLauncher/blob/261fcdaec4467dd4fdde4907cf7863f308ec344d/BUILD.md#vs-build-tools) in the way they are written right now, will result in the files being put in `/game/` instead. Same for CI. I'll try to update docs and CI accordingly o7

F1F7Y: cmake is designed to let you specify where you want to build to. This by design means we can't guard your build directory in `.gitignore` as we don't know where you want to build. The most common approach to building is to build into a `build` directory.

GeckoEidechse: Yes, hence we should have our compilation instructions so that when following them, output will be in `/build/game/`. Anyone with little to no experience in C++ development with CMake will just blindly follow the given commands in the build instructions and right now they are set up so that the resulting binaries will be in `/game/`

Hence we should update our instructions so that when blindly following them, the files will be in `/build/game`.
Whether you than actually follow those instructions or want your output somewhere else is then up to the actual developer ^^

ASpoonPlaysGames (approved):
Code looks good to me, haven't tested to make sure that they actually crash (but they probably do)

PR #809, opened 2024-09-03
Ensure whitespace indentation is consistent in Dockerfile

GeckoEidechse:
Just ensures that indentation is consistent and not a mix of tabs and spaces. Styling change, does not affect functionality or anything.


Alystrasz (approved):
Looks good to me (no more tabs).

PR #810, opened 2024-09-04
engine: Remove uses of Autohook from `runframe.cpp`

ASpoonPlaysGames:
### Code review:

### Testing:
- ...that's all, honestly not sure why we hook this function anymore, probably a legacy hook or something at this point.


F1F7Y (approved):
Changes look good

primedev/engine/runframe.cpp
```diff
@@ -3,17 +3,14 @@
...
-
-// clang-format off
-AUTOHOOK(CEngine__Frame, engine.dll + 0x1C8650,
-void, __fastcall, (CEngine* self))
-// clang-format on
+static void(__fastcall* o_pCEngine__Frame)(CEngine* self) = nullptr;
+static void __fastcall h_CEngine__Frame(CEngine* self)
 {
-	CEngine__Frame(self);
+	o_pCEngine__Frame(self);
```
Alystrasz, line 9: Does this appear in the logs?
The old hook did (`Enabling hook CEngine__Frame`), but I can't find any trace of the new one.

GeckoEidechse: @ASpoonPlaysGames

ASpoonPlaysGames: non-auto hooks dont log when they are enabled iirc

GeckoEidechse:
Testing in prod.

PR #814, opened 2024-09-04
Add dependabot config to check for GitHub Actions updates

GeckoEidechse:
Adds a dependabot config to check for updates to the GitHub Actions we use and makes a PR with the version bump if a new version is found.

### Code review:

_None_

### Testing:

_None_

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,9 @@
+# Configures dependabot
+
+version: 2
+updates:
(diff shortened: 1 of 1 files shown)


GeckoEidechse:
Not much to say here. Just gonna merge cause all it's gonna do is notify us when there are updates to the actions we use.

In the future we could probably extend it to also check for updates on the dependencies we submodule ^^

PR #815, opened 2024-09-04
feat: Make MAD process cancellable

Alystrasz:
A long time ago, during the initial development of the feature, somebody (Jan maybe?) suggested making the mod install process cancellable, in case it takes too long (slow Internet speed, for instance).

This PR does just that, implementing the `ModDownloader::CancelDownload` function (already present in the header btw), that can stop MAD process at two moments:
* during archive downloading, using [`CURLE_ABORTED_BY_CALLBACK`](https://curl.se/libcurl/c/libcurl-errors.html#CURLEABORTEDBYCALLBACK) (returned from download callback function, see https://curl.se/libcurl/c/CURLOPT_XFERINFOFUNCTION.html);
* during archive extraction.

**Must be merged with mods counterpart (https://github.com/R2Northstar/NorthstarMods/pull/865)!**

### Media

[mod_download_cancelling.webm](https://github.com/user-attachments/assets/70dc0c28-511c-470f-9b4c-a51358bde6a8)

*Here, mod install process is interrupted twice, by successively stopping archive downloading and archive extraction; note that this does not prevent mod install from completing successfully when ran a third time.*

### Testing

1. Install both launcher and mods PRs;
2. Launch game and go to server browser;
3. Enable MAD (`allow_mod_auto_download 1`);
4. Try to join `Space battle` server (this server requires a mod that's big enough to leave you enough time to click the new "cancel" button);
5. Enjoy the new "cancel" button!


ASpoonPlaysGames (approved):
Code looks good

What happens if a server requires two mods, and the first one completes but the second is aborted? Do we clean up the already-downloaded mod properly?

Alystrasz:
> What happens if a server requires two mods, and the first one completes but the second is aborted? Do we clean up the already-downloaded mod properly?

The download method is triggered once per mod: this means first mod would be installed, and second mod would be cleaned up.

ASpoonPlaysGames:
> The download method is triggered once per mod: this means first mod would be installed, and second mod would be cleaned up.

Do we enable the mods as soon as they are downloaded? Could this potentially cause issues?

Alystrasz:
> Do we enable the mods as soon as they are downloaded? Could this potentially cause issues?

Nope, mods enabling happens after *all* mods have been downloaded, and only has an effect on `RequiredOnClient` mods.

NachosChipeados:
Works fine 👍 
https://github.com/user-attachments/assets/2d6c7cc3-aa24-42fd-b6a5-2a3b373c3684

NachosChipeados (approved):
Works fine 👍 

Cancelling while the mod is downloading:

https://github.com/user-attachments/assets/adba41df-9e71-48a0-bfd6-605727422fd1

Cancelling while the mod is extracting:

https://github.com/user-attachments/assets/72c46c02-5c5a-414e-879d-28e897fce3f1

GeckoEidechse:
Thanks for testing. I'll push out the current release-candidate as a full patch release, give it a day or two to stabilise and then merge this and prepare the next minor release ^^

wolf109909:
Tested on NorthstarCN but it works fine!

PR #816, opened 2024-09-06
core: Remove unused SourceInterface class

F1F7Y:
### Code review:

Removes unused `SourceInterface` class, moves `InterfaceStatus` enum to `tier1.h`.

### Testing:

If it compiles it works as this was unused. In terms of `InterfaceStatus` iirc source doesnt use the second argument of `CreateInterface` and you should still check for nullptr so even if i did break it it shouldnt affect release.


ASpoonPlaysGames (approved):
Code looks good, not much to say

Alystrasz (approved):
Code looks good, played an entire game with no issues.
(once again, the simple fact that CI build finishes successfully validates this PR)

PR #817, opened 2024-09-07
Rework MAD `cleanup` handles

### LINKED ISSUE #756: Issue on mod download failure
*(Initially posted by @wolf109909 in Discord)*

I think I've found some issues that might occur when mad failed to finish any of the install states in current main branches. Heres the walkthrough:
I've done a few temporary fixes to these problems while debugging myself, here is the first part:

1. user double click on the server with a downloadable mod
2. mod auto download kicks in, but failed when downloading due to poor internet
3. user did `1` again, instead of retrying the download, user got sent into the server and disconnected with out of sync with server.

##### Debugging
In `moddownloader.cpp` at `void ModDownloader::DownloadMod(std::string modName, std::string modVersion)`, the `ScopeGuard cleanup` will get called even when you early return in any of the failing conditions which sets the `modState.state`. this will cause modState.state to always be set to `DONE`, which makes  the UI think the mod is downloaded (detailed info on the second part). I just removed the scope guard and moved to the old code reusing style for temoporary fix.

After that, I tried again. I get these instead:

1. user double click on the server with a downloadable mod
2. mod auto download kicks in, but failed when downloading due to poor internet
3. user gets a download failed warning, and got sent back to the server browser.
4. user did `1` again, but this time got to `3` immediately.
5. user retried again, this time he got to `2` but the download progress bar glitches.
6. things broke

From my debugging, I found `3` is caused by line 39 and line 41 in `menu_ns_moddownload.nut` where the `NSGetModInstallState()` call is right next to `NSDownloadMod( mod.name, mod.version )`, when in CPP the `void ModDownloader::DownloadMod(std::string modName, std::string modVersion)` creates a `requestThread` and making the `NSDownloadMod` in squirrel operation not blocking, giving `NSGetModInstallState()` with the old `g_pModDownloader->modState.state`. Immediately the UI will freak out and send user a `Download Failed` warning BUT in the mean time `ModDownloader::DownloadMod` is still running and didn't get canceled.  Once the user triggers `1` again, the `NSGetModInstallState()` will give the correct state but its set from the previous `NSDownloadMod` call. there we got some kind of race condition between two downloader threads that does filesystem opertaions and computer not very happy(its not that bad tbh just glitchy).

##### Another tempfix

I added simple thread block like the ones used in masterserver.cpp to limit only one  concurrent threads at the same time. then I added a new state on top of `modState.state`, naming it `IDLE`. I set the state to `IDLE` before spawning `requestThread`, making the script able to wait for the first `modState.state` and not just returning. At this stage, I think the whole process is more stable and user-friendly.
ps: I do this in discord instead of a github issue cause I'm not entirely sure if I'm doing this correctly. there are environmental differences between my testing branch and main branch. I will push the fixed code to my branch, but they are not very clean and are just temp fixes.

Alystrasz:
### Changes

* Mod is not marked as successfully installed when the process failed;
* When mod downloading/extraction fails, mod directory is removed.

### Testing

1. Install this PR;
2. Enable MAD (`allow_mod_auto_download 1`);
3. Try to join Space battle server and cancel mod downloading/extraction;
4. Check your local profile, the remote mod should not appear in the `runtime/remote/mods` directory;
5. Check your temporary directory (for me on Windows, it's `C:\Users\Remy\AppData\Local\Temp`), there should be no mod archive.

---

Closes #756.

---

##### TODOs

- [x] ZIP archive is not deleted when download is cancelled (but it is when cancelling extraction)
- [x] Mod is not removed when cancelling archive extraction


Alystrasz:
Waiting for #815 to be merged so we have a simple way to cancel MAD process.

wolf109909:
Tested on NorthstarCN v1.18 branch and it works fine.
Scenario 1: Unplugging ethernet while mod download is happening: Prompted mod verify or download error.
Scenario 2: Trying to join the server without ethernet connection: Immediately prompted mod verify or download error.
Scenario 3: Joining the server with Ethernet reenabled: Verify and download begins normally, with the download state being reset after the first download state update. 
Overall pretty robust, but my internet restricts me from doing more stressful tests due to unstable connection speed to thunderstore and github without a proxy. (literally failed to download once even without touching my network, and the error was handled just fine ! )

NachosChipeados (commented):
Mod directory doesn't seem to get removed when cancelling extract
https://github.com/user-attachments/assets/1932a97b-5df1-4fef-90fa-d0ab6940ae01

NachosChipeados (approved):
Works fine, both the archive and the mod folder get deleted during the download/extract processes

https://github.com/user-attachments/assets/43d80012-d6b3-49e7-ab22-d2efc30e5042

NachosChipeados (approved):
I think the code looks good

GeckoEidechse:
I don't have the opportunity rn to give this is a proper review. I trust @Alystrasz with the suggested change here, especially as @NachosChipeados confirmed it working in testing.

As such merging rn without further reviews.

PR #820, opened 2024-09-08
Remove uses of Autohook from `serverauthentication.cpp`

ASpoonPlaysGames:
### Code review:
- Something to note is that we get a pointer to `CBaseClient__Disconnect` in another file as well, hence the weird name on the original hook. Potentially any calls we were making to `CBaseClient__Disconnect` instead of `_CBaseClient__Disconnect` are bypassing the hook? Something worth investigating at some point, shouldn't affect this PR though.

### Testing:
- Make sure that players can still auth, join and leave a server with their persistence being written etc.


F1F7Y (approved):
Changes look good

GeckoEidechse (approved):
Works in quick testing. Just hosted a server with this PR and joined it, played a bit, and left again.

PR #822, opened 2024-09-12
remove useless log

uniboi:
It's logged in server.dll somewhere already I believe and doesn't make sense to repeat this for every loaded plugin


F1F7Y (approved):
Already logged in `DestroyVMHook`

PR #823, opened 2024-09-12
plugins: Allow plugins to customize their log colors

uniboi:
If the color flag is 0 (expected but undocumented default behaviour of plugins for GetField) NS::Colors::PLUGIN is used as a default.

The Color is packed into a 64 bit integer like this
```cpp
const Color = packed struct(u64) {
    red: u8,
    green: u8,
    blue: u8,
    _: u40 = 0,
};
```
The colors only occupy the **lower 3 bytes** of the integer. The remaining bits are ignored. When determining to use the default color or the provided log color, the lower 12 bits are checked if any is not 0.


catornot:
Were the plugin field and plugin string enums supposed to be non exhaustive?

uniboi:
yes to allow adding more fields in the future they're supposed to be non exhaustive

catornot:
Alright a few plugins crashing is fine for now ig
Will review this pr later today probably

primedev/plugins/plugins.h
```diff
@@ -20,6 +20,7 @@ class Plugin
 	std::string m_location; // path of the dll
 	bool m_runOnServer;
 	bool m_runOnClient;
+	Color m_logColor = NS::Colors::PLUGIN;
```
F1F7Y, line 23: initialize in the constructor, not here thanks <3

primedev/plugins/interfaces/IPluginId.h
```diff
@@ -26,6 +27,10 @@ class IPluginId
 public:
 	virtual const char* GetString(PluginString prop) = 0;
 	virtual int64_t GetField(PluginField prop) = 0;
+
+	static const int64_t COLOR_R_MASK = 0xFF;
```
F1F7Y, line 31: Any reason for this to be in the header when its only needed in the constructor which is in a source file?

F1F7Y (approved):

F1F7Y (changes requested):
The above review wasnt supposed to be an approval smh

catornot (approved):
Looks good me thinks

also pretty colors for plugins :)
image

GeckoEidechse:
Yeah so for some reason it renders as (almost?) black when the colour value is not set which looking at the code shouldn't be the case as it should just be default colour if not set

image

uniboi:
Current rrplug builds return uninitialized values for colors. If a plugin returns 0 as color, a default will be used

primedev/plugins/plugins.cpp
```diff
@@ -69,6 +70,12 @@ Plugin::Plugin(std::string path)
 	m_runOnServer = context & PluginContext::DEDICATED;
 	m_runOnClient = context & PluginContext::CLIENT;
 
+	int64_t logColor = m_pluginId->GetField(PluginField::COLOR);
+	if ((logColor & 0xFFFFFF) != 0)
```
GeckoEidechse, line 74:
```suggestion
	// Apply custom colour if plugin has specified one
	if ((logColor & 0xFFFFFF) != 0)
```

GeckoEidechse:
cc @F1F7Y whether you wanna re-review, given that you requested changes. Otherwise if there's no response I'd merge within a day or two to not stall the PR o7

F1F7Y (approved):
Code looks acceptable

PR #824, opened 2024-09-28
server: Remove uses of Autohook from `buildainfile.cpp`

ASpoonPlaysGames:
### Code review:

### Testing:
- Make sure that `ns_ai_dumpAINfileFromLoad` still works as expected (dumps AIN files on load)


Alystrasz (commented):
Same remark than https://github.com/R2Northstar/NorthstarLauncher/pull/810#discussion_r1831748616, what should I look for in logs when testing this?
The only thing noticeable for now is the old hooks are no longer logged:
```text
[2024-11-06] [23:07:11] [NORTHSTAR] [info] Enabling hook CAI_NetworkBuilder__Build
[2024-11-06] [23:07:11] [NORTHSTAR] [info] Enabling hook LoadAINFile
```

RoyalBlue1 (approved):
Code looks good and ain file writing still workes as intended when tested

PR #825, opened 2024-10-05
Fix rpak aliasing not working when trying to alias towards a vanilla rpak

ASpoonPlaysGames:
Fixing stuff described by @Alystrasz on discord. I don't really like this behaviour as it behaves a bit differently for vanilla rpak loads and modded rpak loads, but aliasing is hacky already so meh.

### DIFF
--- a/primedev/core/filesystem/rpakfilesystem.cpp
+++ b/primedev/core/filesystem/rpakfilesystem.cpp
@@ -342,7 +342,7 @@ void PakLoadManager::UnloadDependentPaks(PakHandle handle)
 static void HandlePakAliases(std::string& originalPath)
 {
 	// convert the pak being loaded to its aliased one, e.g. aliasing mp_hub_timeshift => sp_hub_timeshift
-	for (int64_t i = g_pModManager->m_LoadedMods.size() - 1; i > PakHandle::INVALID; i--)
+	for (int64_t i = g_pModManager->m_LoadedMods.size() - 1; i > -1; i--)
 	{
 		Mod* mod = &g_pModManager->m_LoadedMods[i];
 		if (!mod->m_bEnabled)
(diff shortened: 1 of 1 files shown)


GeckoEidechse:
cc @Alystrasz for testing

primedev/core/filesystem/rpakfilesystem.cpp
```diff
@@ -342,15 +342,15 @@ void PakLoadManager::UnloadDependentPaks(PakHandle handle)
 static void HandlePakAliases(std::string& originalPath)
 {
 	// convert the pak being loaded to its aliased one, e.g. aliasing mp_hub_timeshift => sp_hub_timeshift
-	for (int64_t i = g_pModManager->m_LoadedMods.size() - 1; i > PakHandle::INVALID; i--)
+	for (int64_t i = g_pModManager->m_LoadedMods.size() - 1; i > -1; i--)
```
Alystrasz, line 345: Why removing `PakHandle::INVALID` while it already equals `-1`?

Alystrasz:
Explanation from Spoon on Discord:

> it only worked before because of two bugs in the old system
>
> which meant that when you aliased an rpak, the game would look in the vanilla folder for it first instead of the modded one, but then when it tried to load the vanilla one, it loaded the modded one as well
>
> new system only looks in the modded folder for the alias, so for s2s where it is trying to alias a non-existent rpak to a vanilla rpak that behaviour is broken

Alystrasz (approved):
To test this change, I installed [s2space](https://thunderstore.io/c/northstar/p/odds/s2space/), a mod aliasing vanilla rpaks; then launched a game in private match using `map mp_s2s`.

##### Without PR

image

##### With PR

image

primedev/core/filesystem/rpakfilesystem.cpp
ASpoonPlaysGames: We arent iterating over Pak handles here, so it was a mistake to use the named constant in the first place

uniboi (approved):
Code looks good. Maybe a comment that mentions that vanilla aliases behave a bit different?

GeckoEidechse:
cc @ASpoonPlaysGames to decide whether to merge with or without comment

ASpoonPlaysGames:
Dont think it's necessary, it acts fairly intuitively

GeckoEidechse:
ah fuck, forgot to prefix commit with `core:`. Sorry @F1F7Y :c

PR #758, opened 2024-07-28
fix: Multiple mod versions

Alystrasz:
**Associated mods PR:** https://github.com/R2Northstar/NorthstarMods/pull/824

# Problem

Whether through manual installation or through mod auto-downloading, several versions of the same mod can be installed in Northstar, which currently does not handle this:

[main_status.webm](https://github.com/user-attachments/assets/7662a27e-f94b-474a-b3ee-0e5ad230c7d4)

*This video was recorded using a profile with two Parkour mod versions, `v0.1.2` and `v0.2.1`: note how the client does not distinguish between the two versions (version displayed in the UI is the same for both mods), and is not capable of (de)activating any of them.*

# Description

### New `enabledmods.json` format

Since the current `enabledmods.json` format does not allow for multiple mod versions, it has been reworked.
With this PR, if "old" file format is detected, `enabledmods.json` will automatically be converted to "new" format.

##### Old format

```json
{
    "Northstar.Client": true,
    "Northstar.Coop": true,
    "Northstar.CustomServers": true,
    "Northstar.Custom": true,
    "MRVN Shake It": true,
    "Titan Payload": false,
    "Parkour": true
}
```

##### New format

```json
{
    "Northstar.Client": {
        "1.19.0": true
    },
    "Northstar.Coop": {
        "0.0.0": true
    },
    "Northstar.CustomServers": {
        "1.19.0": true
    },
    "Northstar.Custom": {
        "1.19.0": false
    },
    "Parkour": {
        "0.1.2": false,
        "0.2.1": true
    },
    "MRVN Shake It": {
        "0.0.1": true
    },
    "Titan Payload": {
        "1.2.0": false
    }
}
```

(new `enabledmods.json` format also means mod managers will have to adapt)

### Squirrel-exposed functions

* Functions were reworked to take mod version into account:
    * `NSIsModEnabled(string modName)` is now `NSIsModEnabled(string modName, string modVersion)`;
    * `NSSetModEnabled(string modName, bool enabled)` is now `NSSetModEnabled(string modName, string modVersion, bool enabled)`;
    * `NSIsModRemote(string modName)` is now `NSIsModRemote(string modName, string modVersion)`;
    * `NSGetModVersionByModName(string modName)` is now `NSGetModVersions(string modName)`;
* A lot of functions loop over the list of local mods to retrieve some mod attribute (mod description, download link etc); these functions are called one after the other in the VM, therefore making useless loops over the list of mods:
[...]

PR #826, opened 2024-11-01
refactor: Mod information

Alystrasz:
Currently, to retrieve mod information from launcher, Squirrel VM must summon methods such as `NSGetModDescriptionByModName`, `NSGetModDownloadLinkByModName`, `NSGetModLoadPriority` etc, which all loop over the list of mods until they find the correct one (using mod name as key), which is highly ineffective: this means we need as many iterations over mods list as we need information about the mod.

To solve this problem, this PR introduces a `NSGetModsInformation`, which exposes all mod information to Squirrel VM at once, using a new `ModInfo` structure.

```javascript
global struct ModInfo
{
    string name = ""
    string description = ""
    string version = ""
    string downloadLink = ""
    int loadPriority = 0
    bool enabled = false
    bool requiredOnClient = false
    bool isRemote
    array<string> conVars = []
}
```

This is meant to make reviewing https://github.com/R2Northstar/NorthstarLauncher/pull/758 and https://github.com/R2Northstar/NorthstarMods/pull/824 easier.

**(I highly discourage you to look at the github diff, which is fucked up, and suggest rather looking at the changes commit per commit!)**

**Don't merge without mods counterpart: https://github.com/R2Northstar/NorthstarMods/pull/899**

#### Testing

1. Install both launcher and mods PRs;
2. Start the game;
3. Try to break the mod menu;
4. Connect to Northstar;
5. Try to break the server browser.

### DIFF
--- a/primedev/scripts/client/scriptmodmenu.cpp
+++ b/primedev/scripts/client/scriptmodmenu.cpp
@@ -57,6 +57,68 @@ ADD_SQFUNC("array<ModInfo>", NSGetModsInformation, "", "", ScriptContext::SERVER
 	return SQRESULT_NOTNULL;
 }
 
+ADD_SQFUNC("array<ModInfo>", NSGetModInformation, "string modName", "", ScriptContext::SERVER | ScriptContext::CLIENT | ScriptContext::UI)
+{
+	const SQChar* modName = g_pSquirrel<context>->getstring(sqvm, 1);
+	g_pSquirrel<context>->newarray(sqvm, 0);
+
+	for (Mod& mod : g_pModManager->m_LoadedMods)
+	{
+		if (mod.Name.compare(modName) != 0)
+		{
+			continue;
+		}
+
+		g_pSquirrel<context>->pushnewstructinstance(sqvm, 9);
+
+		// name
+		g_pSquirrel<context>->pushstring(sqvm, mod.Name.c_str(), -1);
+		g_pSquirrel<context>->sealstructslot(sqvm, 0);
+
+		// description
+		g_pSquirrel<context>->pushstring(sqvm, mod.Description.c_str(), -1);
+		g_pSquirrel<context>->sealstructslot(sqvm, 1);
+
+		// version
+		g_pSquirrel<context>->pushstring(sqvm, mod.Version.c_str(), -1);
+		g_pSquirrel<context>->sealstructslot(sqvm, 2);
+
+		// download link
+		g_pSquirrel<context>->pushstring(sqvm, mod.DownloadLink.c_str(), -1);
+		g_pSquirrel<context>->sealstructslot(sqvm, 3);
+
+		// load priority
+		g_pSquirrel<context>->pushinteger(sqvm, mod.LoadPriority);
+		g_pSquirrel<context>->sealstructslot(sqvm, 4);
+
+		// enabled
+		g_pSquirrel<context>->pushbool(sqvm, mod.m_bEnabled);
+		g_pSquirrel<context>->sealstructslot(sqvm, 5);
+
+		// required on client
+		g_pSquirrel<context>->pushbool(sqvm, mod.RequiredOnClient);
+		g_pSquirrel<context>->sealstructslot(sqvm, 6);
+
+		// is remote
+		g_pSquirrel<context>->pushbool(sqvm, mod.m_bIsRemote);
+		g_pSquirrel<context>->sealstructslot(sqvm, 7);
+
+		// convars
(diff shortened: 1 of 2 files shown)


primedev/scripts/client/scriptmodmenu.cpp
```diff
@@ -1,106 +1,76 @@
...
-			g_pSquirrel<context>->pushbool(sqvm, mod.m_bIsRemote);
-			return SQRESULT_NOTNULL;
+			g_pSquirrel<context>->pushstring(sqvm, cvar->Name.c_str());
+			g_pSquirrel<context>->arrayappend(sqvm, -2);
 		}
-	}
-
-	return SQRESULT_NULL;
-}
+		g_pSquirrel<context>->sealstructslot(sqvm, 8);
```
RoyalBlue1, line 51: Maybe make all of this its own function and make it possible to get the modinfo for a singular mod as well?

primedev/scripts/client/scriptmodmenu.cpp
```diff
@@ -1,106 +1,76 @@
...
+		g_pSquirrel<context>->pushstring(sqvm, mod.Name.c_str());
+		g_pSquirrel<context>->arrayappend(sqvm, -2);
 	}
 
-	return SQRESULT_NULL;
+	return SQRESULT_NOTNULL;
 }
 
-ADD_SQFUNC("string", NSGetModDownloadLinkByModName, "string modName", "", ScriptContext::SERVER | ScriptContext::CLIENT | ScriptContext::UI)
+ADD_SQFUNC("bool", NSIsModEnabled, "string modName", "", ScriptContext::SERVER | ScriptContext::CLIENT | ScriptContext::UI)
```
RoyalBlue1, line 73: is this still needed when the info is still in ModInfo?

primedev/scripts/client/scriptmodmenu.cpp
```diff
@@ -151,8 +122,7 @@ ADD_SQFUNC("bool", NSIsModRequiredOnClient, "string modName", "", ScriptContext:
 	return SQRESULT_NULL;
 }
 
-ADD_SQFUNC(
-	"array<string>", NSGetModConvarsByModName, "string modName", "", ScriptContext::SERVER | ScriptContext::CLIENT | ScriptContext::UI)
+ADD_SQFUNC("array<string>", NSGetModVersions, "string modName", "", ScriptContext::SERVER | ScriptContext::CLIENT | ScriptContext::UI)
```
RoyalBlue1, line 125: shouldn't this be in modinfo as well when you can get modinfo for a singular mod as well. Then the usage in main menu should be easy as well

RoyalBlue1 (changes requested):

primedev/scripts/client/scriptmodmenu.cpp
Alystrasz: Do you mean adding a `array<ModInfo> getModInformation( string modName )` method here? That could do it.

RoyalBlue1: Yea but it dosnt need to return an array

Alystrasz: User might have multiple versions of a same mod, hence the array, don't you think?

RoyalBlue1: Yes i thought we could just add an array for the version field but thinking about it what if other fields are different between versions as well

primedev/scripts/client/scriptmodmenu.cpp
```diff
@@ -1,178 +1,153 @@
...
 		{
-			g_pSquirrel<context>->pushinteger(sqvm, mod.LoadPriority);
-			return SQRESULT_NOTNULL;
+			g_pSquirrel<context>->pushstring(sqvm, cvar->Name.c_str());
+			g_pSquirrel<context>->arrayappend(sqvm, -2);
 		}
+		g_pSquirrel<context>->sealstructslot(sqvm, 8);
+
+		// add current object to squirrel array
+		g_pSquirrel<context>->arrayappend(sqvm, -2);
```
RoyalBlue1, line 116: since this is duplicated code why not make a function like `void modToSquirrel(HSquirrelVm* sqvm,Mod& mod)` so when you change the struct you have one less place to change

RoyalBlue1 (changes requested):

primedev/scripts/client/scriptmodmenu.cpp
Alystrasz: Implemented in 9bda30cff38564a38a09023f841bba25819e268d.

PR #829, opened 2024-11-25
refactor: Move mod content related logic to dedicated package

Alystrasz:
Currently, mod manager and mod code are located in the same `modmanager.cpp` package.
To ease development iterations over the mod system, I propose to separate mod manager from mod, and export the latter to a dedicated package.


GeckoEidechse (approved):
Just moves all the Mod class and any related logic to its own file. Manually verified that there were no additional code changes done during the move.

Don't think this needs any testing.

PR #831, opened 2024-11-30
Fix memory leak in scripthttprequesthandler.cpp

xCynDev:
Adds a few calls to `freeaddrinfo()` that were missing (mostly when returning false in `IsHttpDestinationHostAllowed()`).

### Code review:

Not much, check where I've added `freeaddrinfo()`.

### Testing:

Check that it works on your end by making an HTTP call to a non-local, IPv6 address (without `-allowlocalhttp`!)
And then also make a call to a local IPv4 host, again without `-allowlocalhttp`. 

If you don't crash and the requests work, it's all good.


Alystrasz:
@xCynDev can you provide detailed instructions to test your PR?
Describing how you make HTTP calls (CLI command? Mod?) would be a good start! :rocket:

ASpoonPlaysGames (approved):
Code looks good, not much to say really. I also don't really see how this could break in testing given that we aren't storing any of the potentially freed memory anywhere.

PR #832, opened 2024-12-02
refactor: Mod information file export

Alystrasz:
Two refactor changes here (one per commit):
* Deduplication of `enabledmods.json` path (moved to header file);
* Export of `enabledmods.json` creation logic into new `ExportModsConfigurationToFile` method (with associated documentation).


NachosChipeados (approved):
Works fine, playerd an entire match with no issues

primedev/mods/modmanager.cpp
```diff
@@ -628,6 +619,28 @@ void ModManager::SearchFilesystemForMods()
 	std::sort(m_LoadedMods.begin(), m_LoadedMods.end(), [](Mod& a, Mod& b) { return a.LoadPriority < b.LoadPriority; });
 }
 
+void ModManager::ExportModsConfigurationToFile()
+{
+	m_EnabledModsCfg.SetObject();
```
GeckoEidechse, line 624: Does `m_EnabledModsCfg.SetObject();` have to be set again here?

(I don't think I fully understand what that line does... ^^)

Alystrasz: It resets the state of the `m_EnabledModsCfg` object (same as `dict = {}` in js for instance)

GeckoEidechse:
(on Holiday for a few days, maybe someone else could take care of code review to get this merged?)

NachosChipeados (approved):
Code also looks good, manually verified there were no extra changes 👍

primedev/mods/modmanager.cpp
Alystrasz: @GeckoEidechse can this remark be resolved?

PR #833, opened 2024-12-15
scripts/client: Remove uses of Autohook from `scriptbrowserhooks.cpp`

ASpoonPlaysGames:
### Code review:

### Testing:
- Check that the "force external web browser" script flag for `OpenExternalWebBrowser` still works (i.e it opens in a separate web browser instead of the origin/steam/whatever overaly)


RoyalBlue1 (approved):
code looks good and links still open in external browser

PR #834, opened 2024-12-15
server: Remove uses of Autohook from`serverchathooks.cpp`

ASpoonPlaysGames:
### Code review:

### Testing:
- Check that the chat hooks still function as expected

### DIFF
--- a/primedev/server/serverchathooks.cpp
+++ b/primedev/server/serverchathooks.cpp
@@ -59,7 +59,7 @@ static void __fastcall h_CServerGameDLL__OnReceivedSayTextMessage(CServerGameDLL
 void ChatSendMessage(unsigned int playerIndex, const char* text, bool isTeam)
 {
 	bShouldCallSayTextHook = true;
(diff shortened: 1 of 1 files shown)


Alystrasz:
@ASpoonPlaysGames could you specify which function can be called to verify chathooks still work as expected?

ASpoonPlaysGames:
> @ASpoonPlaysGames could you specify which function can be called to verify chathooks still work as expected?

The stuff documented in here I believe https://docs.northstar.tf/Modding/reference/northstar/chathooks/

Alystrasz:
> The stuff documented in here I believe https://docs.northstar.tf/Modding/reference/northstar/chathooks/

Server chat API? Client chat API?
Both?

ASpoonPlaysGames:
Probably both but mostly server ig given file name

Alystrasz (approved):
Code looks good, I confirm all chat methods (both server and client) still work as expected ([here is the test mod](https://github.com/user-attachments/files/19218880/ChathookTestMod.zip) I used to ensure correct behaviour).

PR #835, opened 2024-12-31
Fix reloading plugins

catornot:
Originally it probably worked if all plugins could be reloaded but it also just dropped all the plugins regardless of whether it can be reloaded or not.

So this pr attempts to amend that by individually reloading all the plugins instead of dropping then loading them all.

(this still doesn't really fix it since I can't get any plugin to be reloaded 💢 so that's why it's a draft)

was (wow crash)
image

now (wow still doesn't work!)
image

so basically I need some help but I think this pr is good first stepping stone. :)

ignore the screenshots before this

## how to test

with this pr if upon reloading plugins nothing crashes then that means it worked as intended.

### DIFF
--- a/primedev/plugins/pluginmanager.cpp
+++ b/primedev/plugins/pluginmanager.cpp
@@ -1,6 +1,7 @@
 #include "pluginmanager.h"
 
 #include <regex>
+#include <ranges>
 #include "plugins.h"
 #include "config/profile.h"
 #include "core/convar/concommand.h"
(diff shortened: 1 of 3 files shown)


catornot:
Seemingly it was only an issue of discord rpc not reloading, every other plugins works so this pr can be reviewed now :D

uniboi (approved):
Makes sense to reload plugins in reverse load order. Should tbh the same for unloading as well

GeckoEidechse:
not sure if this was actually tested or not

PR #836, opened 2025-01-04
Add unload and load commands for plugins

catornot:
this pr adds concommands to reload and unload plugins so that they be more easily reloaded on windows. windows will prevent the dll from being modified if it's loaded so it needs to be unload then loaded :P

also this already has #835

### DIFF
--- a/primedev/plugins/pluginmanager.cpp
+++ b/primedev/plugins/pluginmanager.cpp
@@ -1,6 +1,7 @@
 #include "pluginmanager.h"
 
 #include <regex>
+#include <ranges>
 #include "plugins.h"
 #include "config/profile.h"
 #include "core/convar/concommand.h"
@@ -111,13 +112,14 @@ bool PluginManager::LoadPlugins(bool reloaded)
 
 void PluginManager::ReloadPlugins()
 {
-	for (const Plugin& plugin : this->GetLoadedPlugins())
-	{
-		plugin.Unload();
-	}
+	NS::log::PLUGINSYS->info("Reloading plugins");
 
-	this->plugins.clear();
-	this->LoadPlugins(true);
+	for (const Plugin& plugin : this->plugins | std::views::reverse)
+	{
+		std::string name = plugin.GetName();
+		if (plugin.Reload())
+			NS::log::PLUGINSYS->warn("Reloaded {}", name);
+	}
 }
 
 void PluginManager::RemovePlugin(HMODULE handle)
--- a/primedev/plugins/pluginmanager.cpp
+++ b/primedev/plugins/pluginmanager.cpp
@@ -182,7 +182,27 @@ void ConCommand_reload_plugins(const CCommand& args)
 	g_pPluginManager->ReloadPlugins();
 }
 
+void ConCommand_unload_plugins(const CCommand& args)
+{
+	for (const Plugin& plugin : g_pPluginManager->GetLoadedPlugins() | std::views::reverse)
+	{
+		std::string name = plugin.GetName();
+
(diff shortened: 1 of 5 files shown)


primedev/plugins/pluginmanager.cpp
```diff
@@ -182,7 +182,28 @@ void ConCommand_reload_plugins(const CCommand& args)
...
+void ConCommand_unload_plugins(const CCommand& args)
+{
+	for (const Plugin& plugin : g_pPluginManager->GetLoadedPlugins() | std::views::reverse)
+	{
+		std::string name = plugin.GetName();
+
+		if (args.ArgC() >= 2 && !strcmp(name.c_str(), args.Arg(1)))
+			continue;
+
+		if (plugin.Unload())
```
Alystrasz, line 194: Might be wise to print if unloading fails, wdyt?

catornot: yeah it be wise to add a log call there

Alystrasz:
@catornot can you provide me a protocol to test your change?
Describing what *currently* does not work, but should work with this PR?

catornot:
# why
rn it everything should just work if you try using the `reload_plugins` command

this pr adds the ability to unload(`unload_plugins`) then load(`load_plugins`) individual plugins since reloading all the plugins is sometimes buggy (for like no apparent reason and since it's a dev feature, I don't think it's a big issue)

reloading plugins isn't really useful on windows since it's too fast to replace the plugins and replacing them before reloading is not possible on windows which is why manual unload and loading is required.

# how
- get a plugin that be actually reloaded (ex: [octbots](https://github.com/catornot/bp-ort/tree/master/octbots) or [recanim](https://github.com/uniboi/recanim))
- start game
- unload it with `unload_plugins <plugin name>`
- load it `load_plugins`
- profit

Alystrasz:
I've tried using this PR with the [bport plugin](https://thunderstore.io/c/northstar/p/cat_or_not/bp_ort/), but couldn't get it to work:
- `reload_plugins` seems to have no effect (should print plugin name)
- same for `unload_plugins bport` (also tested with `bp_ort` and `bp-ort`)
- `load_plugins` straight up crashes my client:

```text
[2025-02-14] [01:04:00] [PLUGINSYS] [warning] Plugin has already been loaded
[2025-02-14] [01:04:00] [PLUGINSYS] [warning] Unloading invalid plugin 'R2Northstar\plugins\bp_ort.dll'
[...]
[01:04:02] [PLUGINSYS] [errr] Failed to unload plugin at 'R2Northstar\plugins\bp_ort.dll'
[01:04:02] [PLUGINSYS] [warn] Plugin has already been loaded
[01:04:02] [PLUGINSYS] [warn] Unloading invalid plugin 'R2Northstar\plugins\DiscordRPC.dll'
```

catornot:
there might might have been a regression (I would need to confirm that myself) so now I think this pr might not work
but as for the testing: bp_ort can't even be reloaded :trollface: only [octbots](https://github.com/catornot/bp-ort/tree/master/octbots) or [recanim](https://github.com/uniboi/recanim) can be actually reloaded

I kind of dislike how system specific plugin reloading seams to be

catornot:
rn I feel this is not needed anymore (at least for me)

AllusiveWheat (approved):
Code looks good

catornot:
> Code looks good

image

AllusiveWheat:
Tested works good, seems to print a lot on invalid/already loaded plugins

catornot:
hopefully it does indeed work now

catornot:
I just found this gif in my saved ones and felt like posting it

PR #837, opened 2025-01-18
Remove cheat flag from `cl_gib_lifetime`

NachosChipeados:
There is literally 0 reason for it to have it, especially because its ragdoll counterparts `cl_ragdoll_force_fade_time_` don't. No idea what Respawn was thinking to be honest

https://github.com/user-attachments/assets/fa32f5da-c3e0-494f-addb-1d3b064e9a94

### DIFF
--- a/primedev/shared/misccommands.cpp
+++ b/primedev/shared/misccommands.cpp
@@ -365,8 +365,7 @@ void FixupCvarFlags()
 
 	const std::vector<std::tuple<const char*, const char*>> CVAR_FIXUP_DEFAULT_VALUES = {
 		{"sv_stressbots", "0"}, // not currently used but this is probably a bad default if we get bots working
-		{"cl_pred_optimize", "0"}, // fixes issues with animation prediction in thirdperson
-		{"cl_gib_lifetime", "10"} // default value (3) is too low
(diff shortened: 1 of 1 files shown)


primedev/shared/misccommands.cpp
```diff
@@ -364,7 +365,8 @@ void FixupCvarFlags()
 
 	const std::vector<std::tuple<const char*, const char*>> CVAR_FIXUP_DEFAULT_VALUES = {
 		{"sv_stressbots", "0"}, // not currently used but this is probably a bad default if we get bots working
-		{"cl_pred_optimize", "0"} // fixes issues with animation prediction in thirdperson
+		{"cl_pred_optimize", "0"}, // fixes issues with animation prediction in thirdperson
+		{"cl_gib_lifetime", "10"} // default value (3) is too low
```
catornot, line 369: not sure about this one, I would just leave on respawn's default value

NachosChipeados: alright

catornot (approved):
I don't think this needs any testing just a simple convar change

Zanieon (approved):
Tempted to just say a LGTM, because this is one of those simple PRs.

PR #841, opened 2025-03-15
Fix crashing when reading audio samples with wide characters

### LINKED ISSUE #840: Audio files with non-ascii characters crash game
Image
Image

Related: https://github.com/R2Northstar/NorthstarLauncher/pull/667

ASpoonPlaysGames:
Closes #840 

Test using https://thunderstore.io/package/download/Xiaobbuguai/BlueArchivesBeautificationPack/1.1.0/ (the updated version fixed the crashing by renaming the files) and booting the game to the main menu. Might also be worth testing that the audio overrides still work as ~~badly as~~ before.

The external console (and possible also the in-game console) will likely log these incorrectly, as they are wide chars but our logging eventually goes through to WriteConsoleA which doesn't support wide chars. Fixing this would take a bit more effort and is out of scope here.

Previous fixes for wide file paths didn't log through spdlog or anything, so they didn't end up needing to enable `SPDLOG_WCHAR_TO_UTF8_SUPPORT`


NachosChipeados (approved):
Works fine for me 👍 the sounds also play properly in-game

Before:

https://github.com/user-attachments/assets/c42f7187-1c46-458b-a29b-6e7b679138e8

After:


https://github.com/user-attachments/assets/5bd754a5-38f0-4b49-865f-1c8f2d453caf

Alystrasz (approved):
Proposed change looks good to me.

PR #842, opened 2025-03-30
Add LGBT logo option

NachosChipeados:
"They're putting something in the water and turning the logos gay!"
Screenshot 2025-03-30 18_20_09

image


GeckoEidechse (approved):
Confirmed working in quick testing. With the flag set, the logo is the LGBT one.

Code changes are fine, just adds `#define` flag and the `.ico` file itself. Logo is still blue by default (for now :>)

primedev/primelauncher/resources.rc
```diff
@@ -3,6 +3,8 @@
 #include "resource1.h"
 #include "../ns_version.h"
 
+#define USE_LGBT_LOGO 0
```
GeckoEidechse, line 6: Actually this triggers the `ifdef` below as the check is on whether `USE_LGBT_LOGO` is set and not what the value is. Whoops

NachosChipeados: I blame Squirrel

Will fix in a sec 👍

EnderBoy9217:
We adding mods into base NS now?

PR #847, opened 2025-04-12
fix: Add vanilla compatibility for `setplaylistvaroverride` byte-patch

NachosChipeados:
The exact same thing as https://github.com/R2Northstar/NorthstarLauncher/pull/687. I know that Spoon said this should be handled in a different way, but since toggling vanilla isn't a thing right now, i don't really think it matters. If that ever changes this can just be updated.

Another part of why im making this is because i don't want people to download a random `.dll` from NexusMods to fix it https://www.nexusmods.com/titanfall2/mods/62


Alystrasz:
Invoking @ASpoonPlaysGames to review.

ASpoonPlaysGames (approved):
Yeah fine, nobody is currently working on making it toggleable anyway

PR #848, opened 2025-04-17
fix: LGBT logo being enabled by default

NachosChipeados:
Fixes the LGBT logo being enabled by default

See https://github.com/R2Northstar/NorthstarLauncher/pull/842#discussion_r2047929975 for reference


Alystrasz (approved):
Tested manually, changing `USE_LGBT_LOGO` now toggles LGBT logo display as intended.

PR #849, opened 2025-04-25
Set execution markers properly

AllusiveWheat:
Set execution markers properly if there's enough room or ignore command otherwise.
Stops servers from executing commands on the client if not in vanilla.

### DIFF
--- a/primedev/client/r2client.cpp
+++ b/primedev/client/r2client.cpp
@@ -4,11 +4,10 @@ char* g_pLocalPlayerUserID;
 char* g_pLocalPlayerOriginToken;
 GetBaseLocalClientType GetBaseLocalClient;
 
-
 ON_DLL_LOAD("engine.dll", R2EngineClient, (CModule module))
 {
 	g_pLocalPlayerUserID = module.Offset(0x13F8E688).RCast<char*>();
 	g_pLocalPlayerOriginToken = module.Offset(0x13979C80).RCast<char*>();
-	GetBaseLocalClient = module.Offset(0x78200).RCast<GetBaseLocalClientType>();
 
+	GetBaseLocalClient = module.Offset(0x78200).RCast<GetBaseLocalClientType>();
 }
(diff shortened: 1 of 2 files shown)


EladNLG (commented):
Breaks the game, used by NS itself along with a couple of vanilla functions. Also, already fixed [here](https://github.com/R2Northstar/NorthstarLauncher/blob/1277b3711d3aedb48e83d854e3c734c2f368c7ef/primedev/shared/exploit_fixes/exploitfixes.cpp#L282).

AllusiveWheat:
https://github.com/user-attachments/assets/d6eb95a5-a2a4-4581-88db-976d2cc5b8e7

primedev/client/r2client.cpp
```diff
@@ -4,10 +4,11 @@ char* g_pLocalPlayerUserID;
...
 
+
 ON_DLL_LOAD("engine.dll", R2EngineClient, (CModule module))
 {
 	g_pLocalPlayerUserID = module.Offset(0x13F8E688).RCast<char*>();
 	g_pLocalPlayerOriginToken = module.Offset(0x13979C80).RCast<char*>();
-
 	GetBaseLocalClient = module.Offset(0x78200).RCast<GetBaseLocalClientType>();
+
 }
```
GeckoEidechse, line 14: Note to self to reset whitespace changes before merge.

GeckoEidechse (approved):
Confirmed working in quick testing. Joining a server with the exploit that calls `quit` causes unpatched client to quit but with patch the `quit` command is no longer accepted.

primedev/client/r2client.cpp
GeckoEidechse: Resolved with 28b16a0d26deb211a5718284fc6632ec08b6bd04

GeckoEidechse:
Written in vc together with @RoyalBlue1

PR #851, opened 2025-05-15
Add mod version argument to `NSSetModEnabled` function

Alystrasz:
With this PR, the `NSSetModEnabled` function enables a given version of the input mod only, and not all of its versions (which would crash due to files conflicts).

Check mods counterpart PR for more details.

### DIFF
--- a/primedev/scripts/client/scriptmodmenu.cpp
+++ b/primedev/scripts/client/scriptmodmenu.cpp
@@ -92,15 +92,16 @@ ADD_SQFUNC("array<string>", NSGetModNames, "", "", ScriptContext::SERVER | Scrip
 	return SQRESULT_NOTNULL;
 }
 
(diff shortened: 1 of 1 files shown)


ASpoonPlaysGames:
Worth noting that this change of API will break existing mods

Alystrasz:
> Worth noting that this change of API will break existing mods

Do you know of any mod that enables/disables mods using the `NSSetModEnabled` function?
I assumed it was only used by Northstar core mods

ASpoonPlaysGames:
Not off the top of my head, but any mod that modifies those files and stuff will have a bad time

ASpoonPlaysGames:
I'm not saying to not merge this or anything, just that it should be a minor update instead of a patch

Alystrasz:
Since changes proposed in this PR are included in #828, which was tested and approved, this PR is no longer needed.

PR #854, opened 2025-05-19
determine calling plugin by return address

uniboi:
With the old system it was possible to act in the name of a different plugin. For example it was possible to log something using DiscordRPC's sink by just passing the DiscordRPC handle to INSSys::Log.

To test just start the game and observe that DiscordRPC logging still works


ASpoonPlaysGames:
Why not bump the interface version number and remove the unused parameters? Ig there is a concern of needing people to update their stuff

ASpoonPlaysGames (approved):
Code looks good

PR #860, opened 2025-06-09
Add ImGui rendering for debug and stuff

ASpoonPlaysGames:
Still a WIP

Can be accessed from `http:localhost:5000` while the game is running (and not in the boot screen)


ASpoonPlaysGames:
this was mostly just for testing, and at least doing the websockets stuff introduces tonnes of dependencies, not worth.

PR #863, opened 2025-07-18
Reduce templating of squirrel managers

ASpoonPlaysGames:
The amount of templating we have for squirrel managers (`g_pSquirrel`) means that the type of `g_pSquirrel<context>` used in almost all native squirrel functions is left ambiguous, requiring developers to provide sample template arguments for intellisense to be able to provide syntax highlighting, autocomplete etc.

I didn't really see much benefit to having this as a template, instead of just an array which clears this ambiguity.

This PR should be pretty heavily scrutinised for small mistakes that may result in copying things, dangling pointers, that sort of thing. I think it's fine but it could easily be subtly broken I think.

### DIFF
--- a/primedev/masterserver/masterserver.cpp
+++ b/primedev/masterserver/masterserver.cpp
@@ -363,9 +363,6 @@ void MasterServerManager::RequestMainMenuPromos()
 	std::thread requestThread(
 		[this]()
 		{
-			while (m_bOriginAuthWithMasterServerInProgress || !m_bOriginAuthWithMasterServerDone)
-				Sleep(500);
-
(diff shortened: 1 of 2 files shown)


ASpoonPlaysGames:
Had to deal with things that weren't merge conflicts but everything should be good now

uniboi (commented):
Some thoughts that aren't relevant to this PR

ASpoonPlaysGames:
> I guess there's no way to properly test this, apart from playing a few matches?

Yeah pretty much

ASpoonPlaysGames:
> Crashes on start

Should be fixed now

PR #864, opened 2025-08-05
Fix dependency constant registration

ASpoonPlaysGames:
Bad order of operations, was populating and them immediately clearing. Probably my bad, sorry


Alystrasz (approved):
Successfully tested by @NekoUwUGamerfnf, merging.

PR #866, opened 2025-08-07
add initdata size for backwards / forward compat

uniboi:
adding the size allows safely adding members to the InitData struct with backwards compatibility https://github.com/torvalds/linux/blob/master/include/uapi/linux/fs.h#L562-L563


catornot (approved):
Good idea.

Alystrasz (approved):
Simply adding a field to a struct should be fine, right?

catornot:
> Simply adding a field to a struct should be fine, right?
yes, since plugins don't create this struct increasing the size of it is fine and actually this change would allow plugins to detect older versions of the this struct.

Issue #114, opened 2022-03-25
Use snprintf + proper error checking for printing version number


GeckoEidechse:
It might be a good idea to replace this with a snprintf + proper error checking in the future, but it won't overflow with any reasonable version (and we control that), so it's probably ok to leave it for now.

_Originally posted by @pg9182 in https://github.com/R2Northstar/NorthstarLauncher/pull/113#discussion_r835671011_

pg9182:
Holy shit, that's a long title...

PR #867, opened 2025-08-14
Replace sprintf usage with sprintf_s in version.cpp

AllusiveWheat:
Fixes https://github.com/R2Northstar/NorthstarLauncher/issues/114


ASpoonPlaysGames (approved):
Code LGTM, makes sense and closes an old issue

PR #871, opened 2025-10-06
Add FCVAR_REPLICATED flag to sv_alltalk

NachosChipeados:
For https://github.com/R2Northstar/NorthstarMods/pull/935

Without this, clients don't know whether alltalk is enabled or not


catornot:
This should probably be fine but still should tested between different version. aka if an older client client can connect to a server running this version. I am pretty sure that one bob pr should have fixed this problem but better safe than sorry.

PR #872, opened 2025-11-07
Add model reloading

AllusiveWheat:
Adds a new command to reload all models.

### DIFF
--- a/primedev/engine/models.cpp
+++ b/primedev/engine/models.cpp
@@ -10,7 +10,7 @@ void ConCommand_reload_models(const CCommand& args)
 	*g_VpkMode = prevVpkMode;
 };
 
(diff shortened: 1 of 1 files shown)


primedev/engine/models.cpp
```diff
@@ -0,0 +1,20 @@
+static void (*Studio_ReloadModels)(__int64 a1, unsigned int a2) = nullptr;
+static uintptr_t model_loader;
+
+void ConCommand_reload_models(const CCommand& args)
+{
+	CModule filesystem_dll = CModule("filesystem_stdio.dll");
+	char* g_VpkMode = filesystem_dll.Offset(0xe5aa9).RCast<char*>();
+	char back = *g_VpkMode;
```
catornot: the char could be a uint8_t instead.

catornot (approved):
It works (just tested)

PR #878, opened 2025-11-27
Allow for appending to particles_manifest.txt

ASpoonPlaysGames:
Don't use keyvalues, just put it as if you are overriding it. I don't see a case where you want to be overriding all the particles in the game, so yeah.


EM4Volts:
Tried testing this yesterday, I believe it worked but I think confirmation from someone like s2 would be good.

PR #881, opened 2025-12-04
Prevent mod downloading being used in vanilla

ASpoonPlaysGames:
Supercedes #636

Should add checks to all uses of `g_pModDownloader` and give reasonable return values for the squirrel functions


Alystrasz (approved):
I tried :
- starting vanilla+ without this PR: mod downloader got initialized;
- starting vanilla+ with this PR: mod downloader did not get initialized;
- installing this PR, started northstar v1.31.4: I could download latest Onslaught mod without an issue.

Code looks good and seems to behave as expected.

PR #882, opened 2025-12-07
Don't require auth for requesting mainmenupromos

ASpoonPlaysGames:
Worth noting that due to a race condition between executing autoexec_ns_client and calling this function, the main menu will fail to get mainmenupromos the first time.

Should be merged alongside https://github.com/R2Northstar/NorthstarMods/pull/950


pg9182 (commented):
Looks good.

ASillyNeko:
> Looks good.

Idk if it should be merged right now maybe later.

PR #885, opened 2025-12-27
Prevent invalid packet flooding

sonny-tel:
This adds a workaround to the fact that flooding `NET_ReceiveDatagram` with invalid packets, (namely invalid LZSS compression (-3) packets see `engine.dll+0x21B735`) causes the server to stop receiving that frame. This adds a configurable number of times to iterate listening per call from `NET_GetPacket`.

Basically just pasted from [R1Delta security_fixes.cpp](https://github.com/r1delta/r1delta/blob/f04cc67e6247c5e15633b95f84fc69eca7252dc6/engine/security_fixes.cpp#L811)
### Code review:
🙂 

### Testing:
idk probably works 👍


ASillyNeko:
image

ASillyNeko:
image

ASillyNeko:
The formatter is gonna kill you.

PR #886, opened 2025-12-29
Improve compile error messages

ASpoonPlaysGames:
In the console, give the name of the mod that probably caused it, and explain why this happens. Should make it a bit easier for users to troubleshoot without having to dig through files?


AllusiveWheat (approved):
Adds more detail to the script error logs.

PR #889, opened 2026-01-03
MAD: Allow for versioning MWS mods

ASpoonPlaysGames:
Extracts MWS mods to the right directory hopefully.

A MWS mod that has been downloaded should have the `mod.json` found in `remote/mods/author.modname-version/mod.json`

CC @EM4Volts


EM4Volts:
Will test when I'm home over the next days

PR #892, opened 2026-01-13
Fix bots crashing dedicated servers without players

ASillyNeko:


ASpoonPlaysGames (approved):
Conceptually code looks fine, idk if this is the correct logic to be doing here though, @catornot will know more probably

PR #895, opened 2026-02-03
Fix dedicated server playlists_v2 fatal error

ASillyNeko:
Fixes mods overriding `playlists_v2` causing a fatal error on dedicated servers.


AllusiveWheat:
Just tested, using the vanilla playlists file from stryder it still crashes.

ASpoonPlaysGames:
Rip @ASillyNeko

PR #898, opened 2026-03-07
Add clan tag support

86Yin:
Need [@R2Northstar/Atlas/pull/45](https://github.com/R2Northstar/Atlas/pull/45)


catornot (commented):
I think it looks fine and it's fine like this since no one is working on communities support anyway.
(won't be testing it rn since the [atlas](https://github.com/R2Northstar/Atlas/pull/45) side isn't merged yet)

AllusiveWheat:
Code is fine overall, but clan tags should probably support local listen servers as well as remote ones. Currently it only sets clan tags on remote servers

PR #904, opened 2026-03-24
Remove unnecessary usage of AsyncCall in ModConCommandCallback_Internal

### LINKED ISSUE #902: Unnecessary usage of AsyncCall in ModConCommandCallback_Internal
This call to `AsyncCall` can be replaced by a normal `Call`because concommands are executed on the game thread.
Proof: safe usage of a sync squirrel call can be found in [bp_ort](https://github.com/catornot/bp-ort/blob/ac680ec9219ccd4191f1a60f0d1b20232dbf3c8e/bp_ort/src/bots/debug_commands.rs#L298-L304) where the concommand does work fine :)

catornot:
Quite trivial and the explanation for this is in the issue : #902 
would in fact close #902

### DIFF
--- a/primedev/mods/modmanager.cpp
+++ b/primedev/mods/modmanager.cpp
@@ -42,7 +42,7 @@ template <ScriptContext context> void ModConCommandCallback_Internal(std::string
 	{
 		if (command.ArgC() == 1)
 		{
-			g_pSquirrel[context]->AsyncCall(name);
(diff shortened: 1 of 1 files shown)


AllusiveWheat (approved):
While this is fine. It could be useful to unify Call and AsyncCall to both either use `const char*` or `std::string`

PR #910, opened 2026-04-08
Add .dat and .csv to save file extension whitelist

superwfox:
### Changes

Extends the `whitelist` set in `modsavefiles.cpp` from `{".txt", ".json"}` to `{".txt", ".json", ".dat", ".csv"}`.

### Motivation

The save file system writes plain text regardless of extension. However, `.txt` and `.json` files are easily recognized by users, who may be tempted to open and hand-edit them out of curiosity — potentially breaking the file format and rendering the save data unusable.

Allowing `.dat` and `.csv` gives mod developers the option to use less “inviting” extensions (especially `.dat`) to discourage casual edits by end users, reducing the likelihood of accidental data corruption and related support issues.

### Security Consideration

Since the underlying write mechanism is identical (plain text via `std::ofstream`), `.dat` and `.csv` carry no additional risk. All existing safeguards — path validation, folder size limits, and NUL character checks — remain in effect.

### DIFF
--- a/primedev/mods/modsavefiles.cpp
+++ b/primedev/mods/modsavefiles.cpp
@@ -83,7 +83,7 @@ template <ScriptContext context> void SaveFileManager::SaveFileAsync(fs::path fi
 			}
 
 			// If there's a file extension missing here that you need, feel free to make a PR adding it
-			static const std::set<std::string> whitelist = {".txt", ".json"};
+			static const std::set<std::string> whitelist = {".txt", ".json",".dat",".csv"};
 
 			// Check if file extension is whitelisted
 			std::string extension = file.extension().string();
--- a/primedev/mods/modsavefiles.cpp
+++ b/primedev/mods/modsavefiles.cpp
@@ -83,7 +83,7 @@ template <ScriptContext context> void SaveFileManager::SaveFileAsync(fs::path fi
 			}
 
 			// If there's a file extension missing here that you need, feel free to make a PR adding it
-			static const std::set<std::string> whitelist = {".txt", ".json",".dat",".csv"};
+			static const std::set<std::string> whitelist = {".txt", ".json", ".dat", ".csv"};
 
 			// Check if file extension is whitelisted
 			std::string extension = file.extension().string();


EM4Volts:
safeio is supposed to be safe, and at the same time also should give users a feeling of saefety, i dont like the idea of writing binary files or files that are not the usual extensions, people may see them and feel unsafe.
should be a plugin tbh.

_Sent from my Xiami Redmi Note 9 Pro 5G_

superwfox:
> tbh.

How about a denyList?

By the way,Will the doc get updates so developers could acknowledge that only .txt and .json are available

catornot:
> How about a denyList?

that is certainly not the right direction. 

ideally there wouldn't even be any choice of file extensions. users are unlikely to edit those files, since they are hidden away in a sub directory of R2Northstar, where normal users won't be looking either way. 

either way this api was designed for scripts to be as isolated as possible from the system, so file extensions don't really matter when you only have way of reading or writing which is with a string.

superwfox:
> > How about a denyList?
> 
> 
> 
> that is certainly not the right direction. 
> 
> 
> 
> ideally there wouldn't even be any choice of file extensions. users are unlikely to edit those files, since they are hidden away in a sub directory of R2Northstar, where normal users won't be looking either way. 
> 
> 
> 
> either way this api was designed for scripts to be as isolated as possible from the system, so file extensions don't really matter when you only have way of reading or writing which is with a string. 
> 
> 
> 
> 

The reason I’m proposing this is based on a real problem I encountered. While developing [TTDMRecorder](https://new.thunderstore.io/c/northstar/p/Sudark/TTDMRecorder/), I ran into some bad actors. They discovered the temporary storage location I was using for data — which was stored in plaintext CSV at the time — and modified their match records to absurdly inflated values to show off. I wanted the platform to maintain a reasonable level of credibility, but once other users caught on, they started doing the same, flooding the database with junk data. It was a really frustrating few days. That experience pushed me to come up with this approach, but I found that .bat files are no longer allowed in newer versions, which is why I’m submitting this PR.

catornot:
> > > How about a denyList?
> > 
> > 
> > that is certainly not the right direction.
> > ideally there wouldn't even be any choice of file extensions. users are unlikely to edit those files, since they are hidden away in a sub directory of R2Northstar, where normal users won't be looking either way.
> > either way this api was designed for scripts to be as isolated as possible from the system, so file extensions don't really matter when you only have way of reading or writing which is with a string.
> 
> The reason I’m proposing this is based on a real problem I encountered. While developing [TTDMRecorder](https://new.thunderstore.io/c/northstar/p/Sudark/TTDMRecorder/), I ran into some bad actors. They discovered the temporary storage location I was using for data — which was stored in plaintext CSV at the time — and modified their match records to absurdly inflated values to show off. I wanted the platform to maintain a reasonable level of credibility, but once other users caught on, they started doing the same, flooding the database with junk data. It was a really frustrating few days. That experience pushed me to come up with this approach, but I found that .bat files are no longer allowed in newer versions, which is why I’m submitting this PR.

bad actors would always be able to modify stuff on the client, making it harder or less obvious wouldn't discourage them. I can't really say if your model with that mod is flawed since I don't know how it works, but I would advise keeping your stats data in the sqvm at least or better yet on the server (if it can be a server-side mod).

ASpoonPlaysGames:
Yeah changing the file format isn't going to stop anyone from messing with the contents. You'd be better off storing it on the server, or better yet on a proper webserver using http requests

catornot:
Realistically this change doesn't anything since scripts don't really care about the file format, since it's just all strings, but this does add the liability of maintaining access to these file formats in the future. Therefore I think it's better to close this then.

PR #911, opened 2026-04-13
Pin clang version to 16 for formatting in the flake

catornot:
Apparently clang versions matter massively for formatting and since the ci uses clang 16 setting the clang-format version in the flake fixes any formatting differences. this does require adding an older nixpkgs version since newer version no longer package clang 16 lol

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -34,6 +34,22 @@
         "type": "github"
       }
     },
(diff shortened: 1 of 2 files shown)


Alystrasz (approved):
LGTM

PR #913, opened 2026-04-13
Allow external authentication methods to work on linux

catornot:
(the origin sdk hooks for maxima)
didn't really test much

possibly doesn't even need the second hook


AllusiveWheat (approved):

AllusiveWheat:
Just tested on windows, works well.

PR #914, opened 2026-04-17
fix crash when shooting outside map bounds

AllusiveWheat:
Removes the call to Error when a bullet is fired outside of the map bounds.


NachosChipeados (approved):
LGTM 👍. Also shot at some grunts for a little bit just for good measure, and it works fine

https://github.com/user-attachments/assets/b7b4031c-5958-4ad5-8c4e-953b90884e5a

catornot:
merging based on reviews :)

PR #915, opened 2026-04-17
Block empty chat messages

86Yin:
Intercept and ignore empty chat messages to prevent certain mods from crashing.


ASpoonPlaysGames (approved):
Yeah makes sense to me. Doesn't necessarily stop empty *looking* chat messages but presumably mods can handle that

PR #916, opened 2026-04-27
Don't include the various libcurl test build targets

ASpoonPlaysGames:
We don't use them, they are annoying, and they clog up the build targets list significantly.


Alystrasz (approved):
Looks good; played a few games and downloaded mods without an issue.

PR #918, opened 2026-05-09
Make pakLoadApi global

AllusiveWheat:
Makes g_pakLoadApi global


EM4Volts:
lgtm, format check fails but format check is just a hater anyway

catornot (approved):
Simple struct visibility refactor shouldn't need any testing so it's all good

PR #919, opened 2026-05-10
various materialsystem things

EM4Volts:
Overhauls the nsdxcustombuffer stuff.
instead of a single function there is now 3.

Registering a material so it uses the buffer:
`NSRegisterCustomDXBufferForGUID(string MaterialGUID) `

Updating the buffer for a material:
`NSUpdateCustomDXBufferForGUID(string MaterialGUID, array bufferData) `

Deregistering the material from using the buffer:
`NSDeregisterCustomDXBufferForGUID(string MaterialGUID) `

Reason for these changes are to make it cleaner, but at the same time to remove the need for the glueflag in the material when making it.
The previous need for the glueflag interfered with basically all wld materials that need the flag to bind certain lighting related features.

This also adds another function.

Binding a Texture, via its GUID in the rpak system to a specified slot in the shader of said material
`NSBindTextureToMaterial(string rPakMaterialGUID string rPakTextureGUID int shaderBindingSlot)`

this will not properly stream the texture if its a streamable texture and instead use its highes permed mipmap.

At the same time, since this messes with `cmaterialglue.h` i also overhauled the mat_crosshair_printmaterial output to make it more informal, and usable, instead of being just a bit of a fun command.

OLD:
Screenshot_2026-05-08_223039

NEW:
image

### DIFF
--- a/primedev/materialsystem/nscustomdxbuffer.cpp
+++ b/primedev/materialsystem/nscustomdxbuffer.cpp
@@ -6,6 +6,8 @@
 #include <cctype>
 #include <string>
 #include <cstdint>
+#include "core/filesystem/rpakfilesystem.h"
+#include "cmaterialglue.h"
 
 static ID3D11DeviceContext** DeviceContext;
 static ID3D11Device** D3D11Device_14E8DD0;
@@ -15,6 +17,12 @@ struct Ns_Constant_Buffer
 	float data[320];
 };
 
+struct MaterialTextureMappings
+{
+
+	std::array<uint64_t, 30> slots;
+};
+
 AUTOHOOK_INIT()
 
 static Ns_Constant_Buffer NSCustomDXBuffer;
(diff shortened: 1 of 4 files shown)


primedev/materialsystem/nscustomdxbuffer.cpp
```diff
@@ -136,16 +243,83 @@ template <ScriptContext context> SQRESULT NSSetCustomDXBuffer(HSQUIRRELVM sqvm)
 	return SQRESULT_NULL;
 }
 
+template <ScriptContext context> SQRESULT NSBindTextureToMaterial(HSQUIRRELVM sqvm)
+{
+
+	auto rPakMaterialGUIDString = (g_pSquirrel[ScriptContext::CLIENT]->getstring(sqvm, 1));
+	auto rPakTextureGUIDString = (g_pSquirrel[ScriptContext::CLIENT]->getstring(sqvm, 2));
+	auto rPakShaderSlotBindingInt = (g_pSquirrel[ScriptContext::CLIENT]->getinteger(sqvm, 3));
+
```
ASpoonPlaysGames, line 252: Either un-hardcode from CLIENT or remove the template parameter from the function

primedev/materialsystem/nscustomdxbuffer.cpp
```diff
@@ -107,7 +214,7 @@ template <ScriptContext context> SQRESULT NSSetCustomDXBuffer(HSQUIRRELVM sqvm)
 
 	if (!isValidMaterialGUID(guidString))
 	{
-		g_pSquirrel[ScriptContext::SERVER]->raiseerror(
+		g_pSquirrel[ScriptContext::CLIENT]->raiseerror(
```
ASpoonPlaysGames, line 217: Unhardcode CLIENT or remove template from function

primedev/materialsystem/nscustomdxbuffer.cpp
```diff
@@ -86,7 +127,73 @@ bool isValidMaterialGUID(const std::string& str)
...
+		NS::log::SCRIPT_CL->warn(
+			"Attempted to register GUID: {} to the NSCustomDXBuffer system, GUID was already registered", GUIDMaterialGlue_short->guid);
+	}
+	return SQRESULT_NULL;
+}
+
+template <ScriptContext context> SQRESULT NSDeregisterCustomDXBufferForGUID(HSQUIRRELVM sqvm)
+{
+
+	auto rPakMaterialGUIDString = (g_pSquirrel[ScriptContext::CLIENT]->getstring(sqvm, 1));
```
ASpoonPlaysGames, line 166: Unhardcode CLIENT, or remove template from function

primedev/materialsystem/nscustomdxbuffer.cpp
```diff
@@ -86,7 +127,73 @@ bool isValidMaterialGUID(const std::string& str)
...
+		return SQRESULT_ERROR;
+	}
+
+	auto* base = reinterpret_cast<uint8_t*>(AssetFromGUID);
+	auto* GUIDMaterialGlue_short = reinterpret_cast<CMaterialGlue_short*>(base + 16);
+	// we need to add 16 to the pointer, matglueshort is matglue without the first 16 bytes. GetAssetByHash returns a pointer to the full
+	// matglue.
+	if (NSRegisteredCustomBufferMaterials.contains(GUIDMaterialGlue_short->guid))
+	{
+		NS::log::SCRIPT_CL->info("Deregistered GUID: {} from the NSCustomDXBuffer system", GUIDMaterialGlue_short->guid);
```
ASpoonPlaysGames, line 184: get this dynamically based on `context` (if you keep the templating)

primedev/materialsystem/nscustomdxbuffer.cpp
```diff
@@ -86,7 +127,73 @@ bool isValidMaterialGUID(const std::string& str)
 	return true;
 }
 
-template <ScriptContext context> SQRESULT NSSetCustomDXBuffer(HSQUIRRELVM sqvm)
+template <ScriptContext context> SQRESULT NSRegisterCustomDXBufferForGUID(HSQUIRRELVM sqvm)
+{
+
+	auto rPakMaterialGUIDString = (g_pSquirrel[ScriptContext::CLIENT]->getstring(sqvm, 1));
```
ASpoonPlaysGames, line 133: Unhardcode CLIENT or remove template from function

primedev/materialsystem/nscustomdxbuffer.cpp
```diff
@@ -86,7 +127,73 @@ bool isValidMaterialGUID(const std::string& str)
...
+	if (!AssetFromGUID)
+	{
+		g_pSquirrel[ScriptContext::CLIENT]->raiseerror(
+			sqvm, fmt::format("Asset with GUID {} Doesnt Exist", rPakMaterialGUIDString).c_str());
+		return SQRESULT_ERROR;
+	}
+
+	auto* base = reinterpret_cast<uint8_t*>(AssetFromGUID);
+	auto* GUIDMaterialGlue_short = reinterpret_cast<CMaterialGlue_short*>(base + 16);
+	// we need to add 16 to the pointer, matglueshort is matglue without the first 16 bytes. GetAssetByHash returns a pointer to the full
```
ASpoonPlaysGames: I'd prefer if the comment was above the line of code that it's talking about, not below

primedev/materialsystem/nscustomdxbuffer.cpp
ASpoonPlaysGames, line 1: Is there a reason why the squirrel functions are CLIENT only? Would it not be valid to control the materials through UI sometimes? (doesn't UI run every frame and CLIENT not?)

ASpoonPlaysGames (changes requested):
A lot of the same comment mostly. Various places assume CLIENT whilst being templated. They either shouldn't be templated or they shouldn't assume CLIENT

primedev/materialsystem/nscustomdxbuffer.cpp
EM4Volts: will try to adress the other comments later on, just saying that client was originally done because i just thought its the way it should be done, as most stuff around model asset stuff also used client, and i never had an issue of it having problems from doing all the stuff on client since the actual action doesnt really happen in the vm but in shaderexecute, the vm just pushes the data.
since we have runclientfunction (or whatever its called) to run from ui to client its pretty easy to run from ui anyway, i just thought it makes more sense on client

primedev/materialsystem/nscustomdxbuffer.cpp
EM4Volts: the original pr (that is already merged in ns) did this too, i personally do not see a problem with it, it works just fine.

ASpoonPlaysGames (approved):
LGTM

PR #922, opened 2026-05-23
Fix `weapon_reparse` not invalidating compiled weapon files

ASpoonPlaysGames:
Should make `weapon_reparse` always cause a re-compilation of compiled (keyvalued) weapon txt files.

I doubt this will fix the issues with keyvalued files and networking, but I'm not sure if they ever worked as expected tbh? I can't find a commit that it worked with at least


AllusiveWheat (approved):
Code looks good

EM4Volts:
works for making weapon_reparse work properly again.

PR #923, opened 2026-05-25
Add support for ScriptFile and GlobalFunction dependencies

86Yin:


catornot:
what's the motivation behind this? seems kind of pointless when instead mod versions could be provided #685

ASpoonPlaysGames:
Worth mentioning that this approach will also break all existing mods that use dependency constants

86Yin:
> what's the motivation behind this? seems kind of pointless when instead mod versions could be provided #685

[cl_chat_command.nut](https://github.com/86Yin/V/blob/main/mod/scripts/vscripts/callbacks/cl_chat_command.nut)

EM4Volts:
> > what's the motivation behind this? seems kind of pointless when instead mod versions could be provided #685
> 
> [cl_chat_command.nut](https://github.com/86Yin/V/blob/main/mod/scripts/vscripts/callbacks/cl_chat_command.nut)

so you want it to make sure a function, added by a mod exists?
why not just add the mod as a dependency then and check for that

catornot:
If you want to target different versions of northstar or some mods, the versioned dependencies pr should be revived instead. If it's not enough, I think this implementation is not generic enough, since it only considers exposed functions by northstar and init functions for mods. hooking into the squirrel compiler should be the actual solution for this.

ASpoonPlaysGames:
Possibly a better way of implementing this could be to implement `#if defined(thing)`? though idk how that would interact with load order (probably would be undefined if your script loads before, and defined if after)

shouldn't be too hard i think? the `defined` function would look at the root table and see if the item exists

ASpoonPlaysGames:
as a bit of an example of what I was thinking:
```
// this would be defined in native
// ideally this would be a function marked as constexpr (but that doesn't exist in squirrel at the very least, might do in native?)
const table defined = { thing = true }

void function Test()
{
// kinda shit that you have to use a string here, but i mean, its not the worst
#if defined["thing"]
	printt("THING")
#endif
#if !defined["thing"]
	printt("NOT THING")
#endif
#if !defined.thing
	printt("NOT THING")
#endif
// a non existant entry throws a compile error unfortunately (expected a constant expression)
#if !defined.stuff
	printt("NOT STUFF")
#endif
}
```

I'm guessing `#if` just looks at the `consttable` (`getconsttable()`) which is why the fun things don't work.

If global functions could be added to the consttable then this could work

86Yin:
> If you want to target different versions of northstar or some mods, the versioned dependencies pr should be revived instead. If it's not enough, I think this implementation is not generic enough, since it only considers exposed functions by northstar and init functions for mods. hooking into the squirrel compiler should be the actual solution for this. 

I just went ahead and built it as soon as the idea came to me, so I honestly didn’t give those points much thought at the time.

catornot:
You didn't have to close this pr, the pr just needs a few improvements :)

PR #924, opened 2026-06-01
Allows for custom gamemode text for ffa rui

AllusiveWheat:
Allows mod authors to set the statusText var of the ffa gamemode state, replacing the hardcoded ffa text.

Example 
image


primedev/rtech/rui.cpp
```diff
@@ -0,0 +1,176 @@
+#include "rui.h"
+
+struct gamestate_info_ffa_struct
```
catornot, line 3: why is this snake case?

primedev/rtech/rui.cpp
```diff
@@ -0,0 +1,176 @@
...
+	float otherTeamScoreDiff;
+	DWORD factionImageHandle;
+	const char* scoreString;
+	const char* rightTeamScoreString;
+};
+
+using gamestate_info_ffa_t = void(__fastcall*)(RuiFunctions_t*, RuiGlobals*, RuiInstance*, gamestate_info_ffa_struct*);
+gamestate_info_ffa_t o_gamestate_info_ffa = nullptr;
+
+void h_gamestate_info_ffa(RuiFunctions_t* a1, RuiGlobals* a2, RuiInstance* a3, gamestate_info_ffa_struct* a4)
```
catornot: could you give these a* vars some proper names

catornot (changes requested):

primedev/rtech/rui.cpp
AllusiveWheat: Original exported func in ida is in snake case. just prefixing original and hooked versions accordingly.

primedev/Northstar.cmake
```diff
@@ -160,6 +160,7 @@ add_library(
     "squirrel/squirrelclasstypes.h"
     "rtech/pakfile.h"
     "rtech/pakfile.cpp"
+    "rtech/rui.cpp"
```
catornot, line 164: huh why is this added twice?

catornot (approved):
looks good now

PR #926, opened 2026-07-30
-noshaderapi and -nowindow

LightBlueCube:
this contains llm generated code (im sorry)

only fully tested on NorthstarCN, not sure is it works same on here

all changes are required args to activate, otherwise its will runs vanilla northstar behavior (i hope so)

-noshaderapi

before:

粘贴的文件2

after:

粘贴的文件

but reports from windows user says its seems like no difference

and also i guess we dont need d3d11 stub anymore

-nowindow

粘贴的文件

粘贴的文件2

runs dedi without any graphic environment, no xvfb needs anymore

if you want more infomation about the docker images: https://github.com/LightBlueCube/ns-dedi-docker


catornot:
we never needed `xvfb-run` for dedicated servers anyway, tho if this reduces memory it would be good.

LightBlueCube:
> we never needed `xvfb-run` for dedicated servers anyway, tho if this reduces memory it would be good.

https://github.com/pg9182/northstar-dedicated/blob/master/src/nswrap/nswrap.c#L183

i think we need?

sorry, its should be only needed if linux, windows doesnt have to care about it

but still if not creating window, should be help for resources cost

catornot:
> i think we need?

that is the old docker image, the newer one has graphics drivers patched out.

> but still if not creating window, should be help for resources cost

lower resource cost is good, so I will actually review once I get back to my desktop later this week :)

catornot (approved):
just trusting that the patches are good :)

well based on the fact that it does what advertised

PR #930, opened 2026-08-08
Offline LAN mode with a command-line argument

louvenarde:
Hello friends! 

Once a year I run a LAN party at my place and I play LAN games with my friends. If a game is not LAN compatible, e.g. it shipped with always-on sign in requirements, I make it LAN compatible. 

As an avid Titanfall 2 player I like to think I'll be able to play this game with my friends until the heat death of the universe and I've been wanting to play it on offline LAN for a while. However this is made impossible by the Origin sign on requirements, which necessitate an online connection.

For this reason, I have implemented a `-lan` flag on Northstar, which does the following:
- OriginSDK calls are proxied and replaced by local information. For instance, the persona name becomes the Windows username.
- Persistent data is stored locally in %appdata%, so you can still edit your classes and weapons.
- All communications with matchmaking services and other online services are cut - the game works standalone with no external requirement.

image

I usually work open source so I wanted to share this work here so it can be merged upstream, hopefully. However, *do not feel like you have to do it* . If you feel like this goes against the project guidelines, you are free to close it. I have talked about this change with @ASpoonPlaysGames to make sure it was okay to open a pull request, but I will understand if you don't want to merge these changes in.

After all, this de facto works as an Origin bypass - albeit a very limited one, because you get no master server, no server browser, no community activity, no ICE. Only direct connection to your friend's IP address 🔥

All reverse engineering here was performed white room style with public information, I took OriginSDK.dll from another game to compare and work out the bindings.

### Code review:
- The game should work exactly as normal when the `-lan` flag is off. Only when that flag is on should the game work offline. That means this feature must not have any side effect on the normal, default northstar experience.
- With `-lan` enabled, the game should not try to communicate with the outside world (no poll, no post, no get, nothing) except when connecting to a game
- The new hooks should not step over existing hooks
- I did my best to follow the project's style and reuse existing infra, if I fucked up somewhere please point it out :P

### Testing:
The game should not try to launch Origin nor try to communicate with it if it's launched. There should be no crash, no origin error, no brutal disconnection or instability that does not also happen in normal gameplay.


louvenarde:
image

louvenarde:
well you know what, let's call it for a review and you'll tell me what I need to change :)

catornot:
the consensus seems to be either no path for a merge at all or if you remove the origin bypass since it otherwise northstar would be allowing piracy. the local persistence stuff is cool tho :)

I am in the "remove the origin bypass then it's fine" group

louvenarde:
Proxying origin is unfortunately necessary to make the game work fully offline / independently from online services, so feel free to close 🕊

As far as I'm concerned, there is no legal way to obtain Titanfall 2 game files except through Origin or Steam, so this does not constitute an act of piracy. 

Software DRM removal is not piracy in itself, or at least it's not legally sorted out (and probably won't ever be). Distribution of game files is more clear cut. Other projects I've worked on have had DRM removal for years without an issue, distribution of game files however always brought a lot of trouble.

I understand Northstar not wanting to risk it legally, so I can keep this on my fork - I suppose we can reopen the can of worms when EA disappears from the face of the earth.

louvenarde:
Ill check tomorrow what I can hair split and cherry pick (probably the CMakeList and the offline write as you said), but that'll be on another branch

louvenarde:
hold up I had an idea tonight but I'm not sure you'll like it

louvenarde:
Do you like it better this way https://github.com/R2Northstar/NorthstarLauncher/pull/930/changes/583ea6e67cf4f464348fc38280d6fd7b31cbc800#diff-c69846729198f147c2beea5fdde9427cc2c9f44f510ba458a3b80fb3b1ea1de0R8 

You still get the same feature but it now requires Origin™ even for offline play, and the proxy is put behind this compile-time flag so you can have the definition you want for what constitutes piracy or not. If one day the global situation changes you have this ready to run 🤙 

> (Note that this makes the feature practically useless in the meantime, turning it into an "online offline mode", but at least you get the flexibility of changing your mind later on and also you get the other tiny improvements here and there)

ASpoonPlaysGames:
Conceptually this is better imo, but something more specific than `YES`/`NO` for the defines would be appreciated

primedev/server/auth/serverauthentication.cpp
```diff
@@ -208,8 +218,57 @@ void ServerAuthenticationManager::WritePersistentData(CBaseClient* pPlayer)
...
+		const auto dir = path.parent_path();
+		std::filesystem::create_directory(dir);
+
+		std::ofstream localData(path, std::ios::binary | std::ios::out);
+		if (localData.is_open())
+		{
+			localData.write(reinterpret_cast<const char*>(pPlayer->m_PersistenceBuffer), size);
+		}
+	}
+}
```
ASpoonPlaysGames, line 253: This will likely only work for local servers, since pdata is server authoritative. If you want pdata stored on client machines that can be used on servers you'll need to network it to the server and let the server set it on connect (preferably before the persistence baseline is networked)

primedev/server/auth/serverauthentication.cpp
```diff
@@ -208,8 +218,57 @@ void ServerAuthenticationManager::WritePersistentData(CBaseClient* pPlayer)
...
+
+std::filesystem::path ServerAuthenticationManager::GetOfflinePersistentDataPath()
+{
+	constexpr auto filename = "persistent.bin";
```
ASpoonPlaysGames, line 257: `.pdata` is probably a better file extension here, it's what we've used in the past

primedev/server/auth/serverauthentication.cpp
louvenarde: Yes, we discussed this on the discord, the protocol has to be slightly changed - in the meantime persistent player data needs to be at least separated. That also means locally persistent player GUIDs instead of random on each start. I'm gonna look into all that.

louvenarde:
Summary:
- The server no longer writes persistent data offline, only the client does, and only when disconnecting
- There is an exception for self server (mp_lobby hosted by the local client) in which case, they do save their own data
- Persistent data is now named `persistent.pdata`
- In offline write mode, the persistent data is attached to the C2S `connect` payload, making it a huge packet with a bunch of fragmentation. This is not a problem for LAN mode but for a public facing server it can be a little rough, fortunately the packet is compressed so it's rarely 50kB (it will be around 2 or 3 KB from what I saw)
- So you can now edit your class locally or on a server, you will take it with you no matter where you go :)

louvenarde:
Ready for review :)

louvenarde:
Since this has not been reviewed yet I've taken the liberty of tackling a few TODOs on this PR, notably LAN discovery. This brought me to refactor a thing or two on Northstar and I shall push them on this PR tonight or asap, i'm currently having issues with NET Encryption for LAN discovery (using the game-given sockets instead of my own). Will keep you posted, might push a few WIPs 😸

louvenarde:
Heads up! I've updated the branch with LAN Discovery :)
That means LAN games (on the default port) now appear in the server browser naturally.
image

- Since I had to add a new Connectionless message type to the game for that, and there was already one (for Atlas), I decided to refactor and centralize this with a new component called `custom_packet_handler.cpp` Now any script can extend the connectionless protocol by calling `RegisterCustomPacketHandler(...)`;
- I had to encrypt the discovery messages to reuse the server socket (allowing discovery to work on the game port, which is simpler for the host) and therefore I also documented, typed and defined the NET_Encrypt/Decrypt function from the engine for everyone to use

It's ready for review, I won't touch this PR until then, it's feature complete as far as I wanted to take it ❤️

sonny-tel (changes requested):
I don't even know, code here looks extremely LLM based with absolutely no sane placement of anything, i.e hooking NET_SendPacket inside the namespace/class of offline persistence and the random placement of bitbuf is an obvious vibe smell. Moving the server connectionless packet handler is a good idea but the placement and choice is terrible. Among a lot of these changes are random debugging values and convars being changed and committed for no good reason. Also the usefulness of a lan-only mode for Northstar is shaky at best and would best be relegated to a plugin or fork. Functionality aside this needs lots of work and I'm not entirely sure if this will ever be merged anyway

sonny-tel:
After discussing with @catornot, I'm going to close this

louvenarde:
> I don't even know, code here looks extremely LLM based with absolutely no sane placement of anything, i.e hooking NET_SendPacket inside the namespace/class of offline persistence and the random placement of bitbuf is an obvious vibe smell

Thank you for your kind review, just to make it clear I'd rather choke to death by swallowing my own teeth than interacting with anything AI, so this was very much not "vibe" or "AI" or "LLM" assisted or coded or anything. I don't use it for RE. I don't use it for code. I don't use it for art. I don't use it to fill my agenda or brush my teeth.

- Hooking NET_SendPacket inside the namespace is what you do in the rest of the project (hooking directly in the namespace) so this is what I did everytime I could
- However at some occasion it broke encapsulation because I wanted my hook to add data to a private collection, and I wanted the functions that interacted with that data to be kept private to keep them from being called from other places on the project. The way I did this is with a private static function so the hook could go through _and_ the encapsulation would be preserved. If you prefer I can also make all of that public/namespace level. Really the job of a review is to point this out so I can change it, not to call me a fucking AI.

> Moving the server connectionless packet handler is a good idea but the placement and choice is terrible

Thank you for your kind words, may you explain what is the better "choice" and "placement" here? Anything besides "it is terrible" for someone who just arrived on the project? 

> Among a lot of these changes are random debugging values and convars being changed and committed for no good reason

Which ones? ns_isnecure_write and ns_insecure_auth are changed because they allow offline persistence and integrate LAN mode into pre-existing feature, like offline pdata. What other convars?
The debugging is WSA Socket hooking necessary to debug multiplayer communications, I can remove them if you want, but know that they are _already disabled behind a macro_. They're very much not random; i needed them to make this PR work and understand the game's NET protocol!

> Also the usefulness of a lan-only mode for Northstar is shaky at best and would best be relegated to a plugin or fork

This is the only statement that is actually actionable, but it does not help me understand where you want the PR split to start and which parts you want and don't want. Can you be more precise about what you want to keep and where? Do you want the persistent data payload connect? Do you want the persistent data storage? Do you want the proxy? Do you want the offline mode? Do you want the server discovery? Do you want the protocol changes? Do you want the packet handler?

> Functionality aside this needs lots of work

What work specifically! You're reviewing so give me some directions here. I'm all ears.

ASpoonPlaysGames:
Fwiw this does not scream AI to me, regardless of any perceived code quality. I would reopen the PR but it's been split already so I don't see much point

PR #931, opened 2026-08-10
Hardcode cardinals of all exports in wsock32

catornot:
this fixes the recent regression in the release.

I would not even try to justify such a blunder, but this is also just absurd. either way it seems like certain titanfall 2 dlls on windows care a lot more about ordinals than on linux which is how it seemingly never got caught.

not sure if everything needs to be hardcoded but I tested it on fresh install of windows 10 22H2 where it seems to work with this pr. therefore I will push a new release with this change.


AllusiveWheat (approved):
it probably works, seems fine

sonny-tel:
this is a bit excessive, the issue is probably only present due to the original .def file having the typo `getnetbyname=PROXY_getnetbyname @ 1101` instead of `getnetbyname=PROXY_getnetbyname @1101`. This is at least how I fixed this issue 6 months ago I think

louvenarde:
I don't know what this was supposed to fix, but the WSock proxy (from main, not this PR) caused a hard freeze for me right after the logos, and this fixed it (I downloaded an artifact to test it live in my install). The game now launches properly with that proxy, as opposed to freezing completely with a proxy freshly built from main 👍 

I did not test ingame further

catornot:
> this is a bit excessive, the issue is probably only present due to the original .def file having the typo getnetbyname=PROXY_getnetbyname @ 1101 instead of getnetbyname=PROXY_getnetbyname @1101. This is at least how I fixed this issue 6 months ago I think

tested this and it doesn't work, at least for clang-cl.

therefore I am merging this for now at least.

PR #932, opened 2026-08-20
Default compile constants to 0 if not defined 

AllusiveWheat:
Default compile time constants like `#ifdef VANILLIA` to 0 if they do not exist.

### DIFF
--- a/primedev/squirrel/squirrel.cpp
+++ b/primedev/squirrel/squirrel.cpp
@@ -20,7 +20,7 @@
 AUTOHOOK_INIT()
 
 SquirrelManagerManager g_pSquirrel;
-
+thread_local SQCompiler* pIfDirectiveCompiler;
 std::shared_ptr<ColoredLogger> getSquirrelLoggerByContext(ScriptContext context)
 {
 	switch (context)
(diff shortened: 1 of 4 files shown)


primedev/squirrel/squirrel.cpp
```diff
@@ -326,9 +326,42 @@ bool IsUIVM(ScriptContext context, HSQUIRRELVM pSqvm)
...
+		return true;
+
+	if (!pIfDirectiveCompiler || pIfDirectiveCompiler->preprocessorDepth <= 0)
+		return false;
+
+	// evaluate to 0 if we're doing directives because #if with a compile error is stupid
+	pValue->_Type = OT_INTEGER;
+	pValue->structNumber = 0;
+	pValue->_VAL.as64Integer = 0;
+	constexpr std::uint32_t typeHash = 0x9E3779B9u - 0x61C88647u * static_cast<std::uint32_t>(OT_INTEGER);
```
catornot: can we un magic these magic numbers pls

catornot (changes requested):
cool stuff, just trusting the offsets for now

for future work it would be cool to have a table for mod dependency constants implemented in a similar way to create a namespace for them :)

primedev/squirrel/squirrel.cpp
```diff
@@ -326,9 +326,42 @@ bool IsUIVM(ScriptContext context, HSQUIRRELVM pSqvm)
...
+
+template <ScriptContext context>
+bool (*SQCompiler_ResolveLocalOrConstant)(void* pFunctionState, SQObject* pIdentifier, SQObject* pValue, std::uintptr_t* pType);
+template <ScriptContext context>
+bool __fastcall SQCompiler_ResolveLocalOrConstantHook(void* pFunctionState, SQObject* pIdentifier, SQObject* pValue, std::uintptr_t* pType)
+{
+	if (SQCompiler_ResolveLocalOrConstant<context>(pFunctionState, pIdentifier, pValue, pType))
+		return true;
+
+	if (!pIfDirectiveCompiler || pIfDirectiveCompiler->preprocessorDepth <= 0)
```
catornot, line 347: could play leave a comment on what this check does :)

but I assume this checks if this function call is not in a preprocess step?

catornot (approved):
worked and looks good

this is a good start for future work :)

PR #933, opened 2026-08-30
Custom packet handler to extend the Connectionless protocol

louvenarde:
Created a custom packet handler so that the connectionless protocol can be extended safely and easily in more than 1 place. This is a prerequisite for LAN discovery (which needs a discovery message to be registered).
I've also taken the liberty of renaming a var in the NET_ReceiveDatagram because it is misnomed - it assumes a socket is being passed when it is the _socket index_ that is passed, the game registers and manages four sockets and this is an index of that.

### Testing:

I've tested it with LAN mode and the custom packet handler works well, but I don't know how to test the original "Atlas" packet because my game never seems to receive it (breakpoint never hits)

### DIFF
--- a/primedev/engine/custom_packet_handler.h
+++ b/primedev/engine/custom_packet_handler.h
@@ -1,0 +1,19 @@
+#pragma once
+
+struct CustomPacketHandler
+{
+	typedef void (*CustomPacketHandlerType)(void* packetHandler, struct netpacket_s* packet, OUT bool& runOriginalHandler);
+
+public:
+	bool RegisterPacketHandler(uint8_t controller, CustomPacketHandlerType handler);
+
+	CustomPacketHandler();
+
+private:
+	static bool (*o_pHandlePacket)(void* packetHandler, netpacket_s* packet);
+	static bool h_HandlePacket(void* packetHandler, netpacket_s* packet);
+
+	std::unordered_map<uint8_t, CustomPacketHandlerType> customHandlers {};
+};
+
+extern CustomPacketHandler* g_pCustomPacketHandler;
--- a/primedev/Northstar.cmake
+++ b/primedev/Northstar.cmake
@@ -61,6 +61,8 @@ add_library(
     "dedicated/dedicatedlogtoclient.cpp"
     "dedicated/dedicatedlogtoclient.h"
     "dedicated/dedicatedmaterialsystem.cpp"
+    "engine/custom_packet_handler.cpp"
+    "engine/custom_packet_handler.h"
     "engine/gl_matsysiface.cpp"
     "engine/host.cpp"
     "engine/hoststate.cpp"
--- a/primedev/engine/custom_packet_handler.cpp
+++ b/primedev/engine/custom_packet_handler.cpp
@@ -1,0 +1,75 @@
+#include "engine/custom_packet_handler.h"
+#include "r2engine.h"
+#include <shared/exploit_fixes/ns_limits.h>
+
+CustomPacketHandler* g_pCustomPacketHandler;
+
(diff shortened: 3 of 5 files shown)


louvenarde:
> Moving the server connectionless packet handler is a good idea but the placement and choice is terrible

@sonny-tel Would you like to elaborate on this past review and help me move it to a better "placement and choice" ?
In the objective to get this PR going I'd like to address whatever issue any of you might have with it 🙂

primedev/engine/custom_packet_handler.h
```diff
@@ -0,0 +1,19 @@
+#pragma once
+
+struct CustomPacketHandler
+{
+	typedef void (*CustomPacketHandlerType)(void* packetHandler, struct netpacket_s* packet, OUT bool& runOriginalHandler);
+
+public:
```
RoyalBlue1, line 7: using public or private in a struct does not do anything

RoyalBlue1 (commented):
While I think that the custom handler does not need to be a class i dont really care that it is please just make it a singleton than because otherwise the hook might be called multiple times

primedev/engine/custom_packet_handler.h
ASpoonPlaysGames: `public` does nothing in a struct, `private` does.

Structs and classes are basically identical except class members are private by default and struct members are public by default

louvenarde: I made it a singleton by following the structure present elsewhere ( a single instance named g_pCustomPacketHandler etc). Took example on g_serverManager and others

- Do you intend to mean that I should make it a public static class instead? (This is 100% doable, I will be following your specs if that's what you want 🙂 )
- Which hook risks being called multiple times? Normally instantiation is done on the ON_LOAD(engine.dll) hook which I figured would only ever be called once. Is there a situation where this can be loaded and instantiated multiple times? (Maybe I can assert/nullcheck the `g_` variable in that case?)

RoyalBlue1: theres no way of happening it now and we have that pattern in other places so its fine. what i meant was that just having it as a class an unknowing future dev might misuse it and create another instance of the struct since the name handler gives of a multi instance kinda vibe to me. I guess CustomPacketManager would be better but this is fine

RoyalBlue1 (approved):
code is fine as per previous conversation on the pr thread

primedev/engine/custom_packet_handler.h
louvenarde: It does not have to be "fine", it's your spec, I'm happy to rename and restructure as you want ❤️ 
If I accidentally imitated some rot, I should probably do another structure that makes you happy so that I don't add more rot into the overall project. Ideally additions improve the overall project health and do not decrease it

RoyalBlue1: ok my main thing is the name and that i think that hooks should really only be done inside on dll load functions unless its strictly necessary for the implementation. look at primedev/server/auth/serverauthentication.cpp on what i mean

louvenarde: Renaming is no problem.

Putting the hooks inside the ON_DLL macro and outside the ctor is OK to me, but note that I will have to either:
- Break encapsulation (make some class functions public that otherwise would have no business being public)
- Or make another static class just for this hook that is Friends of the custom packet handler class

Server authentication is a good example, e.g. `ServerAuthenticationManager::AuthenticatePlayer()` is clearly meant to be a private function because it's only called within the context of `serverauthentification.cpp`. It's an internal function.
**But it is still set to** `public` because otherwise the hooks can't talk to it, so encapsulation had to be broken for this structure to take place. It's more C-style than C++

I will make the changes tomorrow to break encapsulation and make internal functions publicly accessible like in serverauthentication.cpp

PR #935, opened 2026-08-30
Hopefix stack corruption that occured here one time?

louvenarde:
I had a corruption error here once, as if more data had been copied to the array than there should be. It might have been entirely intermittent but the debugger pointed to this line as the culprit.

I'm not sure why exactly and this error disappeared once I put this fix in place, so that might be it? It looks rather safe so I thought i'd share it. If it's not appropriate or relevant, feel free to close PR and move on

Since the buffer is set to zero with {}, null termination is mandatory, so copying 2048-1 bytes ensure that the buffer will always be null terminated no matter the length of 'statusText'. When I checked it with a debugger it didn't seem usually that long but... it's technically possible that it would be, I suppose


AllusiveWheat (approved):
Code looks good, never had the crash but looks good.

Issue #938, opened 2026-09-02
wsock32 proxy fails to locate R2Northstar/Northstar.dll and falls back to nonexistent root DLL


TB516:
### Description

Northstar 1.31.13 fails to launch because the wsock32 proxy looks for:

`<Titanfall2>\Northstar.dll`

but the DLL is actually installed at:

`<Titanfall2>\R2Northstar\Northstar.dll`

I reproduced this on Linux with Proton GE. A friend also reproduced the same issue on Windows, so it doesn't appear to be Proton-specific.

The error is:

> Northstar Wsock32 Proxy Error  
> Failed to load the Northstar.dll at "<Titanfall2>\Northstar.dll" (126):  
> Module not found.

### Steps to reproduce

1. Install Northstar 1.31.13 normally.
2. Launch Northstar.
3. The wsock32 proxy fails to find `Northstar.dll`.
4. Check the game directory:
   - `Northstar.dll` does not exist in the root.
   - `R2Northstar/Northstar.dll` does exist.

### Workaround

On Linux, I created a symlink:

```sh
ln -s R2Northstar/Northstar.dll Northstar.dll
```

Northstar launches normally after doing this.

This makes it look like the DLL itself is fine and the wsock32 proxy is falling back to the old root path instead of finding the DLL in the profile directory.

### Expected behavior

The proxy should load:

`R2Northstar/Northstar.dll`

without requiring a copy or symlink in the game root.

### Environment

- Northstar: 1.31.13
- Reproduced on:
  - Linux / Steam / Proton GE
  - Windows

I can test a build or provide additional logs if needed.

catornot:
cannot reproduce on linux

TB516:
I installed northstar with r2modman, and my friend who had it on windows also was using r2modman, so it might be the thunderstore version or something cause a third friend using the [viper flatpak](https://flathub.org/en/apps/com.github._0negal.Viper) had it working fine without needing the dll moved. All 3 of us were using a fresh titanfall install btw.

catornot:
oh yeah for r2modman it makes sense, tho I think they need to fix it on their side

TB516:
Ive worked on r2modman before so I can take a look at stuff there, but im not sure why it would be an error from Wsock32 then. Ill look over there and see if theres something weird going on with the file or if its not getting installed correctly or something then give an update.

TB516:
Seems like it is on r2modman's end, so ill go make a fix there!

[...]

PR #939, opened 2026-09-09
Fix loading Northstar from absolute profile paths

TB516:
This PR was created by GPT-6

An absolute `-profile` path was prefixed with the game directory, causing the launcher and wsock32 proxy to miss the profile's `Northstar.dll` and fall back to the game root. Use the existing `std::filesystem` path joining to preserve absolute profiles and resolve relative profiles against the game directory.

Related to #938.

### Code review:

The change is limited to path construction in both loaders. No new dependencies; the existing root-directory fallback is preserved.

### Testing:

- Full Release build passed using MSVC 2022 under Wine through Podman.
- Thomas confirmed the rebuilt proxy launches successfully under Proton with `-northstar -profile="Z:/var/mnt/Games/r2modman-mods/Titanfall2/profiles/Default/R2Northstar"`, without the root DLL symlink workaround.
- To reproduce, use a profile outside the game directory containing `Northstar.dll`, remove any root DLL workaround, and launch with `-northstar -profile="<absolute Windows profile path>"`. The console's `[*] Using:` line should point to the profile DLL. The standalone launcher uses the same fix but has not been runtime-tested.

r2modman also needs a separate Linux fix to supply a `Z:`-prefixed path under Proton. Full verification through a patched r2modman build is still pending.


TB516:
Im not able to test this on windows but can confirm that its good on linux through proton.

ASpoonPlaysGames (changes requested):
This is not what I expected when you said you'd be willing to make a PR.

I could be wrong but I'm fairly certain Northstar doesn't accept AI generated PRs

catornot:
> I could be wrong but I'm fairly certain Northstar doesn't accept AI generated PRs

nothing official but I personally would not merge ai generated stuff but in this case the change is so small I doubt it could be written differently

ASpoonPlaysGames:
Splitting hairs a bit but local variables shouldn't use pascal case

TB516:
I can go fix the casing in a few minutes sure.

Also ya I didn't see any policy on AI, so I threw codex at the issue and asked it to give me stuff to test. I figured since there was no policy it was fine so my bad there.

ASpoonPlaysGames:
> Also ya I didn't see any policy on AI

There's no official one, but most contributors are against AI in general

ASpoonPlaysGames (approved):
Despite this being AI generated, it's a very small change and like @catornot said there really isn't more than the one way to do this

LGTM

PR #940, opened 2026-09-17
fix lag

LightBlueCube:
only fully tested on NSCN, i assume its will work same on here

this patchs make sure server reads the lost user commands instead of just ignore it

https://github.com/user-attachments/assets/e0d77e87-cc27-4155-be20-3c1f088f3678

this is so dumb


LightBlueCube:
UPDATE: we may wanna set a higher default for `cl_cmdbackup` (maximum 7, bcz its storaged as 3bit)

this is for get better playable when sv<-cl pakloss

